PR #79, opened 2014-07-21
Restrict OpenEye tests from running on travis [WIP]

jchodera:
I've added a method `yank.systembuilder.SmallMoleculeBuilder.is_supported()` that returns `True` if the OpenEye libraries can be correctly loaded on this platform, and modified the tests to only run if this is supported.


jchodera:
This is stale.

PR #144, opened 2014-12-15
Began implementing Cython version of replica mixing

pgrinaway:
I've added a Cython file, as well as a `cythonize()` call in setup.py. Still a WIP, but wanted to put up the code now.


jchodera:
We can add tests in a separate PR.

PR #162, opened 2015-01-12
Deleted some functions in analyze.py

kyleabeauchamp:


jchodera:
Merging.

jchodera:
The error is an issue with the new `LennardJonesFluid` in [`openmmtools`](https://github.com/choderalab/openmmtools).  See https://github.com/choderalab/openmmtools/issues/27

PR #169, opened 2015-01-12
Updated license and README

kyleabeauchamp:
Looks like the correct LICENSE file was already present, we just needed to update the README.md file


jchodera:
Looks good.

PR #172, opened 2015-01-12
Deleted several more unused functions.

kyleabeauchamp:


kyleabeauchamp:
AFAIK these functions are unused and can be deleted, but we should double check that we don't need them in some code path that we will need.

jchodera:
These all can go!

Issue #16, opened 2013-11-08
import numpy as np


kyleabeauchamp:
IMHO, this is the "standard" way of using numpy imports.  

Also, it might be good to try to use all numpy, whenever possible, rather than switching between `math` and `numpy`

pgrinaway:
resolved by #179

PR #179, opened 2015-01-14
Changed import numpy to import numpy as np

pgrinaway:
Ok, per #16, I've changed `import numpy` statements to `import numpy as np`. Let me know if I've missed anything...otherwise this should be relatively straightforward.

### DIFF
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -64,8 +64,7 @@ import copy
 import time
 import datetime
 import mixing._mix_replicas as _mix_replicas
-import numpy
-import numpy.linalg
+import numpy as np
 import mdtraj as md
 import mixing._mix_replicas_old as _mix_replicas_old
 from simtk import openmm
@@ -247,7 +246,7 @@ class ThermodynamicState(object):
 
         ARGUMENTS
 
-        positions (simtk.unit.Quantity of Nx3 numpy.array) - positions[n,k] is kth coordinate of particle n
+        positions (simtk.unit.Quantity of Nx3 np.array) - positions[n,k] is kth coordinate of particle n
 
         OPTIONAL ARGUMENTS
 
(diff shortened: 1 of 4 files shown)


Yank/repex.py
```diff
@@ -1340,9 +1339,9 @@ def _propagate_replica(self, replica_index):
         getstate_start_time = time.time()
         openmm_state = context.getState(getPositions=True)
         getstate_end_time = time.time()
-        self.replica_positions[replica_index] = openmm_state.getPositions(asNumpy=True)
```
kyleabeauchamp: -1

Yank/repex.py
```diff
@@ -1340,9 +1339,9 @@ def _propagate_replica(self, replica_index):
...
+        self.replica_positions[replica_index] = openmm_state.getPositions(asnp=True)
         # Store box vectors.
-        self.replica_box_vectors[replica_index] = openmm_state.getPeriodicBoxVectors(asNumpy=True)
```
kyleabeauchamp: -1

Yank/repex.py
```diff
@@ -1465,7 +1464,7 @@ def _minimize_replica(self, replica_index):
         minimized_positions = self.mm.LocalEnergyMinimizer.minimize(context, self.minimize_tolerance, self.minimize_maxIterations)
         # Store final positions
         openmm_state = context.getState(getPositions=True)
-        self.replica_positions[replica_index] = openmm_state.getPositions(asNumpy=True)
+        self.replica_positions[replica_index] = openmm_state.getPositions(asnp=True)
```
kyleabeauchamp: -1

kyleabeauchamp:
Looks reasonable once we fix the openmm copy paste issue and tests pass.

pgrinaway:
Ok, hang on.

kyleabeauchamp:
Merging now.  We should fix the OpenMMTools bug that's causing the segfault, though, as I think we want to get travis working soon.

jchodera:
The segfault must be an openmm bug. We should try to produce a minimal case
for a bug report.

kyleabeauchamp:
Given how long that will take to be released, IMHO we should first do a workaround in openmmtools.  Otherwise the yank release will have to wait until the *next* openmm release.

jchodera:
> Given how long that will take to be released, IMHO we should first do a workaround in openmmtools. Otherwise the yank release will have to wait until the next openmm release.

Agreed.  I had tried to revert this by adding back `LennardJonesGrid`, but that didn't seem to fix the segfault. Will debug a bit more.

PR #181, opened 2015-01-15
Mixing tests

pgrinaway:
This is a very WIP (just started writing) PR for adding tests for the mixing code. I figured I'd open a PR ASAP so everyone knows what is going on.

### DIFF
--- a/Yank/mixing/_mix_replicas.pyx
+++ b/Yank/mixing/_mix_replicas.pyx
@@ -7,9 +7,8 @@ cdef extern from 'stdlib.h' nogil:
 @cython.wraparound(False)
 @cython.cdivision(True)
 @cython.boundscheck(False)
-cpdef long _mix_replicas_cython(long nstates, long[:] replica_states, double[:,:] u_kl, long[:,:] Nij_proposed, long[:,:] Nij_accepted) nogil:
+cpdef long _mix_replicas_cython(long nswap_attempts, long nstates, long[:] replica_states, double[:,:] u_kl, long[:,:] Nij_proposed, long[:,:] Nij_accepted) nogil:
     cdef long swap_attempt
-    cdef long nswap_attempts = nstates*nstates*nstates*nstates
     cdef long i, j, istate, jstate, tmp_state
     cdef double log_P_accept
     for swap_attempt in range(nswap_attempts):
--- a/Yank/mixing/_mix_replicas.pyx
+++ b/Yank/mixing/_mix_replicas.pyx
@@ -12,7 +12,6 @@ cpdef long _mix_replicas_cython(long nstates, long[:] replica_states, double[:,:
     cdef long nswap_attempts = nstates*nstates*nstates*nstates
     cdef long i, j, istate, jstate, tmp_state
     cdef double log_P_accept
-    printf("Attempting to swap all replicas using Cython-accelerated code\n")
     for swap_attempt in range(nswap_attempts):
         i = <long>(drand48()*nstates)
         j = <long>(drand48()*nstates)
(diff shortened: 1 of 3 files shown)


pgrinaway:
Not sure what's going on here with the test w/ reduced potentials drawn from a normal distribution. The chi-square test is indicating that the results are wildly off--either there is something wrong with the mixer or the test (right now I'm leaning toward the latter...). I'll come back to this tomorrow morning after some sleep, and try to figure out what the deal with that test is.

jchodera:
I just realized that I was totally wrong about the marginal distribution of the states s_i for the case where the reduced potential energy elements u_{ij} are not uniform.

If you use uniform u_{ij}, the marginal distribution of s_i should be uniform. So that should work.

For the more general case, we apparently cannot neglect the contribution from the other states.  Let perm(X) denote the [matrix permanent](http://en.wikipedia.org/wiki/Permanent) of X, which is the same as the determinant except with addition instead of subtraction.  Then the marginal distribution of s_1 is

p(s_i) \propto e^{-u_{ij}} perm(X_i)

where X is the submatrix of U = (u_{ij}) that omits row and column i.  This can be done with the `np.delete` command, such as:
```python
Z = np.delete(U, i, axis=0)
X = np.delete(Z, i, axis=1)
```
There may be a more convenient way to do this.

Note that there's really no trick to efficient evaluation of the matrix permanent, so this would be slow for large matrices.  It's best to keep the test systems small as a result.

Numpy doesn't provide an implementation of the matrix permanent, but simple recursive implementations (e.g. [this one](https://github.com/arulalant/numpy-utils/blob/master/lib/numutils.py#L133-L189)) should work fine.

Numerical issues may also be a problem, so it's best to keep the dynamic range of entries of u to be small.

pgrinaway:
Ok, +10 on that, thanks! I'll get started fixing this...

pgrinaway:
Looks like he also has a function to [remove](https://github.com/arulalant/numpy-utils/blob/master/lib/numutils.py#L192-L257) a certain column and row of a matrix, though I'm not sure how efficient that is.

jchodera:
Correction again! The matrix X should be the permanent of the matrix whose elements are exp(-u_ij), with the ith row and column removed.

pgrinaway:
Just FYI, this test fails on the weave mixing code, too. Still looking into it.

jchodera:
Goodness!  That's a little worrisome.  Let me know what you find.

One idea is that the weave code may not be mixing enough...

pgrinaway:
Ok, that's a good point. Both the weave and Cython code decide the number of swap attempts based on `nstates^4`. Because of the complexity of calculating the permanent, I am using an 8x8 matrix... maybe 4096 isn't enough? Perhaps I should just add a parameter to decide the number of swap attempts.

pgrinaway:
Ok, I increased the number of swap attempts to 2**22 for a 4x4 matrix, which caused one test success and several with more reasonable p-values. However, it's no longer working. For a 2x2 matrix, it passed, but I don't know if there is any value to that.

pgrinaway:
Alright, I'm taking a look at the R tests for [parallel tempering](http://www.stat.umn.edu/geyer/mcmc/), in `tests/temp-par.R`.
```
 for (i in 1:nmodel) {
     for (j in 1:nmodel) {
         if (out$neighbors[i, j]) {
             my.accepti[i, j] <-
                 mean(accept.swap[my.i.swap == i & my.j.swap == j])
         }
     }
 }
 identical(my.accepti, out$accepti)
````
Besides being in an indecipherable language (hah), is this just testing acceptance rates? That might be a more straightforward thing to do. Continuing to look...

pgrinaway:
Ok, following a discussion with @jchodera, I think we'll add in the `test_even_mixing` and remove the random energies version until we figure this out. Going to remove said code now (I'll keep it in a branch for further investigation).

pgrinaway:
I'm running nosetests on my machine right now, because travis has been segfaulting on something unrelated.

pgrinaway:
Ok, it failed for some other reason. I'm going to make this less verbose, then we can merge.

jchodera:
Great!

Issue #93, opened 2014-10-04
Remediate OpenMM version dependency tracking in setup.py now that we will use conda


jchodera:
There's a bunch of code in `setup.py` we probably don't need if we're using version-specific dependencies via `conda`:
https://github.com/choderalab/yank/blob/master/setup.py#L63-L70

Will deal with this later.

jchodera:
It is now:
https://github.com/choderalab/yank/blob/master/setup.py#L112-L146

Should this now look like the [`setup.py` from MDTraj](https://github.com/mdtraj/mdtraj/blob/master/setup.py)?

pgrinaway:
Resolved by #185

PR #185, opened 2015-01-17
Begin removing extraneous code from setup.py/cleaning up

pgrinaway:
Removed the `check_dependencies` function and the call to it to address #93

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -109,41 +109,6 @@ def find_package_data(data_root, package_root):
             files.append(relpath(join(root, fn), package_root))
     return files
 
-def check_dependencies():
-    from distutils.version import StrictVersion
-    found_openmm = True
-    found_openmm_61_or_later = True
-    found_numpy = True
-
-    try:
-        from simtk import openmm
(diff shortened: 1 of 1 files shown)


pgrinaway:
Not sure what the deal with that error in Travis is, but I think it is probably for a separate PR (given that it has been occurring in previous builds)

jchodera:
Yeah, there are some things we need to fix in separate PRs.

pgrinaway:
Ok, otherwise I think this is all that had to happen with setup.py? Correct me if I'm wrong...

pgrinaway:
@jchodera @kyleabeauchamp am I missing something else that we wanted to do here or is this ready to merge? I think that we only needed to remove the dependency check, right?

jchodera:
I think that's it, right?

kyleabeauchamp:
Lgtm

pgrinaway:
Ok, I'm going to go ahead and merge this and take care of the travis error in a future PR.

Issue #187, opened 2015-01-21
Fixing travis


jchodera:
Any idea what might be going on here with this travis failure?  I'm not even sure where the test is that is causing this failure.

https://travis-ci.org/choderalab/yank#L1357-L1371
```
======================================================================
ERROR: test suite for <module 'yank.commands' from '/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/yank-0.9.0-py2.7-linux-x86_64.egg/yank/commands/__init__.py'>
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/nose/suite.py", line 209, in run
    self.setUp()
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/nose/suite.py", line 292, in setUp
    self.setupContext(ancestor)
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/nose/suite.py", line 315, in setupContext
    try_run(context, names)
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/nose/util.py", line 465, in try_run
    (name, obj))
TypeError: Attribute setup of <module 'yank.commands' from '/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/yank-0.9.0-py2.7-linux-x86_64.egg/yank/commands/__init__.py'> is not a python function. Only functions or callables may be used as fixtures.
```

jchodera:
OK, I'm very confused. Running `nosetests Yank --nocapture --verbosity=2 --with-doctest` on my laptop does show a bunch of different errors:
```
[LSKI1497:~/code/yank/yank.jchodera] choderaj% nosetests Yank --nocapture --verbosity=2 --with-doctest
Doctest: Yank.alchemy.AbsoluteAlchemicalFactory ... ok
Doctest: Yank.alchemy.AbsoluteAlchemicalFactory._is_restraint ... ok
Doctest: Yank.alchemy.AbsoluteAlchemicalFactory.createPerturbedSystem ... ok
Doctest: Yank.alchemy.AbsoluteAlchemicalFactory.createPerturbedSystems ... ok
Doctest: Yank.alchemy.AbsoluteAlchemicalFactory.perturbContext ... ok
Doctest: Yank.alchemy.AbsoluteAlchemicalFactory.perturbSystem ... ok
Doctest: Yank.alchemy.AlchemicalState.__init__ ... ok
[...]

PR #188, opened 2015-01-21
[WIP] Attempting to fix nosetests on travis

jchodera:
Trying to figure out what is going on...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,13 +11,16 @@ install:
 script:
   # this builds the binary, unpacks it, and runs the tests
   - conda build devtools/conda-recipe
-  # Push to binstar anyway.
-  #- bash -x devtools/ci/after_sucess.sh  # Temporarily holding off on forced binstar push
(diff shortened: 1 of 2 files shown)


kyleabeauchamp:
So the build already runs the tests, so I don't think we need to separately run nose tests.

kyleabeauchamp:
OK, so I see you moved the tests back out of the recipe.

jchodera:
> So the build already runs the tests, so I don't think we need to separately run nose tests.

That wasn't working---the tests are failing for some reason I don't understand (since they work on my machine).  I've moved them out of the build recipe, possibly temporarily.

jchodera:
Wait a sec---I *do* see these failures on my laptop during a conda build, but not when I run standalone nosetests.  I'm still not sure where the errors in #187 are coming from.

jchodera:
Deprecated; fixing through a new PR based on @kyleabeauchamp's suggestion from #187.

PR #189, opened 2015-01-21
Changed "yank setup" to "yank prepare" throughout to avoid weird nose bug

jchodera:
This attempts to address #187. Lots of changes were necessary---not sure if I got everything changed.


examples/SAMPL-implicit/README.md
```diff
@@ -20,7 +20,7 @@ The mbondi2 radii are used, with OBC GBSA in YANK.
 
 Set up the simulation to alchemically decouple benzene, putting all the output files in `output/`:
 ```tcsh
-yank setup binding amber --setupdir=setup --ligname=MOL --store=output --iterations=1000 --restraints=harmonic --gbsa=OBC2 --temperature=300*kelvin --minimize --verbose
+yank preparebinding amber --setupdir=setup --ligname=MOL --store=output --iterations=1000 --restraints=harmonic --gbsa=OBC2 --temperature=300*kelvin --minimize --verbose
```
kyleabeauchamp: space

pgrinaway: prepare binding ?

pgrinaway:
I think [this error](https://travis-ci.org/choderalab/yank/builds/47839560#L1348-L1369) refers to retaining the old `yank setup binding` instead of the new `yank prepare binding`

jchodera:
Is there a way to get nose to print out what file it's processing tests for?  Trying to figure out where these errors come from is becoming infuriating.

kyleabeauchamp:
By default I think nose prints `filename.functionname()`, but that gets replaced by the docstring when present.

bas-rustenburg:
Perhaps

```
  --logging-format=FORMAT
                        Specify custom format to print statements. Uses the
                        same format as used by standard logging handlers.
                        [NOSE_LOGFORMAT]
```

Could be useful?

jchodera:
There's also some plugins that may help:
https://pypi.python.org/pypi/nose-progressive/

jchodera:
Looks like we're just failing on the push to AWS because I haven't finished configuring the AWS encrypted keys.  If this looks good to everyone else, we should probably merge it.

PR #190, opened 2015-01-22
Cleanup CI

kyleabeauchamp:
In the last PR, we merged in some things that we probably don't need.  I think we should delete them to make our CI system less complicated.


kyleabeauchamp:
Whenever possible, I want to use msmbuilder as the template for our CI.  MDTraj is somewhat more complicated.

jchodera:
Got it. I ended up having to add these things to get it to work, but I
agree that less complexity is better.

Even setting up a skeleton for a new project is enormously time consuming.
Wish there was a solution to that.

pgrinaway:
I think it might be a good idea to merge this PR, then I can start trying to deal with whatever issues arise as a result. Otherwise, the only error that seems to be showing up relates to the AWS key, and after this is merged I can fix that.

kyleabeauchamp:
Do it

jchodera:
I can drop by and help tinker with the AWS console if needed.

Issue #131, opened 2014-11-25
Number of iterations?


kyleabeauchamp:
Should it be possible to adjust the number of iterations after setup?  The command line inputs suggest it's allowed, but it doesn't seem to work:

```
kyleb@src:~/src/choderalab/yank/examples/host-guest-implicit$ yank run --store=output --verbose --iterations=600
[...]

Iteration 812 / 1000
```

This was allegedly fixed by #125, but I'm still having issues.

jchodera:
I think this is fixed now. Reopen if not.

pgrinaway:
I'm checking right now

pgrinaway:
Ok, so if I do:
```
yank setup binding amber --setupdir=setup --ligname=MOL --store=output --iterations=1000 \
--restraints=harmonic --gbsa=OBC2 --temperature=300*kelvin --verbose

yank run --iterations=800 --store=output --verbose
```

It indicates that it's running 1000 iterations. We can either just change the docs to have the setup set the number of iterations or figure out what's going on here.

pgrinaway:
(also, I just cloned the repo 20 min ago)

pgrinaway:
Ok, I had thought my example didn't work because I specified a smaller number of iterations in the `yank run` command, but after trying this:

```
yank setup binding amber --setupdir=setup --ligname=MOL --iterations=10 --store=output \ 
--restraints=harmonic --gbsa=OBC2 --temperature=300*kelvin --verbose

yank run --iterations=80 --store=output --verbose
```

It only runs 10 iterations.

jchodera:
Addressed by #195.

PR #195, opened 2015-01-22
Bugfix for number of iterations specified via command line

jchodera:
Addresses #131. Still testing.


jchodera:
I think this is working now.

jchodera:
All tests pass on my machine. Merging.

Issue #108, opened 2014-11-07
Report on platform being used


jchodera:
In verbose mode, we should report on the OpenMM platform being used.

jchodera:
Addressed by #197.

PR #197, opened 2015-01-22
[WIP] Print OpenMM platform being used if no platform is specified

jchodera:
Addresses #108


jchodera:
All tests pass locally too.

Issue #198, opened 2015-01-23
Debug why platform is defaulting to CPU even when CUDA and OpenCL are available


jchodera:
I've just noticed this after implementing #197.

jchodera:
Addressed by #200.

PR #200, opened 2015-01-23
Fastest available platform now chosen by default

jchodera:
This addresses #198.


jchodera:
All tests pass on the cluster.

jchodera:
I added a message to emit a warning if the number of CPU threads is set to 1 because `--mpi` is specified.  Warning will only be produced when `--verbose` is specified.

jchodera:
Tests pass on the cluster.

PR #203, opened 2015-01-23
Add warning if user requests ligand randomization for explicit solvent

jchodera:
The simulation will still run, but a warning will be issued that no ligand randomization will take place.


jchodera:
Seems to be working!

Issue #156, opened 2015-01-12
Online analysis seems to produce exception


pgrinaway:
Raising a new issue instead of tacking on to #81, but when I specify the `-o` option for `yank setup` I get an error about not finding the `.nc` files (though maybe we just don't want to support this option for the setup...). However, when I use it as an option on `yank run` I get this:
```
Traceback (most recent call last):
  File "/Users/grinawap/yanktest2/bin/yank", line 9, in <module>
    load_entry_point('yank==0.9.0', 'console_scripts', 'yank')()
  File "/Users/grinawap/yanktest2/lib/python2.7/site-packages/yank-0.9.0-py2.7.egg/yank/cli.py", line 98, in main
    dispatched = getattr(commands, command).dispatch(args)
  File "/Users/grinawap/yanktest2/lib/python2.7/site-packages/yank-0.9.0-py2.7.egg/yank/commands/run.py", line 61, in dispatch
    yank.run(mpicomm, options=options)
  File "/Users/grinawap/yanktest2/lib/python2.7/site-packages/yank-0.9.0-py2.7.egg/yank/yank.py", line 404, in run
    simulation.run(niterations_to_run=niterations)
  File "/Users/grinawap/yanktest2/lib/python2.7/site-packages/yank-0.9.0-py2.7.egg/yank/repex.py", line 950, in run
    self._analysis()
  File "/Users/grinawap/yanktest2/lib/python2.7/site-packages/yank-0.9.0-py2.7.egg/yank/repex.py", line 2386, in _analysis
    indices = t0 + timeseries.subsampleCorrelatedData(u_n[t0:], g=g) # TODO: This could be computed as part of 'timeseries.detectEquilibration()'
TypeError: unsupported operand type(s) for +: 'int' and 'list'
```
after the second iteration.

jchodera:
We should punt to 2.0 or remove command-line option.

jchodera:
Addressed in #204.

PR #204, opened 2015-01-23
Fix online analysis mode

jchodera:
Addressing #156.


jchodera:
This is now fixed, but there's a boatload of code duplication in the analysis code that we need to pare down in a future PR.  This will require making some decisions about how the analysis infrastructure should work, though.

Also, we should add more tests.

jchodera:
We also probably want to come up with some way to integrate information from solvent and complex calculations to provide true "online" estimates of the total free energy of binding, since the estimates provided here are only for the thermodynamic leg currently being simulated.  That is much less useful.

PR #209, opened 2015-02-02
Updated conda recipe to install examples

jchodera:
This change to the conda dev package installs the examples in `anaconda/share/yank/examples/`, which is the most logical place for them to go.

I'll also update the `omnia` package recipe.


jchodera:
Updated conda recipe in https://github.com/omnia-md/conda-recipes/pull/143

PR #211, opened 2015-02-03
Eliminate old examples, reduce default number of iterations, and ensure example tests run just one iteration

jchodera:
* Eliminated `SAMPL-implicit` and `crown-ether` examples
* Changed default number of iterations to 100
* Ensure example tests run just one iteration


kyleabeauchamp:
LGTM

PR #212, opened 2015-02-04
Fixed issues with docopt defaults causing problems

jchodera:
Still testing and updating examples.


jchodera:
Should be good to merge if tests pass.

jchodera:
I see a number of issues that seem to be due to the specification of default options with `docopt`.  The problem seems to be that specifying a default for an optional argument means that the optional argument ends up being filled in with the default value by default, which defeats the purpose of an optional argument!

jchodera:
There's now an issue with locating where the `examples/` get installed from within the `yank` package:
https://github.com/choderalab/yank/blob/master/Yank/utils.py#L4-L24

These are installed now in `/path/to/user/anaconda/share/yank/examples/`.  Any idea how to figure this out?  Do we assume that `yank` source files are in `/path/to/user/anaconda/lib/python2.7/site-packages/yank-*.egg/` and construct relative paths from there?

kyleabeauchamp:
So the general idea of most of our other packages was to include extra data inside the site-packages directory.  This is hard-coded in most of the `utils.get_filename()` functions that we use in various repositories.  

I think we could either:

1.  Use the egg path and `../../../../../../`
2.  Use environment variables
3.  Duplicate files in both locations.  

So when we last discussed this, my argument for *not* using the `share/yank/examples` files in `nose` tests is that the user-accessible examples are meant to be *edited* by users as they learn how to use Yank.  We don't want user-edited files to be passed along to `nose`.  To me, that was an argument for having *duplicated* examples--one in `site-packages` for `nosetests` and the other in `examples` for user consumption.

jchodera:
Sounds like the simple solution may be to have the same set of examples
installed both in yank-*.egg/ and share/yank.  Perhaps we want to have
setup.py install into the egg, and the conda recipe also install into
share/yank? Or should setup.py handle both?

kyleabeauchamp:
I suppose we should do it all in setup.py.  I don't like the idea of accidentally getting into strange states due to the installation method.

jchodera:
I'm still very confused about the openmm conda package install errors in travis:
```
/home/travis/miniconda/envs/_test/bin/.openmm-post-link.sh: line 32: /home/travis/miniconda/envs/_test/bin/python: No such file or directory
/home/travis/miniconda/envs/_test/bin/.openmm-dev-post-link.sh: line 32: /home/travis/miniconda/envs/_test/bin/python: No such file or directory
```

jchodera:
Weird.  My version of nosetests 1.3.4 has an `--immediate` option:
```
  --immediate           print errors and failures as they happen, as well as
                        at the end
```

jchodera:
Huzzah!

PR #218, opened 2015-03-24
Bugfix for pushing docs to s3

jchodera:


jchodera:
For some reason, the HTML pages aren't being pushed to S3.

jchodera:
Just fixed the bug that @rmcgibbo pointed out in #219.  Will need to merge to see if this works.

Issue #219, opened 2015-03-24
Docs not being pushed to S3


jchodera:
I'm trying to figure out why the docs are not being correctly pushed to the `getyank.org` S3 bucket.

Anybody have any ideas?

rmcgibbo:
From the end of the log on https://travis-ci.org/choderalab/yank/builds/55597090
```
/home/travis/build/choderalab/yank
+python devtools/ci/push-docs-to-s3.py
devtools/ci/index.html -> s3://getyank.org/index.html  [1 of 1]
 81 of 81   100% in    0s  1210.69 B/s  done
Done. Uploaded 81 bytes in 1.0 seconds, 81.00 B/s. Copied 0 files saving 0 bytes transfer.
Done. Your build exited with 0.
```

I would try installing and building the docs on your workstation, setting your AWS credentials as environment variables, and then walking through the  `after_sucess.sh`, and fiddling with it. I suspect you will find a bug in the script.

rmcgibbo:
The problem is that `push_docs_to_s3.py` is trying to find the docs in `docs/_build/html/`, but they're actually in `docs/build/` (not no `html`) subdirectory.

jchodera:
Ah!  Thanks so much!

jchodera:
Addressed in #218.

jchodera:
OK, docs are being pushed, but stylesheet (?) issues.
http://getyank.org/latest/index.html

kyleabeauchamp:
According to my debug console, I get the following error:

```
The stylesheet http://getyank.org/latest/_static/css/theme.css was not loaded because its MIME type, "text/plain", is not "text/css". index.html
Use of getPreventDefault() is deprecated.  Use defaultPrevented instead.
```

jchodera:
Maybe an S3 config issue?  Checking...

kyleabeauchamp:
The file's content looks pretty reasonable.  Here is the diff to the mdtraj / msmb version, which has only slight customization:


```
diff theme.css yank 
1908d1907
< 
2563c2562,2564
< 
---
> .wy-table-responsive table td,.wy-table-responsive table th {
>     white-space:nowrap
> }
3738,3800d3738
< }
< 
< 
< 
< 
< /*
< 
< ==========================================================================
< RTM Customizations
< ==========================================================================
< */
< 
< #references .wy-table-responsive {
<     /* prevent the references section from having extra padding between 
<     entries */
<     margin-bottom:0px;
<     max-width:100%;
[...]

PR #220, opened 2015-03-24
Added option to guess mime types when pushing to S3.

jchodera:
This will hopefully fix S3 pushing of CSS files.


jchodera:
Thanks to @rmcgibbo suggestions in #219

kyleabeauchamp:
+1

PR #239, opened 2015-06-04
Update examples

jchodera:
This is a general clean-up pass for all examples, along with the addition of a nosetest to ensure all examples can run a short number of iterations without exceptions.


jchodera:
Will need to speed up minimization to get these tests to pass.

Issue #237, opened 2015-06-03
Yank analyze.py module. Filename error


nividic:
Hi There,

running the analyse module of Yank in the example folders after the completion of the simulations:

yank anayze -s output

an error is raised. In the created output directory two files are present  ‘complex-explicit.nc’ and ‘solvent-explicit.nc’. Actually, the program is looking in some points for ‘complex.nc’ and ‘solvent.nc’ therefore I ended up with the temporary solution to copy the files with the new filenames and therefore in the output directory I had four files (‘complex-explicit.nc’, ‘complex.nc’, ‘solvent-explicit.nc’, ‘solvent.nc’).  This problem seems to be present also in the anaconda installation of Yank.  
 
Cheers,

Gaetano

jchodera:
I think #241 addresses this.

PR #241, opened 2015-06-04
Faster equilibration and fix for analysis issues

jchodera:
This attempts to speed up the example testing.


jchodera:
I think this addresses #237 as well.  Testing.

jchodera:
Tests pass on our local systems.  Travis unfortunately times out.  Still working on ways to speed this up.

PR #249, opened 2015-06-11
[WIP] Use logging to control verbosity

andrrizzi:
#3 

I added a function in yank.utils that can be used to easily configure the root logger. This function is called in the different dispatch() functions in yank.commands to make the logging command-based. Commands "run" and "prepare" save the log files "run.log" and "prepare.log" in the store folder, the other commands don't save any file.

I'm not sure how to handle the mpi case so this doesn't completely fix the issue. Right now, when the "--mpi" argument is used the logging level is set to logging.INFO and the file is not saved.

I also plan to substitute all the print() functions (at least the ones called during the commands run and prepare) with logger.info() calls. Otherwise those won't be included in the log file.

### DIFF
--- a/Yank/alchemy.py
+++ b/Yank/alchemy.py
@@ -42,9 +42,6 @@ import simtk.unit as unit
 import logging
 logger = logging.getLogger(__name__)
 
-# DEBUG
-logging.basicConfig(level=logging.DEBUG)
-
 #=============================================================================================
 # PARAMETERS
 #=============================================================================================
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -427,9 +427,6 @@ def analyze(source_directory, verbose=False):
        If True, verbose output will be generated.
 
     """
-    # Turn on debug info.
-    # TODO: Control verbosity of logging output using verbose optional flag.
-    logging.basicConfig(level=logging.DEBUG)
 
     # Storage for different phases.
     data = dict()
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -81,20 +81,22 @@ def show_mixing_statistics(ncfile, cutoff=0.05, nequil=0):
             Tij[istate,istate] = 1.0
 
     # Print observed transition probabilities.
-    print "Cumulative symmetrized state mixing transition matrix:"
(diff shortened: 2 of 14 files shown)


jchodera:
This is great!

Is there a way to also output log information to the terminal if desired (e.g. if `--verbose` is specified)?

For the MPI case, what is the main issue here?  Is it that we need to find a way to aggregate logging info from other processes that are not the root process?

kyleabeauchamp:
I may have solved the mpi issue in the repex repo, it might be worth
looking.

andrrizzi:
@jchodera 
Yes, with the current design the log information is always output to the terminal. If --verbose is not specified only the logger.info("Message") (and higher levels such as CRITICAL, ERROR, ...) are printed to the terminal. If --verbose is specified also logger.debug("Message") are printed.

As an option, you can save the log information into a file (together with printing it in the terminal). Here all the messages are saved, regardless of whether --verbose is specified or not.

For the MPI case, yes that's the main issue that I can see. Currently only the root node displays verbose messages.

@kyleabeauchamp 
Thanks! I'll look into that.

jchodera:
This sounds great!  There appear to be lots of travis failures, though---are you still debugging this?

andrrizzi:
Yes, I've notice that. The bug in commits b9bfbc6 and 444cef1 is fixed but Travis build fails because: "Your test run exceeded 50 minutes. One possible solution is to split up your test run."

The master branch is affected by this as well.

jchodera:
Ah!  This isn't a problem then.  If the nose tests pass locally for you, go ahead and merge!

andrrizzi:
Nose tests passed here, I don't think I have the write permissions though so I can't merge the pull request.

For the mpi, right now only the root node prints and saves the log entries and the worker nodes only print warnings and errors. Maybe the integration of the different logs could be opened in a new issue (marked for later than yank 1.0) and #3 can be closed?

kyleabeauchamp:
I merged this PR and also created a team `yank-devs` that has write access to the yank, repex, and openmmtools repositories.  You should be good to merge pull requests on those repos.

jchodera:
Thanks everybody!

PR #250, opened 2015-06-15
[WIP] Abl-imatinib in explicit solvent example

andrrizzi:
With the default nonbonded method (CutoffPeriodic, set by prepare.py:setup_binding_amber) the system explode after very few integration steps (usually less than 10) and the positions go to nan at the first iteration.

It might be worth to review [setup.leap.in](https://github.com/andrrizzi/yank/blob/master/examples/abl-imatinib-explicit/setup/setup.leap.in). It's the first time I used tleap and I may have made some silly mistake. To solvate the complex I used a buffer of TIP3P water of 10 Angstrom (the default cutoff length), I hope that's enough.


andrrizzi:
Nevermind, I accidentally left the --gbsa parameter in the preparation of the calculation. The first 4 iterations run fine now. I'll let the full experiment run tonight.

Should we add a warning when yank detects both a periodic box and an implicit solvent?

jchodera:
> Should we add a warning when yank detects both a periodic box and an implicit solvent?

Great idea!

jchodera:
If this example is ready to merge, you can do the implicit solvent / periodic box check in a separate PR.

jchodera:
Merging this so we can all test!

Issue #252, opened 2015-06-16
Abl-imatinib example in explicit solvent


andrrizzi:
I guess this also is related to #168 .

In this example, with the current default alchemical protocol for explicit solvent and complex the Markov chain is decomposable. The state mixing transition matrix after 1000 iterations for the complex is
```
0     1     2     3     4     5     6     7     8     9    10    11    12
1.000
      0.698 0.250 0.053
      0.250 0.532 0.212
      0.053 0.212 0.683 0.053
                  0.053 0.888 0.052
                        0.052 0.900
                                    0.912
                                          0.937
                                                0.950
                                                      0.915 0.052
                                                      0.052 0.829 0.109
                                                            0.109 0.809 0.074
                                                                  0.074 0.916
```
with protocol
```python
def defaultComplexProtocolExplicit(cls):
    ...
    alchemical_states.append(AlchemicalState(1.00, 1.00, 1.00, 1.)) # fully interacting
    alchemical_states.append(AlchemicalState(1.00, 0.00, 1.00, 1.)) # discharged
    alchemical_states.append(AlchemicalState(1.00, 0.00, 0.95, 1.))
    alchemical_states.append(AlchemicalState(1.00, 0.00, 0.90, 1.))
    ...
    alchemical_states.append(AlchemicalState(1.00, 0.00, 0.00, 1.)) # discharged, LJ annihilated
```
the solvent is in the same situation.

jchodera:
Fixed by #254, though alchemical protocol likely needs more tuning.

jchodera:
The #254 fix also fixes the MPI capabilities, so you can now use parallel MPI runs using the included `run-torque.sh` script.

PR #254, opened 2015-06-16
Fix abl explicit example

jchodera:
This WIP is to fix the abl explicit example.


jchodera:
This attempts to address #252.

jchodera:
I believe things are working now.  Nosetests pass locally.  

Some optimization of the alchemical protocols will be needed for this system, but we can do that separately.

Issue #213, opened 2015-02-10
Set chunksize for netcdf variables


jchodera:
e.g. https://github.com/choderalab/openpathsampling/blob/master/openpathsampling/storage/snapshot_store.py#L371

jchodera:
Updated chunksize pointer:
https://github.com/choderalab/openpathsampling/blob/master/openpathsampling/storage/snapshot_store.py#L177-L192

jchodera:
Addressed by #256.

PR #256, opened 2015-06-17
Add chunksizes and compression to NetCDF storage for repex.

jchodera:
Adding suggested chunk sizes to NetCDF storage should keep file sizes from growing too quickly.

Compression has also been enabled for positions.

This addresses #213.


jchodera:
Local nosetests pass.

PR #258, opened 2015-06-17
Speed up travis-ci

jchodera:
This is an attempt to reduce travis-ci runtime by reducing the default max number of minimization iterations.


jchodera:
We may want to consider using a different minimization scheme altogether, like the [VariableLangevinIntegrator](http://docs.openmm.org/6.2.0/api-python/classsimtk_1_1openmm_1_1openmm_1_1VariableLangevinIntegrator.html)

Issue #161, opened 2015-01-12
Change ligand selection to use MDTraj DSL


jchodera:
We should change from having the user specify the ligand residue name to allow the user to specify the ligand atoms via the MDTraj DSL:
http://mdtraj.org/latest/atom_selection.html

Closing #96 to implement this.

jchodera:
We discussed that it may make sense for @andrrizzi to attempt this.

andrrizzi:
Can we close this? Is there anything else we must implement to address this?

jchodera:
This is done!

I'd love to switch to a PyMOL-like syntax, though, since it could be much more powerful.  But that would involve implementing a PyMOL-like selection syntax in either MDTraj or parmed.

PR #259, opened 2015-06-17
[WIP] Use MDTraj DSL to specify ligand atoms.

andrrizzi:
This address #161 .

I changed the parameter --ligname=RESNAME [default: MOL] to --ligdsl=LIGANDDSL [default: “resname MOL”] that takes a DSL string.

I wanted to call the parameter simply ligand but it’s in conflict with systembuilder --ligand and I got the error "docopt.DocoptLanguageError: --ligand is not a unique prefix: --ligand, --ligand?".

Should I change its name to something else? "--ligatoms=DSLSTRING" ? Or maybe change the name of the parameter for systembuilder?

Once the name is settled, I'll change also the syntax in the documentation, the examples and nosetests.

### DIFF
--- a/Yank/cli.py
+++ b/Yank/cli.py
@@ -27,7 +27,7 @@ Usage:
   yank [-h | --help] [-c | --cite]
   yank selftest [-v | --verbose]
   yank platforms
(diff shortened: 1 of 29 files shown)


jchodera:
Awesome!

jchodera:
We might want to temporarily deprecate the `SystemBuilder` CLI for now, since it's clear we need to rethink it to some degree.

@pgrinaway : What do you think?

andrrizzi:
Nosetests passed locally. There were also a couple of wrong ligand res names in the examples so I fixed them. Now resolving conflicts and merging.

jchodera:
Thanks!

andrrizzi:
Can we close this? Is there anything else that must be implemented to address this?

Issue #255, opened 2015-06-17
Consider using modified HamiltonianExchange that only updates Context parameters


jchodera:
The `copy.deepcopy` of explicit solvent objects is painfully slow, taking ~15 s to copy a single explicit solvent Abl:imatinib system.

This will be sped up somewhat by OpenMM implementing a fix proposed in https://github.com/pandegroup/openmm/issues/542, but we could likely skirt this in Yank alone if we are willing to give up the generality of having each thermodynamic state in `repex` be a separate `System` entity and instead use a single reference `System` with a list of Context parameters to be updated for each thermodynamic state.

This would still give us the generality that we need for alchemical states, but we will have to be a bit more "hacky" in how we support additional `System` objects whose energies are to be computed (e.g. for the [anisotropic dispersion correction](https://github.com/choderalab/yank/issues/50)).

Debugging the current Yank is so slow for explicit solvent systems that I am tempted to say we should implement an `AlchemicalHamiltonianExchange` subclass of `ModifiedHamiltonianExchange` (or mixin) that just overrides the bits needed to do this.

jchodera:
I will likely also change the alchemical state to a `dict` based storage scheme where an arbitrary number of alchemical parameters can be defined and used to construct a protocol, instead of the current class with member variables.  This will help when we implement `yank optimize` to automatically search for optimal paths in alchemical space, since it will be easier to give it a set of dictionary keys and endpoints and have it optimize the paths in between for any number of alchemical parameters using a Monte Carlo search.

jchodera:
This is done!

PR #261, opened 2015-06-17
Added draft of MHC-peptide-TCR setup using Yank Python API

jchodera:
I've used `pdbfixer`, `mdtraj`, and the Yank Python API to set up a MHC-peptide-TCR free energy calculation.

While the `Yank` API is simple, there's a fair amount of work that needs to be done to get to this point, so we should think about what we can do to help make this easier.


andrrizzi:
I believe that utils.config_root_logger() must be called before using [logger.info()](https://github.com/jchodera/yank/blob/fix-abl-example/examples/mhc-peptide-tcr/mhc-peptide-tcr.py#L71), or it won't display anything.

I guess most of the initial code could be handled by the systembuilder when it will be in place?

jchodera:
This should be ready to go once tests pass.

jchodera:
This example uses the Python Yank API to set up the calculation.

The `mhc-peptide-tcr.py` script does the following:
* retrieves the specified PDB file
* discards any chains except the ones you want
* builds in missing atoms and optionally minimize the structure
* for each phase (solvent, complex) solvates the system (implicit or explicit), and extracts atom numbers for receptor, ligand, complex, and solvent
* sets up the Yank storage

Running or resuming the simulation can then be done through the Yank Python API or the command line, which means you can easily run it using MPI.

This example works on explicit solvent as well as implicit solvent, but the explicit solvent setup is currently unusably slow.  We will need to speed up the `deepcopy` of systems or use the Context parameter perturbation approach suggested in #255.

The Yank API turns out to be simple, but there is still a good deal of work pushed onto the user in setting up the phases of the calculation.  I wonder if we can n break some modular helper code out of the CLI to help make this easier.  @andrrizzi suggests that this functionality could be built into a revised `SystemBuilder`.

Should be ready to use, though will obviously need a lot of tweaking.

jchodera:
Hm.  I'm still getting `NaN` energies for this particular example.  Not sure why yet.

jchodera:
OK, that last fix still didn't fix things.  I suspect I'm getting something wrong in determining which atoms are part of the alchemical subset.  Or something!

jchodera:
I've determined that the atom selection for the alchemical region in the driver script isn't working correctly, so the bug must be somewhere in here:
https://github.com/jchodera/yank/blob/fix-abl-example/examples/mhc-peptide-tcr/mhc-peptide-tcr.py#L277-L295

jchodera:
The numerical instabilities all seem to derive from this one line:
https://github.com/choderalab/yank/blob/master/Yank/alchemy.py#L540
```
sterics_energy_expression += "reff_sterics = sigma*((softcore_alpha*(1.-lambda_sterics) + (r/sigma)^6))^(1/6);" # effective softcore distance for sterics
```
This seems to give numerical issues for a system of this size.

Working on a fix.

jchodera:
This is working reasonably well now except for some issues we will need to work around in separate PRs:
* OpenMM/MDTraj/pdbfixer `Topology` objects suffer from a loss of information when being transformed.  As a result, identifying the peptide chain in the `solvent` phase of the calculation **does not currently work**.  I am just testing the `complex` phase in implicit solvent right now.
* Explicit solvent setup is very, very slow, but speed should improve after the proposal in #255 is implement
* There is a bug with `pdbfixer` in adding protons in weird geometries for big systems like this.  I've filed a bug report: https://github.com/pandegroup/pdbfixer/issues/92
* `pdbfixer` also chokes on pdbid `1AO7` because the chains in the `ATOM` records are not fully reflected in the `SEQRES` records.  Bug report is here: https://github.com/pandegroup/pdbfixer/issues/93

Wil merge once nosetests show things are working.

jchodera:
The `VariableLangevinIntegrator` bug seems to be only confined to cases where the forces start out as zero, so we can revert back later.  See [this openmm bug report](https://github.com/pandegroup/openmm/issues/981).

Issue #143, opened 2014-12-13
Allow control over more repex options via command line


jchodera:
We may want a more general way to specify arbitrary `repex` options via the command line.

For example, the user may want to set `nsteps_per_iteration`, or any of the other attributed saved to `metadata`:
```
['collision_rate', 'constraint_tolerance', 'timestep', 'nsteps_per_iteration', 'number_of_iterations', 'equilibration_timestep', 'number_of_equilibration_iterations', 'title', 'minimize\
', 'replica_mixing_scheme', 'online_analysis', 'verbose', 'show_mixing_statistics'] 
```
I'm unsure of the easiest way to allow these to be specified, however, especially since some of them have to be converted to `simtk.unit.Quantity` objects.

pgrinaway:
Just going through and cleaning up the issues to decide what we need to work on for 1.0--is there a consensus here?

kyleabeauchamp:
I think we concluded YAML + 1.0

jchodera:
Sounds good!

andrrizzi:
This has been implemented. Closing.

PR #273, opened 2015-06-26
Updates to number of alchemical states for MHC example

jchodera:
This increases the default number of alchemical states for the complex in implicit solvent to handle the MHC example, though we should really implement the JSON-based specification of alchemical intermediates as suggested in #143


jchodera:
Build failed due to timeouts and an internal nose error, which is weird, but will need to merge this.

PR #274, opened 2015-06-26
Added R-alpha-ionone example from Rosa Luirink

jchodera:
This adds an explicit solvent example for R-alpha-ionone binding to cytochrome P450 from @luirink.


jchodera:
This seems to be working.

PR #275, opened 2015-06-26
Bugfix to analyze.py to enable pymbar3 interoperability

jchodera:
This attempts to gracefully fall back to python3 if python2 is not found.


jchodera:
This seems to work in local tests.

Issue #85, opened 2014-08-27
Add additional "slots" for System objects for which energies will computed by no exchanges are allowed


jchodera:
In order to implement the heterogeneous dispersion correction for explicitly solvated systems, we need to add support for additional fully-interacting OpenMM `System` objects with enlarged cutoffs for which potential energies will be computed each iteration but no exchanges allowed to take place.

Lnaden:
I closed this since the issue put forward here is addressed by #483, however, if that is not enough, feel free to reopen

PR #277, opened 2015-07-04
Major YANK speedup

jchodera:
This is the beginning of a PR to increase both startup time and throughput, and to decrease memory footprint, by caching a single `Context` object containing context parameters to switch between alchemical states.  This will make many things faster and easier.

So far, I've replaced the `AlchemicalState` object with a new one that subclasses `dict` so that we can have a one-to-one correspondence between `Context` parameters and alchemical parameters.  This will make these easier to extend in future.


jchodera:
This seems to be working on some examples now!

Need to do more thorough testing to make sure it passes all tests before committing, but major infrastructure changes (except for addressing #85) are done.

jchodera:
Startup time is now ~20 s for p-xylene-implicit and ~55 s for p-xylene-explicit!  I can probably still shave a bit more time off.

jchodera:
All tests seem to pass!  Yay!

PR #279, opened 2015-07-07
Human serum albumin binding to aspirin example

jchodera:
Still working out the bugs here.


jchodera:
This appears to be working now.  Will wait for tests to merge.

Note that I am uncertain about the appropriate protonation or charge state of aspirin to HSA.  Will need to look into this.

jchodera:
All tests pass, so merging!

PR #282, opened 2015-07-08
Initial steps for gromacs input support

jchodera:
Starting work on gromacs support via `yank prepare binding gromacs`.

Initial import of gromacs example files thanks to Davide Provasi (Filizola lab).


jchodera:
This won't fully be working until there is mdtraj DSL support for gromacs residue names: https://github.com/mdtraj/mdtraj/issues/892

jchodera:
Merging this; will take care of any other fixes needed later.

PR #301, opened 2015-08-23
Adding additional tests of binding free energies

jchodera:
Adding more tests of free energy calculations.


jchodera:
I still need to add the standard state correction here.

jchodera:
Nearly all tests pass on the GPU cluster.  I'm not sure why travis is having so much trouble, but I'll debug in a separate PR.  We really need to speed up or eliminate the tests of examples in order to get the travis test times down to a reasonable time.

PR #316, opened 2015-09-22
Factor out alchemy.py to external module

jchodera:
`alchemy.py` has now been factored out as a separate module:
http://github.com/choderalab/alchemy


jchodera:
OK, I think this is finally working!

jchodera:
I think my mucking with root logger stuff is still a bit funky, but we can fix that in a separate PR.

PR #318, opened 2015-10-06
First draft YANK YAML format

andrrizzi:
See [examples/yank.yaml](https://github.com/andrrizzi/yank/blob/pipeline/examples/yank.yaml) for proposed example of YAML format.


andrrizzi:
I implemented a class that "unfolds" the combinatorial tree we talked about. Also I tried to come up with a design that neatly describes different types of free energy calculation. I think something like this would be quite powerful and simple enough (unless I forgot about some of the use cases we want to cover).

```yaml
options:
  yank:
    ..
  repex:
    ..
  alchemy:
    annihilate_sterics: False
    annihilate_electrostatics: True

# The names in components, alchemical_transformation and alchemical_atoms
# are searched both in molecules and in experiment.components. Pre-set
# phases don't have to be explicitly defined and can be overwritten by
# the user. For pre-set phases alchemical_atoms and alchemical_protocol
# assume their default values.
phases:
  # Pre-set phases equivalents, these don't actually have to be defined
  # -------------------------------------------------------------------
  absolute_binding: &default_absolute_binding_phase
    components: [solvent, receptor, molecule]
    alchemical_transformation: molecule->null
  absolute_solvation: &default_absolute_solvation_phase
    components: [solvent, molecule]
    alchemical_transformation: molecule->null
  relative_binding: &default_relative_binding_phase
    components: [solvent, static, molecule1]
    alchemical_transformation: molecule1->molecule2
  relative_solvation: &default_relative_solvation_phase
    components: [solvent, molecule1]
    alchemical_transformation: molecule1->molecule2
  phase_change: &default_phase_change_phase
    components: [solvent1, molecule]
    alchemical_transformation: solvent1->solvent2

  # User-defined phases examples
  # ----------------------------
  bosutinib-afatinib:
    components: [vacuum, abl-2hyy, bosutinib]
    alchemical_transformation: bosutinib->afatinib
    alchemical_atoms: auto
  imatinib-bosutinib:
    components: [vacuum, abl-2hyy, imatinib]
    alchemical_transformation: imatinib->bosutinib
    alchemical_atoms:
      dummy_atoms: [imatinib12, imatinib13, bosutinb14]
      transform_atoms: [imatinib15->bosutinib16, imatinib17->bosutinib18]
    alchemical_protocol: 20states_protocol

  # User-defined that "extend" pre-set phases using anchors
  # -------------------------------------------------------
  relative_complex1000:
    <<: *default_relative_binding_phase
    alchemical_protocol: 20states_protocol
    phase_options:
      repex.niterations: 1000
  relative_complex2000:
    <<: *default_relative_binding_phase
    alchemical_protocol: 20states_protocol
    phase_options:
      repex.niterations: 2000
  relative_solvent:
    <<: *default_relative_solvation_phase
    phase_options:
      repex.niterations: 200

# An example of experiment that uses only pre-set phases. The default alchemical
# factory can be easily inferred from the phase and appropriate default protocol
# can be used. In the future the user should be able to set its own alchemy class
# alchemical_factory: mymodule.MyAbsoluteAlchemicalFactory
basic_experiment:
  components:
    receptor: [abl-2hyy, src]
    ligand: [imatinib, afatinib, bosutinib]
    solvent: [PMEtip3p, RFtip3p, PMEtip4p, GBSAobc1, GBSAobc2]
  phases: [absolute_binding, absolute_solvation]

# An example of experiment with user-defined phases. The section "components" is not
# needed here since the user-defined phases use directly molecules
relative_experiment:
  hmr: [yes, no]
  phases: [imatinib-bosutinib, bosutinib-afatinib]

# A similar example that uses the section "components"
customized_experiment:
  components:
    static: abl-2hyy
    molecule1: imatinib
    molecule2: [bosutinib, afatinib]
    solvent: [PMEtip3p, RFtip3p, PMEtip4p, GBSAobc1, GBSAobc2]
  phases: [relative_complex1000, relative_complex2000, relative_solvent]

# An example of experiment with protein mutations using pre-set phases.
mutations_experiment:
  components:
    static: imatinib
    molecule1: abl-2hyy
    molecule2: [abl-2hyy:VAL31ALA, abl-2hyy:ASP137ASH]
    solvent: vacuum
  phases: [relative_binding, relative_solvation]
```

jchodera:
This is awesome!  Is the YAML file you include in the comment also a file in the repo that we can make line-by-line comments on?

andrrizzi:
Now there is. I figured that if we need to comment the previous version of examples/yank.yaml we should be able to refer to the lines of a specific commit.

jchodera:
Hm.  Can you actually rename the file to something like `yank2.yaml`?  I think it will be easier to comment if it shows all of the lines as added changes.

andrrizzi:
Sure I can. Isn't it easier to comment on something like [this](https://github.com/andrrizzi/yank/blob/9154fba2febe55428c5c93dfbc7052d878b926ee/examples/yank.yaml#L87-L104) though? That link refers permanently to the the last commit. You can also find the version we discussed the last time [here](https://github.com/andrrizzi/yank/blob/07bf13520dbc4d6aef0d26769d92659175c889a7/examples/yank.yaml#L72-L76). Otherwise I'll create a new file.

jchodera:
Good point.  I'll just highlight line groups and make links, like you did.  Will do that now.

jchodera:
## Overall summary
This is great!  I think we can specify a very flexible range of experiments this way.

## Things to fix
### Specification of phases
Currently, the [specification of phases using pre-set phases](https://github.com/andrrizzi/yank/blob/9154fba2febe55428c5c93dfbc7052d878b926ee/examples/yank.yaml#L90-L104) looks like this:
```YAML
phases:
  # Pre-set phases equivalents, these don't actually have to be defined
  # -------------------------------------------------------------------
  absolute_binding: &default_absolute_binding_phase
    components: [solvent, receptor, molecule]
    alchemical_transformation: molecule->null
  absolute_solvation: &default_absolute_solvation_phase
    components: [solvent, molecule]
    alchemical_transformation: molecule->null
```
It's important to remember that a binding free energy calculation actually requires *two* separate `phase`s: (1) the ligand in complex, and (2) the ligand in solvent. Similarly, a solvation free energy calculation requires two phases: (1) ligand in solvent, and (2) ligand in gas phase.  

We will also need to specify either a hardcoded calculation type (e.g. `absolute-binding` which needs a `complex` and `solvent` phase; `solvation` which needs a `solvent` and `vacuum` phase) or a flexible way to specify how various phases are to be combined (along with standard state corrections) to compute free energies.

This idea can be refined, but here's a stab at this:
```YAML
# Define alchemical protocols.
alchemical_protocols:
  complex_protocol: &complex_alchemical_protocol
    factory: AbsoluteAlchemicalFactory
    selection: ligand
    lambda_electrostatics: [1.0, 0.9, 0.8, ...]
    lambda_sterics: [1.0, 0.9, 0.8, ...]
  solvent_protocol: &solvent_alchemical_protocol
    factory: AbsoluteAlchemicalFactory
    selection: ligand
    lambda_electrostatics: [1.0, 0.95, 0.9, ...]
    lambda_sterics: [1.0, 0.95, 0.9, ...]
  vacuum_protocol: &vacuum_alchemical_protocol
    factory: AbsoluteAlchemicalFactory
    selection: ligand
    lambda_electrostatics: [1.0, 0.95, 0.9, ...]
    lambda_sterics: [1.0, 0.95, 0.9, ...]

# Define some default alchemical protcols
protocols:
  # Absolute binding free energy calculation
  absolute_binding: &default_absolute_binding_protocol
    type: absolute-binding    
    phases: 
      complex:
        components: [solvent, receptor, ligand]
        alchemy: 
          <<: *complex_alchemical_protocol
      solvent:
        components: [solvent, ligand]
        alchemy: 
          <<: *solvent_alchemical_protocol
  # Solvation free energy calculation
  solvation: &default_solvation_protocol
    type: solvation
    phases: 
      solvent:
        components: [solvent, ligand]
        alchemy: 
          <<: *solvent_alchemical_protocol
      vacuum:
        components: [ligand]
        alchemy: 
          <<: *vacuum_alchemical_protocol
  # Partitioning between two solvent phases
  relative-solvation: &default_relative_solvation_protocol
    type: relative-solvation
    phases: 
      solvent1:
        components: [solvent1, ligand]
        alchemy: 
          <<: *solvent_alchemical_protocol
      solvent2:
        components: [solvent2, ligand]
        alchemy: 
          <<: *solvent_alchemical_protocol
```
I don't quite know how relative calculations would work out yet.  The difficulty there is that a `RelativeAlchemicalFactory` would need to take in a base `system` along with initial and final ligands and then create a new `system` with an alchemical parameter to transform between them, which is a bit different than first creating a system that is a union of a few components and then operating on this fixed system.

Will edit this to continue...

andrrizzi:
With the design I had in mind, an absolute binding calculation will be something like this
```yaml
phases:
  absolute_binding: &default_absolute_binding_phase
    components: [solvent, receptor, molecule]
    alchemical_transformation: molecule->null
  absolute_solvation: &default_absolute_solvation_phase
    components: [solvent, molecule]
    alchemical_transformation: molecule->null

basic_experiment:
  components:
    receptor: abl-2hyy
    ligand: imatinib
    solvent: PMEtip3p
  phases: [absolute_binding, absolute_solvation]
```
I guess the names of the defined phases are a little misleading, sorry about that. ```absolute_binding``` and ```absolute_solvation``` should be called ```complex``` and ```solvent```. But the experiment will effectively have two phases. Am I missing something?

jchodera:
> I guess the names of the defined phases are a little misleading, sorry about that. absolute_binding and absolute_solvation should be called complex and solvent. But the experiment will effectively have two phases. Am I missing something?

I should have clarified that the predominant issue was really one of naming, rather than of structure.  But we still need to specify that the calculation is an absolute binding free energy calculation somehow.  Something as simple as this might work:
```YAML
phases:
  complex: &default_absolute_complex_phase
    components: [solvent, receptor, molecule]
    alchemical_transformation: molecule->null
  solvent: &default_absolute_solvent_phase
    components: [solvent, molecule]
    alchemical_transformation: molecule->null

basic_experiment:
  components:
    receptor: abl-2hyy
    ligand: imatinib
    solvent: PMEtip3p
  type: absolute-binding
  phases: [complex, solvent]
```

jchodera:
My alternative was to suggest a grouping of phases into a protocol:
```YAML
# Define some default alchemical protcols
protocols:
  # Absolute binding free energy calculation
  absolute_binding: &default_absolute_binding_protocol
    type: absolute-binding    
    phases: 
      complex:
        components: [solvent, receptor, ligand]
        alchemy: 
          <<: *complex_alchemical_protocol
      solvent:
        components: [solvent, ligand]
        alchemy: 
          <<: *solvent_alchemical_protocol
```
which could then be included into your experiment:
```
basic_experiment:
  components:
    receptor: abl-2hyy
    ligand: imatinib
    solvent: PMEtip3p
  << *absolute_binding
```

jchodera:
This was also an attempt to specify the alchemical transformations in a bit more specific way.  Instead of
```YAML
   alchemical_transformation: molecule->null
```
we need to know what factory to use and what parameters to feed the factory, such as
```YAML
  complex_protocol: &complex_alchemical_protocol
    factory: AbsoluteAlchemicalFactory
    selection: ligand
    lambda_electrostatics: [1.0, 0.9, 0.8, ...]
    lambda_sterics: [1.0, 0.9, 0.8, ...]
```

andrrizzi:
Oh, now I understand. Good point. So something like ```basic_experiment.type``` would be necessary.

I love the idea of grouping phases together in protocols. I think it would make it clearer for the user. I wonder if there are use cases in which this will introduce a lot of redundance during the phases definition though.

For the alchemical factory, I thought that a default alchemical factory could have been inferred from the phases description (together with ```basic_experiment.type```), and that a particular (or user-defined) ```alchemical_factory``` could be specified in the ```experiment``` section since it will probably be constant for the whole experiment. Something like
```yaml
basic_experiment:
  components:
    receptor: abl-2hyy
    ligand: imatinib
    solvent: PMEtip3p
  type: absolute-binding
  factory: userModule.MyAlchemicalFactory
  phases: [absolute_binding, absolute_solvation]
```

For the alchemical protocols, I also think it's a good idea to specify them in the phase [like here](https://github.com/andrrizzi/yank/blob/9154fba2febe55428c5c93dfbc7052d878b926ee/examples/yank.yaml#L118). I thought that the default protocol of the alchemical factory could be applied when it was not explicitlyl specified.

For the system building, I imagined the ```YamlBuilder``` class parsing something like this
```yaml
phases:
  bosutinib-afatinib:
    components: [vacuum, abl-2hyy, bosutinib]
    alchemical_transformation: bosutinib->afatinib
```
and inferring that it has to build the first system from the vacuum solvent, abl-2hyy, and bosutinib and the second one from vacuum, abl-2hyy and afatinib. But I probably don't have enough experience to understand what are the real problems in implementing this.

jchodera:
> I love the idea of grouping phases together in protocols. I think it would make it clearer for the user. I wonder if there are use cases in which this will introduce a lot of redundance during the phases definition though.

I think the redundancy can be eliminated using your anchor trick:
```YAML
# Define some default alchemical protcols
protocols:
  # Absolute binding free energy calculation
  absolute_binding: &default_absolute_binding_protocol
    type: absolute-binding    
    phases: 
      <<: *absolute_complex_phase
      <<: *absolute_solvent_phase
```

> For the alchemical protocols, I also think it's a good idea to specify them in the phase like here. 
```YAML

  # User-defined phases examples
  # ----------------------------
  bosutinib-afatinib:
    components: [vacuum, abl-2hyy, bosutinib]
    alchemical_transformation: bosutinib->afatinib
    alchemical_atoms: auto
  imatinib-bosutinib:
    components: [vacuum, abl-2hyy, imatinib]
    alchemical_transformation: imatinib->bosutinib
    alchemical_atoms:
      dummy_atoms: [imatinib12, imatinib13, bosutinb14]
      transform_atoms: [imatinib15->bosutinib16, imatinib17->bosutinib18]
    alchemical_protocol: 20states_protocol
```
I guess I hadn't read this closely enough.

> I thought that the default protocol of the alchemical factory could be applied when it was not explicitlyl specified.

I'd like to try to get away from having default alchemical protocols specified in the code, since this is so highly system-dependent.  We can move toward having some sort of auto-tuning method in the code to select an guess at an alchemical protocol automatically, but I think a default protocol schedule really doesn't belong in the code.

> For the alchemical factory, I thought that a default alchemical factory could have been inferred from the phases description (together with basic_experiment.type), and that a particular (or user-defined) alchemical_factory could be specified in the experiment section since it will probably be constant for the whole experiment.

I'm just as happy with this and the other proposals.  I think the main thing we should think about now is what choice would make the underlying implementation as simple as possible and yet give us as much flexibility as we might need at the YAML level.

For example, it would be great if we could reuse a simple idiom where we simply unpack arguments at each stage, like this:
```python
options = yaml.load(filename)
args = options['yank']
yank = Yank(*args)
```
and within `Yank`:
```python
# Build phases
for phase in phases:
   args = phases[phase]
   self._create_phase(*args)
```
We don't necessarily have to make the YAML exactly match the existing code, but if we can at least make it easy to adapt the code to parse the YAML without a huge amount of extra special-case coding, that will make everything much easier!

andrrizzi:
All great points, I agree with everything you pointed out.
> I think the main thing we should think about now is what choice would make the underlying implementation as simple as possible

Absolutely, I didn't think of it that way. Then, with the implementation I have in mind, grouping phases into protocols like you proposed is probably the way to go.

jchodera:
OK, this is great progress.  Feel free to merge if we want to continue moving forward here.

The next steps would be to try to solidify a "first pass" at a functional YAML processing strategy, which I think means reorganizing the YAML structure a bit to better match the call structure for YANK.

In pseudocode, this is roughly something like this (though we may need to refine this idea):
* Unroll the YAML file using the scheme you've created.  We'll need some way to generate a unique directory for each YANK calculation that is run.  Should we try to create a clever concatenation of identifiers for system component names and experiments, or should we just use a unique serial directory name and record the combination of directory name and unrolled options (as YAML files) in an output directory?
* Process the `setup:` block (or some name we decide) to direct the preparation of various system components using pipeline tools in `openmoltools`
* Create a `thermodynamic_state` from temperature and pressure specified in some block
* Process the `phases:` block to set up create `system` objects for each of the phases, compute `atom_indices` for each component
* Initialize `yank = Yank()` and call `yank.create()` with the options in the `yank:` block passed as the `options` argument, and the other arguments (`phases`, `systems`, `positions`, `atom_indices`, `thermodynamic_state`) passed
* `Yank` then calls `create` on each phase, passing the phase options along in the proper nested fashion.  We'll need to make sure to think about this some more to maximize our use of simple argument unrolling idioms like
```YAML
for phase in phases:
   args = phases[phase]
   self._create_phase(*args)
```

andrrizzi:
Great! I'll merge this and start playing with the code. Hopefully this way I'll have a better picture of what needs to be tweaked to tackle those next steps.

jchodera:
Thanks!

PR #319, opened 2015-10-15
Slack comments

bas-rustenburg:
Experimenting with travis notifications for our slack page.  Would this be interesting?

Sorry for the messy looking commit. The travis command line application cleaned up the indentations when I added the encryption token.


jchodera:
Is this still useful @bas-rustenburg ?

bas-rustenburg:
You can close this. I don't think it'll be productive to fix the conflicts that have emerged in time.

Issue #314, opened 2015-09-17
Simplifying options handling


jchodera:
The YAML framework introduced by #310 is fantastic, but it seems like we might be able to simplify how we handle options to make things a little simpler.  The `yank.util.YankOptions` object, for example, currently stores 
* default options
* CLI options
* YAML options
* runtime options

and has to correctly juggle precedence among all of these.

Is there some way we can refine this somehow by allowing options to be specified by either YAML, command-line, or default initialization and just making sure to read these in the correct order?

 Perhaps we can brainstorm how each module can keep a central list of options that it wants to manage, while still allowing us to validate that the options specified by the user are valid and have the correct units.  Is there a simple way to do this?

bas-rustenburg:
How about collecting every set of options separately into `dict`s and then combining those in one, consistent order of preference that we can also add to the documentation:

Say, 
* start with default options dictionary for every module,
* override keys by yaml options that were added,
* replace keys by cli options,

not sure what "runtime" options means. Can you change options during the middle of a run?

Then, at the end, validate the `dict` to make sure options are correct and compatible.

You can store the options as a yaml input file, similar to what gromacs does, if you want to manually  validate, or easily rerun with the same options.

andrrizzi:
I agree that the options in Yank aren't currently handled very well. I also don't like too much that we pass the options as an argument to each function that needs them. I'm not sure though how the independent repex and alchemy repos will handle options when they will be integrated, so we should probably come up with a design that can be easily shared among all the repos.

Keeping the runtime and YAML options separated may be useful for people that want to use Yank through its API. For example, say you want to run multiple Yank experiments from a single script with different options. You can use the API to load a different YAML file for each experiment, or start from a basic YAML file and hardcode the few options that you want to tweak for each experiment. To do this, Yank needs to remember which were the default options and which the ones given through YAML. I agree that we can merge default and command line options while parsing them since they won't change during a single run, but I would keep YAML and runtime options in a separated dictionary in memory and handle their precedence.

bas-rustenburg:
> but I would keep YAML and runtime options in a separated dictionary in memory and handle their precedence.

Could you clarify what you mean by runtime options? Does yank take input from `stdin` during the run?

andrrizzi:
Sure, sorry it's not clear, I couldn't find a better name. By runtime options I mean those that are hardcoded in your Python script (and that are not default options). For example, if you want to check the effect of the integration time step on the simulation, you can do something like
```python
[...]

PR #320, opened 2015-10-15
Second draft YAML syntax

andrrizzi:
I figured it would be clearer to discuss few small use cases rather than a single monolithic one so I created a YAML cookbook.

Here's a possible pseudocode that matches the info in the YAML file to the code steps.

1. Unroll the experiment section of the YAML file
2. Create appropriate output folder(s).
  * A possibility here is to create a folder for each experiment and a subfolder (cleverly named) for each combination. The general molecule files created during setup should be kept in a single folder to avoid creating the same files in multiple places during combinatorial experiments.
3. Match the components described in the phases (ligand, receptor, imatinib, PMEtip3p)
  * Check both in section `molecules` and in `experiment.components`
  * Output a single experiment where molecules/solvents have been uniquely identified
4. Create a reference system for each phase
  * If the user has specified the prmtop/inpcrd (or gro/top) files we can use directly OpenMM
  * Otherwise parametrize, charge the molecule with `openmmtools` as specified in `molecule.parameters` and/or `molecule.charge`
  * Create the systems as specified in `experiment.setup` (e.g. `leap` or `systembuilder`)
  * In the future, at this point we could generate protonation states, add mutations and perform loop refinement.
5. Compute `atom_indices` as specified in `molecule.select`. If we built the system from leap or systembuilder we should already know how to select them.
6. For relative calculations we need to load the final ligand specified in `alchemical_transformation:ligand1->ligand2`.
7. Create thermodynamic state according to `experiment.thermodynamics`
8. Inizialize `Yank` and configure it with the alchemical factory object specified in `experiment.alchemical_factory`.
9. Call `yank.create()` with YAML `options` and `phases`, `systems`, `positions`, `atom_indices`, `thermodynamic_states` created. Setup the platform, precision, mpi, verbose according to `options.yamlbuilder` and call `yank.run()`.

About maximizing the use of unrolling argument idioms in the phase description. This will probably be highly dependent on the future alchemy classes API (which should probably be designed independently), but in general I think a good compromise between YAML clarity and ease of implementation would be to group the API arguments in a single YAML key (here `alchemical_protocol`). Something like
```yaml
phases:
  complex: &complex_phase
    components: [PMEtip3p, abl, imatinib]
    alchemical_transformation: imatinib->bosutinib
    alchemical_protocol:
      lambda_electrostatics: [1.0, 0.9, 0.8, 0.7, 0.6, 0.5, 0.4, 0.3, 0.2, 0.1, 0.0]
      lambda_sterics: [1.0, 0.9, 0.8, 0.7, 0.6, 0.5, 0.4, 0.3, 0.2, 0.1, 0.0]
      molecule1_dummy_atoms_indices: [12, 13]  # imatinib dummy atoms
      molecule2_dummy_atoms_indices: [14]  # bosutinib dummy atoms
      transform_atoms_indices: [[15, 16], [16, 17]]  # couples of imatinib-bosutinib atoms
      any_other_API_parameter: value
```
The code then would look like
```python
experiments = CombinatorialTree(yaml.load('yank.yaml'))
for exp in experiments:
    # Do steps 2-7
    # To create the systems and load the final molecule object in relative calculations, we will
    # have to parse and process phase.components and phase.alchemical_transformation here
    yank = Yank(exp['options'])
    AlchemyClass = getattr('alchemymodule', exp['alchemical_factory'])
    yank.alchemy = AlchemyClass()

    # Isolate the 'alchemical_protocol' dictionary to pass to Yank.create()
    phases = {phase_name: phase['alchemical_protocol'] for phase_name, phase in exp['phases'].iteritems()}
    yank.create(systems, positions, ..., phases)
```
This way we could easily modify YAML syntax and API without changing the parsing code.

For now, I don't think we need to be specific about the syntax details for features that haven't been implemented yet in Yank, as long as we find a general structure that we like.

### DIFF
--- a/Yank/commands/prepare.py
+++ b/Yank/commands/prepare.py
@@ -464,14 +464,12 @@ def dispatch_binding(args):
             else:
                 raise Exception("Platform selection logic is outdated and needs to be updated to add platform '%s'." % platform_name)
 
-    yank.options.cli = options
-
-    # Parse YAML configuration file and create YankOptions object
+    # Parse YAML options, CLI options have priority
     if args['--yaml']:
-        yank.options.yaml = YamlBuilder(args['--yaml']).options
+        options.update(YamlBuilder(args['--yaml']).options)
 
     # Create new simulation.
-    yank.create(phases, systems, positions, atom_indices, thermodynamic_state)
+    yank.create(phases, systems, positions, atom_indices, thermodynamic_state, options=options)
 
     # Report success.
     return True
--- a/Yank/yank.py
+++ b/Yank/yank.py
@@ -31,7 +31,6 @@ import simtk.openmm as openmm
 
 from . import sampling, repex, alchemy
 
-from utils import YankOptions
 from alchemy import AbsoluteAlchemicalFactory
 from repex import ThermodynamicState
 from sampling import ModifiedHamiltonianExchange # TODO: Modify to 'from yank.sampling import ModifiedHamiltonianExchange'?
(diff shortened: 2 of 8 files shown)


andrrizzi:
Also, any idea on why conda installation is broken?

If there are no objections, I would merge this and proceed to implement in a separate pull request the options interface that we discussed in #314. I'd like to have a somewhat stable options API before implementing the YAML parsers.

jchodera:
Go for it!

I can fix the Travis issues in a separate PR.

Please go ahead with the options stuff!

J

jchodera:
Do you mind if I make some tweaks to the cookbook recipes in a separate PR?

andrrizzi:
Absolutely, go ahead. I don't think I'm going to touch them in the near future. I'm just waiting for results of local tests before merging.

Issue #324, opened 2015-11-06
Restraint type and randomize ligand don't work


andrrizzi:
I searched through the whole project but I can't find any place in which the options ``'restraint_type'`` and ``'randomize_ligand'`` (which are set in ``prepare.py``) are used. As a consequence, they always assume the default values set in ``Yank.__init__()`` (``self.randomize_ligand = False`` and ``self.restraint_type = 'flat-bottom'``).

Is this a bug or am I missing something?

andrrizzi:
This should be fixed now, closing.

jchodera:
Thanks!

PR #325, opened 2015-11-10
Implemented new options API

andrrizzi:
Summary:
* Now ``Yank`` and ``ReplicaExchange`` both have a class variable ``default_parameters`` that specifies their options with their respective default value. These parameters are stored as object attributes on initialization.
* Options for both ``Yank`` and ``ReplicaExchange`` are passed on ``__init__`` with ``**kwargs``.
* It is not possible to pass repex options through ``Yank.create()`` and ``Yank.run()`` anymore (BREAKS THE API).
* A generic utility function ``validate_parameters()`` can be used to infer the expected parameter types from the default values specified by the classes, and validate/convert them appropriately, for example
```python
>>> parameters = {  # these could be obtained from YAML options
>>>        'show_energies': True,
>>>        'number_of_iterations': 1.1,
>>>        'timestep': '2.0*femtosecond'
>>>    }
>>> parameters = utils.validate_parameters(parameters, ReplicaExchange.default_parameters,
>>>                           check_unknown=True, process_units_str=True, float_to_int=True)
>>> parameters
... {'show_energies': True, 'number_of_iterations': 1, 'timestep': Quantity(value=2.0, unit=femtosecond)}
```

Local tests almost pass, but the error doesn't seem related to this pull request.

### DIFF
--- a/Yank/commands/prepare.py
+++ b/Yank/commands/prepare.py
@@ -418,9 +418,6 @@ def dispatch_binding(args):
         if phase == 'complex-explicit':
             logger.info("solvent and ions : %9d" % len(atom_indices[phase]['solvent']))
 
-    # Initialize YANK object.
-    yank = Yank(store_dir)
-
     # Set options.
     options = dict()
     if args['--nsteps']:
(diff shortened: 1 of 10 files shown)


andrrizzi:
This fixes #324 and addresses #314. If there are no objections, I would merge this and implement the YAML parsing part in a different PR.

jchodera:
Great!

Does it do units checking as well?

andrrizzi:
Yes! [Here](https://github.com/andrrizzi/yank/blob/options-api/Yank/utils.py#L505-L506).

jchodera:
Awesome!

jchodera:
Feel free to merge if tests pass locally!

andrrizzi:
Merging then.

PR #327, opened 2015-11-17
Fix storage of scalar options in NetCDF and --randomize-ligand option

jchodera:
This fixes the storage of scalar options dictionaries in the NetCDF storage, which were not properly being converted back to scalar types.

Also fixes `--randomize-ligand`.


steven-albanese:
Is this good to be merged?

andrrizzi:
Looks great!

jchodera:
Yes!

PR #329, opened 2015-11-19
First implementation of YAML interface

andrrizzi:
What's implemented here
* setup of the simulation through the path (openeye ->) (epik ->) antechamber -> leap.
* support for IUPAC name, SMILES, mol2 and pdb formats
* molecules generated by OpenEye whose atoms overlap with other molecules are fixed
* options and syntax validation
* experiments over combinations of receptors/ligands/solvents/options
* generation of YAML file for each experiment combination for reproducibility

What is NOT implemented here
* alchemy stuff
* support for multiple-molecule files
* start simulation directly from amber/gromacs file

I still have to add documentation and refactor some redundant parts of the code, but this is should be mostly done.

Also I have implemented in [this section of utils.py](https://github.com/andrrizzi/yank/blob/yaml-builder-basics/Yank/utils.py#L667-L669) few things that I think would be better placed in openmoltools when they are stable. There is a low-level TLeap interface and few functions to deal with Schrodinger's tools.


andrrizzi:
I think this is ready to be merged.

andrrizzi:
I found a weird import chain when trying to run doctest on Yank/yamlbuild.py

[``yank.yamlbuilder: import openmoltools``](https://github.com/andrrizzi/yank/blob/yaml-builder-basics/Yank/yamlbuild.py#L24) -> [``openmoltools:__init__: from openmoltools import utils``](https://github.com/choderalab/openmoltools/blob/master/openmoltools/__init__.py#L28) -> [``openmoltools.utils: from commands import getoutput``](https://github.com/choderalab/openmoltools/blob/master/openmoltools/utils.py#L18)

The last line search in ``yank.commands`` instead of the [standard library commands](https://docs.python.org/2/library/commands.html) and the test stops. Should we rename the ``yank.commands`` module into something that does not risk to clash with standard libs (in another pull request)?

jchodera:
Looking at this now!

andrrizzi:
Merging now. I'm going to create an issue about the ``commands`` module renaming.

Issue #331, opened 2015-11-28
Looks like there is a test failure in master


jchodera:
There seem to be a few problems in `yamlbuild.py`:
https://travis-ci.org/choderalab/yank#L1532-L1606

andrrizzi:
Yeah, I think it is caused by the openmoltools import of ``commands`` discussed in #330 .

jchodera:
Ah, OK!  Sorry!

PR #333, opened 2015-12-03
Tweaks to YAML interface

andrrizzi:


andrrizzi:
This fixes #331, implements #334, renames the camelCase options into underscore_case, and fix the paths in the YAML file generated for each combination.

jchodera:
This looks great!  I'm not sure if the automatic underscore-to-OpenMM-CamelCase conversion is going to drive us insane or not, but for now, I'm willing to give it a go!

andrrizzi:
Yeah, I figured it could become confusing so I used the function that convert from camel case to underscore only locally when calling ``createSystem()``. Everything is stored and handled as underscore. Values of openmm options are still in camel case, so for example:
```yaml
solvent:
  vacuum:
    nonbonded_method: NoCutoff
```

jchodera:
> Values of openmm options are still in camel case

Sounds good!

andrrizzi:
Merging then.

PR #335, opened 2015-12-14
Specify alchemical protocol in YAML script

### LINKED ISSUE #168: Allow users to specify alchemical protocols in a file that can be read and written by YANK
This will also be useful for `yank reoptimize`

andrrizzi:
This will allow to specify the alchemical protocol in the YAML script.


jchodera:
Woohoo!  

What's the YAML format for this?

andrrizzi:
For now, I was thinking we could settle with something like this
```yaml
protocol:
  absolute-binding1:
    phases:
      complex:
        alchemical_path:
          lambda_electrostatics: [1.0, 0.99, 0.98, 0.96, 0.94, 0.92, 0.9, 0.8, 0.7, ...]
           lambda_sterics: [1.0, 0.99, 0.98, 0.96, 0.94, 0.92, 0.9, 0.8, 0.7, 0.6, …]
        annihilate_electrostatics: yes
        annihilate_sterics: yes
      solvent:
        alchemical_path:
          lambda_electrostatics: [1.0, 0.9, 0.8, 0.7, 0.6, 0.5, 0.4, 0.3, 0.2, 0.1, 0.0]
           lambda_sterics: [1.0, 0.9, 0.8, 0.7, 0.6, 0.5, 0.4, 0.3, 0.2, 0.1, 0.0]
        annihilate_electrostatics: yes
        annihilate_sterics: yes
  absolute-binding1:
    ...


experiments:
  components:
    ...
  protocols: [absolute-binding1, absolute-binding2]  # combinations over alchemical protocols
```
I don't like much so that many levels of indentation but I found it hard to come up with something simpler. If we plan to put everything within ``complex`` and ``solvent`` we could decide to eliminate ``phases`` though.

Also, do we want to support a default ``alchemical_path`` or do we want to make it a mandatory field?

jchodera:
Looks great!

> Also, do we want to support a default alchemical_path or do we want to make it a mandatory field?

Let's make it mandatory.  I'd rather not hard-code the protocol until we have an automated way of selecting protocols adapted to specific problems.  Once we have that feature, we can provide a default that does this automated selection.

andrrizzi:
> Let's make it mandatory.

Ok!

andrrizzi:
``alchemical_path`` is implemented but I couldn't add the support for ``annihilate_electrostatics `` and ``annihilate_sterics`` because I realized there is no way to specify it through the current Yank API. I'll open a separate issue for this. Meanwhile, I think this is good to be merged.

andrrizzi:
Now that I think about it, I could add support for ``annihilate_electrostatics`` and ``annihilate_sterics`` in the general ``options`` section that the user will be able to overwrite in the specific phase in the future. Something like

```yaml
options:
  number_of_iterations: 100
  annihilate_electrostatics: yes
  annihilate_sterics: yes
  ...
protocols:
  absolute-binding:
    phases:
      complex:
        alchemical_path:
          ...
        options:
          number_of_iterations: 1000
          annihilate_electrostatics: no
      solvent:
        ...
```

jchodera:
> Now that I think about it, I could add at least support for annihilate_electrostatics and annihilate_sterics in the general options section that the user will be able to overwrite in the specific phase in the future. 

That would be useful!  Otherwise, there is the potential that we might have different phases set these differently, which would lead to a real error in the binding affinities since the alchemically annihilated states would not be equivalent.

andrrizzi:
Done! This should be ready to be merged.

jchodera:
Hooray!

Can we get a couple of people to quickly look over this?  @pgrinaway and @bas-rustenburg maybe?

jchodera:
This looks great! Let's run with it for now.

andrrizzi:
Ok, merging then.

jchodera:
Thanks!

PR #341, opened 2016-01-13
Suppress Epik output/error messages.

bas-rustenburg:
This suppresses any stderr/stdout messages that epik produces.


andrrizzi:
Nice! Are there errors in epik that we want to catch for which ``check_call()`` does not raise ``CalledProcessError``?

bas-rustenburg:
I haven't extensively tested all that Epik can do, but I could try making epik fail and seeing whether this catches it.

andrrizzi:
I'm checking the [manual](https://isp.ncifcrf.gov/files/isp/uploads/2010/07/epik20_user_manual.pdf). I wonder if the options ``–NO_JOBCONTROL`` does it.

bas-rustenburg:
Good call! I think that might also work, and leaves any other warnings that may be of interest.

bas-rustenburg:
Nicer solution, added in new commit. Good to merge if checks pass.

bas-rustenburg:
Whoops, left something behind. Fixing.

andrrizzi:
Looks good to me.

PR #343, opened 2016-02-05
Pack molecules and multi molecules files

andrrizzi:
This will add support to multi-molecules/models files (pdb, sdf, smiles) and various improvements to molecules packing.

Question: **Do we want starting conditions to be reproducible**? In other words, do we want the algorithm that packs the molecules (i.e. resolve clashing atoms and pull close molecules) to be deterministic? Because if the answer is yes, we cannot randomly propose the ligand's position in the box and I think we're stuck with heuristic translations and rotations. The alternative would be seed always the same number at the beginning of the algorithm.


jchodera:
While we could use a deterministic subrandom or pseudorandom sequence
generator, I don't see the value in having the starting conditions be
deterministic if the remainder of the calculation is also not
deterministic. So random is fine!

andrrizzi:
> I don't see the value in having the starting conditions be deterministic if the remainder of the calculation is also not deterministic

I was thinking more about debugging than reproducible results, but maybe there is not much in there either. Doing it randomly will effectively implement the ``randomize_ligand`` option with explicit solvent as well (although technically it would be outside ``Yank``'s API).

jchodera:
I think `numpy.random` uses a fully deterministic random number generator, so as long as we seed `numpy.random.seed()` with a particular `int` or `int` array seed for tests for all nodes, we should get deterministic output for debugging.

andrrizzi:
> as long as we seed ``numpy.random.seed()`` with a particular ``int`` or ``int`` array seed for tests for all nodes, we should get deterministic output for debugging

Perfect! Is [this algorithm](https://github.com/andrrizzi/yank/blob/pack-multimolecule/Yank/yamlbuild.py#L187-L247) more or less what you had in mind?

jchodera:
I think this looks good!  

Do you want to have a maximum number of iterations it can run before it gives up and throws an exception?

andrrizzi:
Nope! Good idea! I'll set it to 1000.

andrrizzi:
What if instead after 1000 iterations I just increase the maximum distance allowed and try again? This way we could avoid failures and adapt the algorithm when the ligand is bigger than expected.

andrrizzi:
I guess we can actually do both.

andrrizzi:
Based on a quick search, it looks like I cannot parse multiple models in a single ``pdb`` file neither with OpenMM ``PDBFile`` nor with ``tleap``. The same goes for ``mol2`` and ``sdf`` files, so the only solutions I see right now are either doing it manually or introducing a ParmEd dependency in Yank (which would not solve multi-molecules ``sdf`` files anyway since I don't think they're supported).

Two big questions are: How much is this going to be useful? Should I prioritize other features (mutations and chain selection)?

andrrizzi:
Thank you @maxentile for reminding me about ``mdtraj``!

jchodera:
Good point about the OpenMM PDBFile reader limitations, though.  It *writes* multimodel PDB files, so presumably it should read them.  Suggest this on the issue tracker?

andrrizzi:
> Suggest this on the issue tracker?

I will. So I'll use ``mdtraj`` to split ``pdb`` files and I'll write some code myself to split ``mol2`` and ``sdf``.

jchodera:
For Tripos-format `mol2` and `sdf`, we should *always* be using the OpenEye tools.  There's too much weird crap that is needed to manage these files otherwise, and there is no sense in reinventing the wheel.

andrrizzi:
> we should *always* be using the OpenEye tools

Ok!

andrrizzi:
Nevermind about the ``PDBFile`` issue, somehow I ended up on the API documentation of a previous OpenMM version. The current implementation support multi-frame pdb files. My bad.

andrrizzi:
This is ready to be merged. Summary
* random algorithm for ligand packing
* support for ``smiles`` (i.e. ``csv``) files
* support for ``sdf`` files
* support for multi-frame ``pdb`` and ``mol2`` files
* added ``select`` keyword in molecule description

Example of the syntax introduced
```yaml
molecules:
  abl:
    filepath: multi-frame.pdb
    select: 1  # 0-based, select only the second frame
  all-ligands:
    filepath: FDA-approved.csv  # or FDA-approved.smiles
    parameters: antechamber
    select: all  # run one experiment for each molecule in in multi-molecule.sdf
  imatinib:
    filepath: FDA-approved.sdf
    parameters: antechamber
    select: [1, 2]  # extract second and third conformers in FDA-approved.sdf with OpenEye
```

jchodera:
Awesome!  Thanks!  Looks good to me.

andrrizzi:
Great! Merging.

PR #344, opened 2016-02-17
Optional packing and strip off protons

andrrizzi:
This should be ready to be merged.

Add an option to avoid random packing in case one wants to start from the binding pocket. Also, this PR adds a ``strip_protons`` option to remove hydrogens from a PDB before feeding it to ``leap``. Example of the syntax introduced:

```yaml
options:
  pack: no  # default is yes

molecules:
  abl:
    filepath: abl.pdb
    strip_protons: yes  # default is no
```


jchodera:
Nice!

andrrizzi:
One doctest is failing. It looks like with the latest release of ``mdtraj 1.6`` something got broken. When calling ``topology.from_openmm`` I get ``AttributeError: 'Residue' object has no attribute 'segment_id'``. I can't reproduce it locally with ``mdtraj 1.5.1``. I think [the last three commits here](https://github.com/mdtraj/mdtraj/commits/master/mdtraj/core/topology.py) changed something.

Should we fix ``mdtraj`` version to ``1.5.1`` in the conda recipe? And maybe open an issue in ``mdtraj``?

jchodera:
Let's fix `mdtraj` to `1.5.1` as you suggest, and hopefully file a bug report too.

andrrizzi:
This is ready to be merged for me. Summary
* added option to strip ``pdb`` protons before feeding it to ``tleap``
* added option to prevent random packing
* save ``sdf`` file as well when running ``epik`` to save the tags and fix formal charge argument in ``antechamber``.
* fixed ``mdtraj`` version to 1.5.1

jchodera:
Awesome!  Merge when ready!

PR #349, opened 2016-02-26
Fix epik penalties parsing, and resume. More MPI debug messages

andrrizzi:
Various minor fixes:
* Fix epik penalties parsing.
* Fix ``resume_simulation`` when nc files don't exist.
* Non-root nodes do not print debug messages on ``stdout``, but now each non-root node create a separate log file so that we can still keep track of what happens in non-root processes.

The PR should be ready to be merged unless somebody has objections.


jchodera:
Cool!

PR #351, opened 2016-03-01
Always use OpenEye charges instead of sqm when OE toolkit is available

andrrizzi:
As per title. Should be ready to be merged.

### DIFF
--- a/Yank/tests/test_yaml.py
+++ b/Yank/tests/test_yaml.py
@@ -23,6 +23,7 @@ import itertools
 from nose.tools import raises
 
 from yank.yamlbuild import *
+from mdtraj.formats.mol2 import mol2_to_dataframes
 
 #=============================================================================================
 # SUBROUTINES FOR TESTING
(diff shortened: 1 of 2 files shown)


jchodera:
Thanks!

I worry that autodetection of OpenEye and selecting behavior based on this can cause unexpected differences in runs if OpenEye is misconfigured or not installed or the license file is not current.

What about adding an option to the YAML file, or changing the charge model to `openeye-am1bcc` to select this?

andrrizzi:
I see. What if I changed a little the syntax to be
```yaml
molecules:
   benzene:
      name: benzene
      parameters: openeye-am1bcc  # this means openeye charges + antechamber
```
Or maybe it's better ``parameters: openeye-antechamber``?

There is currently no syntax to explicitly separate parametrization of charges from other parameters, but we can add it if that's preferable.

jchodera:
How about openeye:am1bcc-gaff?

andrrizzi:
Sounds great.

andrrizzi:
Done!

jchodera:
Looks good!

PR #353, opened 2016-03-29
Update conda recipe to use -dev versions of omnia packages

jchodera:
This PR updates the conda dev recipe to use the `-dev` versions of omnia packages so that we can test against `openmm-dev` and `alchemy-dev`.

Note that this means the recipe won't build/install on `osx` right now---you have to use `python setup.py install` or `pip install`---but I'm slowing converting all the build systems over to add an `osx` build to travis.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -17,15 +17,15 @@ requirements:
     - scipy
     - setuptools
     - netcdf4
-    - openmm
+    - openmm-dev
     - mdtraj 1.5.1 # TODO remove me when segment_id bug is fixed (this is installed through openmoltools)
     - openmmtools >=0.7.2
-    - pymbar
+    - pymbar-dev
     - ambermini
     - docopt
-    - openmoltools
+    - openmoltools-dev
     - sphinxcontrib-bibtex
-    - alchemy >=1.1.3
+    - alchemy-dev
     #- gcc 4.8.2 # [linux]
     #- gcc 4.8.2 # [osx]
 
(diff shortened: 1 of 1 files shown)


jchodera:
Rats.  This installs both `parmed` and `parmed-dev`, and both `openmm` and `openmm-dev`.

jchodera:
Any idea what is bringing in the `parmed` dependency?

jchodera:
OK, I think I've fixed it.  I'm a little surprised there's still an antechamber failure---I thought the `openmmtools-dev` would have fixed that.

Can you review and merge when ready @andrrizzi?

jchodera:
I see what happened with the antechamber failure---the binstar token had expired.

andrrizzi:
Looks good! I'll retrigger the tests before merging.

andrrizzi:
Looks like the `GFORTRAN_1.4` error is back. I'm looking into it.

jchodera:
I just retriggered the build again because there ended up being two
different versions of alchemy-dev on binstar. Cleaned out the old one and
retriggered.

jchodera:
Looks like travis is saying that build 3 of `ambermini` is being pulled in, not build 4.  Build 3 brings in the wrong version of `libgfortran`.

I wonder if we should actually bump the `ambermini` version number so we can pin the version number instead of having anaconda sometimes grab the wrong build.  Or delete the old build number 3 version from anaconda cloud (which is probably easier).

andrrizzi:
Agree. I didn't even know that `conda` could arbitrarily decide which build to install. If users cannot control which build to install, then both solution sounds good.

jchodera:
I've deleted the problematic `ambermini` build number 3 from anaconda cloud and am retriggering travis

jchodera:
Woohoo!  Tests pass now.

andrrizzi:
Awesome! Merging.

PR #354, opened 2016-03-30
Fix broken Miniconda link

jchodera:


jchodera:
I'm confused as to why we're still getting
```
    Exception: antechamber failed to produce output mol2 file
    command: antechamber -i in.mol2 -fi mol2 -o out.mol2 -fo mol2 -s 2 -c bcc
    output:
    --------------------------------------------------------------------------------
    /home/travis/miniconda/envs/_test/bin/../bin/sqm: /home/travis/miniconda/envs/_test/bin/../lib/libgfortran.so.3: version `GFORTRAN_1.4' not found (required by /home/travis/miniconda/envs/_test/bin/../bin/sqm)
```
since we're now using `openmoltools-dev` and `ambermini`. Do we have to remove the `scipy` dependency from something here?

andrrizzi:
Does `openmoltools-dev` follows the recipe in `choderalab/openmoltools/devtools`? Because if that's the case, I don't think we updated that recipe to include `libgfortran 1.0` in the run requirements yet.

jchodera:
It does, but I just took `libgfortran 1.0` out of `ambermini`, which is where that dependency is actually brought in.  Should I add `libgfortran 1.0` back to `ambermini`?

andrrizzi:
Ok then I think we should be set if we just add `libgfortran 1.* # [linux and np == 19]` [here](https://github.com/choderalab/openmoltools/blob/master/devtools/conda-recipe/meta.yaml#L40)

jchodera:
D'oh, I had completely forgotten that [scipy 0.17 was the culprit](https://github.com/mila-udem/blocks/issues/836#issuecomment-142112468). I've made your change!

PR #358, opened 2016-04-07
Scripts for rendering replica movies of Abl:imatinib

jchodera:
This adds a script that @steven-albanese can use to render a trajectory of Abl:imatinib with MacPyMOL.

usage: Run `visualize.tcsh` to generate a movie from replica 0 of the `complex-explicit.nc` NetCDF file created by the YAML file in this example.

You should install `ffmpeg` first via `macports` or some other scheme.


andrrizzi:
Nice!

steven-albanese:
is this ready to be merged?

Issue #369, opened 2016-05-02
Time to cut 0.8.0 release


jchodera:
Issue #364 was fixed some time ago, but we haven't cut a new release since then. Since `alchemy` has now been updated to 1.2.0 and there have been a number of other improvements, it's time to cut an 0.8.0 release.

The release will use OpenMM 7.0 (technically on `conda` as `7.0rc1`), but since `openmm-dev` has some API-breaking changes added, I am going to first temporarily revert the `devtools/conda-recipe/meta.yaml` requirements to release requirements so we can test the 0.8.0 release in travis.

andrrizzi:
+1

Let me know if I can help.

jchodera:
I've just merged the `alchemy` 1.2.0 release conda recipe. Once that is built, I'll check in the updated `devtools/conda-recipe/meta.yaml` and we can fix any issues with the pinned version numbers in the recipe.

jchodera:
This is done!

PR #370, opened 2016-05-02
Update conda dev recipe to use pinned versions for 0.8.0 release

jchodera:
See #369

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -38,8 +38,8 @@ requirements:
     - netcdf4
     - openmm ==7.0rc1 # temporarily test against stable version
     - mdtraj 1.5.1 # TODO remove me when segment_id bug is fixed (this is installed through openmoltools)
-    - openmmtools-dev
-    - pymbar-dev
+    - openmmtools ==0.7.5 # temporarily test against stable version
+    - pymbar ==2.1beta0 # temporarily test against stable version
     - ambermini
     - docopt
     - openmoltools ==0.6.9 # temporarily test against stable version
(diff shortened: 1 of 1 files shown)


jchodera:
Travis is failing because somehow `openmm-dev` is still being brought in:
https://travis-ci.org/choderalab/yank/builds/127249181#L1592

Any ideas why @andrrizzi?  Does one of the other tools with pinned versions require `openmm-dev`?

andrrizzi:
I think `openmmtools-dev` in the test dependencies requires `openmm-dev`. Also, I'm not sure what happens during travis tests, but `openmoltools-dev` in the run dependencies requires `openmm-dev` as well.

jchodera:
Thanks for catching that!

andrrizzi:
No problem! I think there's still [this line](https://github.com/jchodera/yank/blob/master/devtools/conda-recipe/meta.yaml#L57) to correct.

jchodera:
Thanks!  Fixed.

andrrizzi:
Ah! We need to release `openmoltools 0.7.0` to release `yank` because we need the new `schrodinger` module. We already have a PR for that on `conda-recipes`. I'll see if tests pass there and we can proceed.

jchodera:
Thanks, @andrrizzi!

jchodera:
Looks like you cut `openmoltools` 0.7.0 already.  Do you want to update the [conda recipe](https://github.com/omnia-md/conda-recipes/tree/master/openmoltools) or should I?

jchodera:
Tests pass. I think we can cut a new release now.

Issue #339, opened 2015-12-17
Anisotropic dispersion correction


jchodera:
This is just a reminder to implement the long-range [anisotropic dispersion correction](http://dx.doi.org/10.1021/jp0735987).

PR #371, opened 2016-05-12
[WIP] Anisotropic dispersion correction

jchodera:
Began implementation of anisotropic dispersion correction by adding a `fully_interacting_state` for which all free energies are to be computed with respect to in each phase of modified Hamiltonian exchange.


jchodera:
Fixes #339

Yank/yank.py
```diff
@@ -316,12 +316,38 @@ def _create_phase(self, phase, reference_system, positions, atom_indices, thermo
         # Make a deep copy of the reference system so we don't accidentally modify it.
         reference_system = copy.deepcopy(reference_system)
 
+        # Store a serialized copy of the reference system.
+        from simtk.openmm import XmlSerializer
+        metadata['reference_sysem'] = XmlSerializer.serialize(reference_system)
```
andrrizzi: Is there a reason to store the systems both in the `metadata` and in the `thermodynamic_state` groups in the `.nc` file?

andrrizzi:
I can work on resolving conflicts and fix tests in a separate branch starting from here. Looks good to me in general beside that comment.

Yank/yank.py
andrrizzi: Also I'm not sure why we need to store the alchemically *un*modified reference system in addition to the `fully_interacting_system`.

jchodera: The fully-interacting system is the one we want to compute the reference free energy to, but it may have an enlarged cutoff, or might be perturbed in some other way. We probably also want the original `System` object (the unmodified reference system) in the NetCDF file as well.

jchodera:
> I can work on resolving conflicts and fix tests in a separate branch starting from here. Looks good to me in general beside that comment.

Thanks!

andrrizzi:
I've created a new branch in the main repo called `anisotropic-dispersion-correction` where I merged your commits and resolve the conflicts. Will keep working there.

Issue #382, opened 2016-06-01
Add test.sh test driver scripts to all examples


jchodera:
We need to add `test.sh` scripts to each example so that automated testing will catch any issues.

cc: https://github.com/choderalab/yank/issues/381

andrrizzi:
Sorry if I don't understand, but what's the advantage of adding `test.sh` files instead of just running the currently present `run.sh` with a lower number of iterations?

I think you removed the `test.sh` files with [this commit](https://github.com/choderalab/yank/commit/c3566477b2773bc705ffeb6697769b412d4628bb), although I wasn't in the lab so I don't know what was the rationale.

jchodera:
We're just going to use travis tests for now.

Issue #381, opened 2016-06-01
Add Jenkins badge for testing examples to `README.md`


jchodera:
To catch failing examples, we should add a Jenkins badge to indicate whether the current examples are passing or failing:

[![Build Status](https://jenkins.choderalab.org/buildStatus/icon?job=test-yank-linux-xeon-gtxtitan-jak)](https://jenkins.choderalab.org/job/test-yank-linux-xeon-gtxtitan-jak/)

Source:
```
[![Build Status](https://jenkins.choderalab.org/buildStatus/icon?job=test-yank-linux-xeon-gtxtitan-jak)](https://jenkins.choderalab.org/job/test-yank-linux-xeon-gtxtitan-jak/)
```
I just realized that there were a few things broken with the Jenkins example test configuration. I've hopefully fixed those now!

jchodera:
That's odd---the `test.sh` files have disappeared from the examples.  We'll need to add these back to get the automated example testing to run, @andrrizzi.  These should be like `run.sh`, but should ideally run a shorter simulation that simply ensures things are working properly.

jchodera:
Fixed by #386

PR #385, opened 2016-06-02
Convert p-xylene implicit example to YAML

andrrizzi:
T4 lysozyme + p-xylene in implicit solvent example converted to YAML.

I was thinking about deleting
* `p-xylene-implicit/run-torque-extend.sh`: this can be done with the `resume_simulation` option
* `p-xylene-implicit/setup/`: most of this files are recreated by the setup pipeline the things that would be lost are
  * `p-xylene-implicit/setup/generate-mol2-from-name.py`: automatic pipeline does that now
  * `p-xylene-implicit/setup/181L.pdb`: the simulation needs only the pdbfixed version
  * `p-xylene-implicit/setup/187L.pdb`: this is a different conformation of the pdb that we don't use in the example

In general, the less useless files are in these folders, the more readable is the example for the user + the fastest the installation through `setup.py` (which I use a lot during development). Objections/thoughts?


jchodera:
Can you tell me more about the `resume_simulation` option?  Do you need to add `resume_simulation: yes` to actually resume, or does this happen automatically?

> p-xylene-implicit/setup/: most of this files are recreated by the setup pipeline

Sounds fine to remove things that are recreated via the setup pipeline!

> In general, the less useless files are in these folders, the more readable is the example for the user

+1!

jchodera:
How many iterations should we have these examples run for?

On the one hand, it's useful to have the examples be short for people to not lose patience with them.

On the other hand, users have actually just run the examples for the provided length and expected to get reliable site-resolved binding free energies, so it may be useful to ensure we provide enough iterations to get reasonable results.

andrrizzi:
> Do you need to add `resume_simulation: yes` to actually resume, or does this happen automatically?

You need to add it explicitly. Trying to resume without setting it raises an exception telling you that it doesn't want to overwrite files.

> How many iterations should we have these examples run for?

Maybe we can just add a message in the README about this (or echo it on the shell) with instructions on how to change the number of iterations (modify the line in the YAML script) and let the user choose.

For the default value, does 250 for explicit and 500 for implicit sound good?

> Sounds fine to remove things that are recreated via the setup pipeline!

Is it ok if I remove also `generate-mol2-from-name.py`, `181L.pdb`, and `187L.pdb` even if they are not recreated? (see above for the reasons)

jchodera:
> You need to add it explicitly. Trying to resume without setting it raises an exception telling you that it doesn't want to overwrite files.

If we have `resume_simulation: yes` specified, it will automatically try to resume if the simulation files exist; otherwise it creates the simulation files if they don't exist?  That sounds like ideal behavior!

> Maybe we can just add a message in the README about this (or echo it on the shell) with instructions on how to change the number of iterations (modify the line in the YAML script) and let the user choose.

This sounds good, until we add some sort of "run until converged" automated mode.

> For the default value, does 250 for explicit and 500 for implicit sound good?

What do we want to accomplish here? If we want to set it to something short for testing, 10 iterations may be sufficient. If we want to set it to something that at least allows analysis go give something that is starting to be useful, 100-500 is good. If we want a real calculation, 1000 is probably the bare minimum.

> Is it ok if I remove also generate-mol2-from-name.py, 181L.pdb, and 187L.pdb even if they are not recreated? (see above for the reasons)

Yes!

andrrizzi:
> If we have resume_simulation: yes specified, it will automatically try to resume if the simulation files exist; otherwise it creates the simulation files if they don't exist?

Exactly.

> What do we want to accomplish here?

I was going for a compromise between not losing patience and reasonable final numbers. But if we want to be sure about the final numbers, then I agree that 1000 is the minimum.

For automatic testing of the yaml examples, I was thinking we could run a temporary copy of the yaml script that we programmatically modify to set `number_of_iterations: 2`.

Also, do we want to minimize in the examples? Minimizing all the 40 replicas usually takes a lot of time and I'm not sure it affects that much the final free energies in this case.

andrrizzi:
With the last commit I've implemented the code for automatically testing yaml examples.

Basically, it creates a temporary `test.sh` script from `run.sh` with only 2 iterations.

I'm not sure how this fits into the "testing the examples with Jenkins" (#382, #381) , but I think this could be a nice alternative to create `test.sh` files for each example.

jchodera:
> With the last commit I've implemented the code for automatically testing yaml examples.

Awesome! How does testing work? Does it happen automatically in `.travis.yml`?

Should we reconfigure Jenkins to do something else for testing, then? Jenkins has access to actual GPUs, so the thought was that we could do more heavy-duty testing that way:
https://jenkins.choderalab.org/job/test-yank-linux-xeon-gtxtitan-jak/configure

andrrizzi:
> How does testing work?

The examples testing is labeled 'slow' so travis doesn't touch it, but in general I suspect that testing doesn't work at all beside doctests (see #387 ). I'll do some experiments with the travis script to verify/solve it.

andrrizzi:
I think I understand what's wrong with tests. I'll merge this and fix it into a different PR.

jchodera:
Great, thanks!

PR #395, opened 2016-06-08
Small cleanup and remove mutable default arguments

### LINKED ISSUE #394: Eliminate mutable default arguments
cc: https://github.com/choderalab/perses/pull/211

See http://docs.python-guide.org/en/latest/writing/gotchas/

andrrizzi:
Remove unused functions and mutable default arguments (fix #394 ).


andrrizzi:
Ready to be merged! The only 2 tests failing are the `repex` ones that we still need to correct.

jchodera:
Woohoo!  Thanks!  

In general, it's best if we can not shorten `openmmtools` to `omt`, but there's certainly no need to change this.

Looks good to me otherwise. Merge when ready!

andrrizzi:
> it's best if we can not shorten openmmtools to omt

ok! Merging.

PR #396, opened 2016-06-09
Set recommended GB radii and expand options for solvent

### LINKED ISSUE #216: Allow to set cutoff with implicit solvent (and explicit solvent) simulations
When working with a big system, it could be useful to include the possibility to set the cutoff for non-bonded interactions, when you use the nonbonded method --nbmethod=CutoffPeriodic or --nbmethod=CutoffNonPeriodic, so for both explicit and implicit solvent. And if the cutoff is not given, it could use getcutoffdistance(), as is now implemented for the explicit solvent case.

This could possibly be useful for a next version (in my opinion). 

I have now included it in my "own" yank version. However, without a dispersion correction. Would this need also the dispersion correction in the implicit solvent case?

andrrizzi:
In this PR:
* Set GB radii to recommended settings by Amber for the implicit solvent model
* Fix #216 (allow set nonbonded cutoff with implicit solvent)
* Expand YAML solvent options to everything that can be passed to `createSystem()` by using introspection.

### DIFF
--- a/Yank/pipeline.py
+++ b/Yank/pipeline.py
@@ -123,6 +123,43 @@ def find_components(topology, ligand_dsl, ligand_net_charge=0,
     return atom_indices
 
 
+# See Amber manual Table 4.1 http://ambermd.org/doc12/Amber15.pdf
+_OPENMM_TO_TLEAP_PBRADII = {'HCT': 'mbondi', 'OBC1': 'mbondi2', 'OBC2': 'mbondi2',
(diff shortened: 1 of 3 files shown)


andrrizzi:
This is ready to be merged!

The third thing is not complete because the YAML syntax checking still allow to specify a limited number of options, but I'll fix that when I'll switch to schema in a separate PR.

The two tests failing are the known repex ones.

jchodera:
Awesome!  Code looks good to me.

PR #403, opened 2016-06-13
Keep original ligand conformation during OE charging

### LINKED ISSUE #402: Keep same input configuration when generating charges with OpenEye scheme
Just need to change the value of a function parameter during the setup.

andrrizzi:
Fix #402 

If the ligand conformation is already specified, we should force it to avoid moving ligands that should start in the binding pocket.

Waiting for tests to pass.


andrrizzi:
Ready to be merged. Only repex tests failing.

jchodera:
Great, thanks!

PR #404, opened 2016-06-14
Fix handling of periodic boundary conditions for non-periodic systems

### LINKED ISSUE #398: Debug replica exchange harmonic oscillator test failures

jchodera:
Fixes #398

### DIFF
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -1055,8 +1055,7 @@ class ReplicaExchange(object):
         if self.show_energies:
             self._show_energies()
 
-        # Analysis objet starts off empty.
-        # TODO: Use an empty dict instead?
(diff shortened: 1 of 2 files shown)


jchodera:
The positions in `ReplicaExchagne` were being retrieved by `getState(getPositions=True,enforcePeriodicBoundaries=True)`, which would wrap particles in non-periodic systems by accident. This now selects whether or not periodic boundaries should be enforced based on whether the system is periodic or not.

jchodera:
This should be ready to review and merge, @andrrizzi.

(Replica exchange tests pass locally.)

jchodera:
Woohoo!  Travis tests pass.

andrrizzi:
Awesome! Looks good! Merging.

PR #408, opened 2016-06-22
Validate YAML syntax with schema

### LINKED ISSUE #390: Switch to a proper library to validate YAML syntax
Currently the validation is sone manually in the code. There are language-agnostic data validation libraries for python that can be used to validate YAML, JSON, etc.

[schema](https://github.com/keleshev/schema) is a good option that is already packaged in the omnia conda channel.

andrrizzi:
Fix #390

This PR is going to be a little messy because of heavy refactoring, sorry. It adds validation of YAML syntax through `schema` library which is a new dependency that can be already found in omnia conda channel.

Still have to commit a few tests.


andrrizzi:
Any idea on the [new repex error](https://travis-ci.org/choderalab/yank/builds/139535566#L1809)? Looks like an OpenMM issue, probably due to the fact that I had to switch to `openmoltools-dev` from `openmoltools-0.7.0`, but I can't reproduce it locally neither on linux nor on OS X.

jchodera:
The error is for `barostat.setTemperature(state.temperature)`, whose OpenMM interface was last updated in https://github.com/pandegroup/openmm/pull/1468

It's now called `barostat.setDefaultTemperature()`, and only affects the default temperature (and not the current temperature) so we will have to make sure our code is doing what we want.

This was changed Apr 21, so the latest OpenMM release doesn't have the new interface, but the `openmm-dev` version does.  Suggest we just use `try...except` or use `if hasattr(barostat, 'setDefaultTemperature')` to get this to work with both versions of the code.

andrrizzi:
Thanks so much! That fixed the problem. Tests pass. Feel free to merge.

jchodera:
Huzzah!

PR #413, opened 2016-06-27
Setup all systems before running Yank

### LINKED ISSUE #392: Have combinatorial experiment setup actually set up all systems before running any
This might be more convenient for most users.

andrrizzi:
Fix #392. Waiting for tests to pass.

### DIFF
--- a/Yank/tests/test_yaml.py
+++ b/Yank/tests/test_yaml.py
@@ -138,7 +138,6 @@ def test_yaml_parsing():
 
     options:
         verbose: true
-        mpi: yes
         resume_setup: true
         resume_simulation: true
         output_dir: /path/to/output/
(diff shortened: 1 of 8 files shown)


andrrizzi:
Ready to be merged!

andrrizzi:
I'm about to commit also automatic detection of MPI execution.

jchodera:
Looks good!  

Do we still want to allow the `mpi:` YAML option in case someone wants to force behavior on or off, or is this just more potential for trouble than its worth?

andrrizzi:
I removed it, but I can add it back. Technically the way to force it is to run yank with `mpirun`. Those environment variables get set for each process only when you run your script with MPI, otherwise those variables are not set.

jchodera:
OK, sounds good to me!

andrrizzi:
Great! I've tested it on the cluster and it seems to work well. Merging.

PR #420, opened 2016-07-07
[WIP] Reintroduce solvation free energies in Yank and remove hardcoded phase names

andrrizzi:
This modify phase creation and analysis to be independent from hardcoded phase names, and modify the analysis to support solvation free energies again.


andrrizzi:
Tests pass! This might be a good point to review @jchodera . The rest of the PR will be only a slight modification of the YAML syntax to support this changes.

Question: which conventions do we want to put on the thermodynamic cycle direction? Right now I'm assigning the sign + for the complex phase and - for the solvent. This depends also on the protocol used though (whether we start with fully interacting and move to decoupled, or the opposite).

jchodera:
Reviewing now!

jchodera:
Looks great!

I wonder if there's a more meaningful syntax we can use for the thermodynamic leg direction. If we define the free energy of a thermodynamic leg (alchemical phase) as always going from `lambda = 0 -> 1`, then we might want the direction to represent the sign we use in reconstructing a property of interest. But maybe the direction shouldn't be associated with the `AlchemicalPhase`, but with the object we form out of several `AlchemicalPhase` objects to compute a specific property.

For example, if we ever want to compute partition coefficients among N solvents, we can actually form N(N-1) different partition coefficients using pairs of these free energies combined with different signs. So maybe we want to separate the "direction" concept from the `AlchemicalPhase` and instead associate it with one or more properties that can be computed from the collection of `AlchemicalPhase` results afterwards?

andrrizzi:
Good point! Let me think about it, I'll try to come up with a possible solution.

jchodera:
I can imagine a case where we have several experiment generators that each
produce several alchemical phases and analyses objects that specify how to
compute quantities of interest from those phases.

The analyses objects can be as a coefficient to use when combining the free
energy from each phase, assuming we are always considering the free energy
of lambda from 0 to 1 is the free energy for that phase.

This might be another case where we can introduce an identifier for the
experiment generator into the YAML once we grow to the point where it is no
longer unambiguous, but we would then have to build in default behavior if
this is omitted.

andrrizzi:
If we don't want to specify the sign within the phase, I think we'll have to pass the signs of each phases to `analyze` each time. What about modifying the `analyze.analyze()` signature to something like
```python
def analyze(analysis)
```
where `analysis[ncfile_path]` is the coefficient to use for the analysis (e.g. `analysis['output/complex.nc'] == 1`. We'll also have to modify `yank analyze` CLI command to something like
```
yank analyze --phase "output/complex.nc" --sign "1"  --phase "output/solvent.nc" --sign "-1"
```

jchodera:
I think the problem is that we decided to have separate NetCDF files for
each alchemical phases. This was originally because we wanted to run
different phases in parallel, but this isn't something we support at the
current time. We've had luck with Perses having each phase write to a
different group of the NetCDF file, leaving the root group free for meta
information about the whole calculation, like how to combine things.

I'd eventually like to move to something like that, but for now, what about
creating a Calculation object containing several AlchemicalPhase objects
and several AnalysisScheme objects that contain information on how to
combine the AlchemicalPhase objects, and just pickling this to disk. (In
fact, we only need the filenames for the AlchemicalPhase objects for now.)
We can then point the analysis tool at this pickle file.

In future versions, we can encapsulate all of this in a single NetCDF file
after we migrate to a new repex that can use groups of already existing
NetCDF files to store each alchemical phase.

andrrizzi:
> and just pickling this to disk

Sounds great! For now, I would just dump a simple text `analysis` file in the same folder containing paths and signs. We don't need more complicated things right now, and I'm worried that the user will end up with tons of unusable data, if the pickle file gets deleted by mistake (regenerate the pickle file by recreating exact AlchemicalPhase and Analysis objects sounds a bit like an overkill).

jchodera:
Please go with whatever format is easiest! I thought pickling a list or object might be easier than reading/writing text files (we don't want to write a parser!), but the main goal is to do whatever is easiest right now but be able to do something more clever in the future (like store this all in a single NetCDF).

andrrizzi:
I've opted to dump the analysis object again through YAML so that it's easy to recreate and we don't have to write a parser for it. For now it's saved as
```yaml
- [phase1_name, 1]  # .nc file will be "phase1_name.nc", 1 is the coefficient for DeltaF
- [phase2_name, -1]
```
**Question**: I'm about to create a hydration free energy example. Is there a particularly meaningful system I should set up that we can use to validate YANK?

jchodera:
> For now it's saved as
```
- [phase1_name, 1]  # .nc file will be "phase1_name.nc", 1 is the coefficient for DeltaF
- [phase2_name, -1]
```
YAML is fine, but I think the information might be just a bit too rudimentary. For each computed property, we'll want to store the following 
* Computed property name (e.g. `binding free energy`, `hydration free energy`, `transfer free energy from octanol to water`)
* Each alchemical phase NetCDF filename that should be included (you already have this too)
* the sign for each alchemical phase (you already have this too)
* Any additive corrections (e.g. the standard state correction)
* We may also want to store information about the thermodynamic state at which property is computed (e.g. 298 K and 1 atm), especially if we might later want to compute properties at different thermodynamic states too

For test hydration free energies, the following are useful:
* TIP3P water in TIP3P water at 300 K and 1 atm (I think this is available in the literature)
* benzene in TIP3P water at 300 K and 1 atm
* Computed hydration free energies for much of the [FreeSolv set](http://www.ncbi.nlm.nih.gov/pubmed/24928188) is available now. You can find the latest version on [GitHub here](https://github.com/choderalab/FreeSolv). The primary experimental data (and is all available [in a pickle file](https://github.com/choderalab/FreeSolv/tree/master/primary-data), and there are several formats available (IUPAC names, SMILES, Tripos mol2, SDF, etc.). Mobley group calculated values with TIP3P/GAFF/AM1-BCC are available in the [annotated database](https://github.com/choderalab/FreeSolv/blob/master/database.pickle), which you can also peruse in [text form](https://github.com/choderalab/FreeSolv/blob/master/database.txt).

andrrizzi:
Thanks! This is super helpful. I'll go with the first two for starters.

> I think the information might be just a bit too rudimentary.

Sure! I don't think we have to support user editing of this file for yank 1.0, so we'll be able to modify it as necessary when we'll have more sophisticated needs. We don't even have to stick with YAML. This is just to make things work.

jchodera:
> Sure! I don't think we have to support user editing of this file for yank 1.0, so we'll be able to modify it as necessary when we'll have more sophisticated needs. We don't even have to stick with YAML. This is just to make things work.

Agreed. We 100% don't want users editing this, but we *do* want to make sure we store all the information that is needed to compute the free energies for the things we want. This will be a much better place to store the standard state correction in the long run, for example.

andrrizzi:
I'll now test the examples on the cluster, and if everything passes, I'd like to merge this if you agree @jchodera . I have updated here the examples to the new YAML syntax, and I'd like to discuss these tomorrow morning.

jchodera:
Thanks! I've been called to Jury Duty tomorrow, so won't be in. Go ahead and merge if it seems like things are working. We can add the hydration free energy calculations to the Jenkins tests.

andrrizzi:
Ok! All tests pass and things seem to run smoothly in MPI mode too. Merging.

jchodera:
Awesome!

PR #421, opened 2016-07-13
Support for inclusion of multiple parameter files

### LINKED ISSUE #360: How can we specify additional forcefield loading statements for LEaP?
For dealing with divalent cations, we need to do more than load a `leaprc.*` file.

There is an additional step of loading another `frcmod` file:
```
source leaprc.ff14SB 
loadAmberParams frcmod.ionslrcm_hfe_tip3p 
```
Do we currently have the capability to specify these extra lines to be passed to leap?

### LINKED ISSUE #410: Changes to YAML syntax for multiple parameters and start from prmtop files
I've thought about how to change the syntax to accomodate both #360 and #384, and I've come to 2 conclusions.

* **We can't use normal lists to represent combinations**, or we won't be able to specify simple lists. I think we should change the syntax to become
```yaml
molecules:
  mol_id:
    filepath: 'ligands.sdf'
    parameters: [leaprc.ff99SB, S2P.frcmod, S2P.off]
    select: !Combinatorial [0, 1, 5]
```
* **We should separate system setup in its own block**.  The final part of the script would become
```yaml
...

systems:
  system_from_files:
    complex_phase: 'my_complex.prmtop'
    solvent_phase: 'my_solvent.prmtop'
    ligand_dsl: 'resname MOL'
    solvent: PME-solvent
  system_automatic_setup:
    receptor: abl
    ligand: !Combinatorial [imatinib, bosutinib]
    solvent: PME-solvent
    pipeline: 'leap' # I'd add this once we have an OpenMM pipeline as well
    extra_tleap_commands:
    - loadAmberParams frcmod_S2P
    - anyOtherCommand

experiments:
  system: !Combinatorial [system_from_files, system_automatic_setup]
  protocol: absolute-binding
```
This also has the advantage that **when we'll split the setup pipline into its own tool, the YAML syntax will remain unchanged**.

How does it look? Better solutions?

andrrizzi:
* Fix #410 : expose details of molecule pipeline, and allow specification of parameters at the system level too.
* Fix #360 : support for inclusion of multiple parameter files.


andrrizzi:
Ready to be merged!

jchodera:
Awesome, thanks!

PR #424, opened 2016-07-15
Anisotropic dispersion correction

### LINKED ISSUE #339: Anisotropic dispersion correction

andrrizzi:
This continue the work in #371 to implement anisotropic long range dispersion correction.

Fix #339 .

### DIFF
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -1048,8 +1048,55 @@ class ReplicaExchange(object):
         for replica_index in range(self.nstates):
             self.replica_states[replica_index] = replica_index
 
+        # Check to make sure NetCDF file exists.
+        if not os.path.exists(self.store_filename):
+            raise Exception("Store file %s does not exist." % self.store_filename)
+
+        # Open NetCDF file for reading
+        logger.debug("Reading NetCDF file '%s'..." % self.store_filename)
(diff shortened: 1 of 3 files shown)


andrrizzi:
I've been digging into pymbar interface and code to understand what's the best way to do this.

I think I just have to use the function `mbar.computePerturbedFreeEnergies(u_kn)` where `u_kn` is the `1 x nreplicas x niterations` array containing the energies of the fully interacting state. Is that correct?

andrrizzi:
Ah, no I misunderstood the interface. For that to work the number of states must be equal to the original one. Will keep digging.

jchodera:
Let me tackle that part in a separate PR! It's nontrivial.

andrrizzi:
Ok! Feel free to merge this if think the changes are good then!

andrrizzi:
If it's just implementing in python formula (8) of the manuscript I think I can handle it if you don't have much time. Otherwise I'll work on setting up validation.

jchodera:
No, it's reworking the input `u_kln` matrix to add another state from which no samples were drawn. Should be easy. I can do it now!

Will merge this right away and you can move onto other things!

andrrizzi:
Thanks!

PR #425, opened 2016-07-16
[WIP] Parallelize travis

jchodera:
This is an attempt to parallelize travis, following the [perses travis](https://github.com/choderalab/perses/blob/master/.travis.yml).


jchodera:
Hm, this doesn't seem to save much time. Current serial tests take 16 min, but parallel tests take 26 min because, while some tests run concurrently, not all do.

Some issues I can see:
* All parallel branches still need to build yank and install all the examples, which seems to take ~7 min.
* Testing `yank.tests.test_cli` tests `yank.tests.test_cli.test_selftest`, which runs *all* nosetests again.

andrrizzi:
Also, some tests currently are all `slow` or there are no tests (e.g. `yank.test.test_examples` and `yank.mixing`), so in those cases it basically only install yank and quit.

jchodera:
Closing for now, since this was a lot slower.

PR #427, opened 2016-07-17
Add FreeSolv-mini example

andrrizzi:
Also fixes existing YAML examples syntax.

### DIFF
--- a/examples/hydration-smiles/freesolv.yaml
+++ b/examples/hydration-smiles/freesolv.yaml
@@ -1,0 +1,49 @@
+---
+options:
+  minimize: yes
+  verbose: yes
+  output_dir: .
+  number_of_iterations: 5000
+  temperature: 300*kelvin
+  pressure: 1*atmosphere
+  softcore_beta: 0.0
+
+molecules:
+  molecule:
+    filepath: freesolv-mini.smiles
+    openeye:
+      quacpac: am1-bcc
+    antechamber:
+      charge_method: null
+    select: all
+
+solvents:
+  RF:
+    nonbonded_method: PME
+    nonbonded_cutoff: 9*angstroms
+    clearance: 9*angstroms
+  vacuum:
+    nonbonded_method: NoCutoff
+
+systems:
+  hydration-system:
+    solute: molecule
+    solvent1: RF
+    solvent2: vacuum
+    leap:
+      parameters: ['leaprc.ff14SB', 'leaprc.gaff']
+
+protocols:
+  hydration-protocol:
+    solvent1:
+      alchemical_path:
+        lambda_electrostatics: [1.00, 0.75, 0.50, 0.25, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00]
+        lambda_sterics:        [1.00, 1.00, 1.00, 1.00, 1.00, 0.95, 0.90, 0.85, 0.80, 0.75, 0.70, 0.65, 0.60, 0.50, 0.40, 0.30, 0.20, 0.10, 0.00]
+    solvent2:
+      alchemical_path:
+        lambda_electrostatics: [1.00, 0.75, 0.50, 0.25, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00]
+        lambda_sterics:        [1.00, 1.00, 1.00, 1.00, 1.00, 0.95, 0.90, 0.85, 0.80, 0.75, 0.70, 0.65, 0.60, 0.50, 0.40, 0.30, 0.20, 0.10, 0.00]
+
+experiments:
+  system: hydration-system
+  protocol: hydration-protocol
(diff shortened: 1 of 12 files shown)


andrrizzi:
This is ready to be merged.

@jchodera Can you take a look at the [protocol used for validation](https://github.com/andrrizzi/yank/blob/validation/examples/hydration-smiles/freesolv.yaml) before I try to run the calculations? In particular, do we want to go through `epik` and add ions if it gives us a charged molecule?

examples/hydration-smiles/freesolv.yaml
```diff
@@ -0,0 +1,49 @@
...
+    nonbonded_method: PME
+    nonbonded_cutoff: 9*angstroms
+    clearance: 9*angstroms
+  vacuum:
+    nonbonded_method: NoCutoff
+
+systems:
+  hydration-system:
+    solute: molecule
+    solvent1: RF
```
jchodera: Do you want to call this solvent `water`?
Also, we probably want to change this to PME if we want to match `FreeSolv`.

examples/hydration-smiles/freesolv.yaml
```diff
@@ -0,0 +1,49 @@
...
+    solvent2: vacuum
```
jchodera: Change this solvent name to `vacuum`?

examples/hydration-smiles/freesolv.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+protocols:
+  hydration-protocol:
+    solvent1:
+      alchemical_path:
+        lambda_electrostatics: [1.00, 0.75, 0.50, 0.25, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00]
+        lambda_sterics:        [1.00, 1.00, 1.00, 1.00, 1.00, 0.95, 0.90, 0.85, 0.80, 0.75, 0.70, 0.65, 0.60, 0.50, 0.40, 0.30, 0.20, 0.10, 0.00]
+    solvent2:
+      alchemical_path:
+        lambda_electrostatics: [1.00, 0.75, 0.50, 0.25, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00]
```
jchodera: I think many fewer lambda values were necessary here. Did the [FreeSolv paper](http://www.ncbi.nlm.nih.gov/pubmed/24928188) describe the protocol used for vacuum?

examples/hydration-smiles/freesolv.yaml
andrrizzi: Oh! The solvent is PME (see 'nondbonded_method'), I just forgot to rename it after copy-pasting from another example. I'll call it `water`.

examples/hydration-smiles/freesolv.yaml
jchodera: I see the FreeSolv paper didn't, so it's probably best to refer to [this earlier paper](http://pubs.acs.org/doi/abs/10.1021/ct800409d) for protocol.

examples/hydration-smiles/freesolv.yaml
andrrizzi: Not sure I understand, the solvent name is already `vacuum`, right?

jchodera:
> In particular, do we want to go through epik and add ions if it gives us a charged molecule?

All the molecules in FreeSolv should be simulated in their neutral form, I believe.

examples/hydration-smiles/freesolv.yaml
jchodera: Oh!  I see.  I was confused why we used `solvent1` again below in the protocols, but this makes sense.

examples/hydration-smiles/freesolv.yaml
jchodera: Ah!  Thanks!

examples/hydration-smiles/freesolv.yaml
```diff
@@ -0,0 +1,49 @@
...
+  pressure: 1*atmosphere
+  softcore_beta: 0.0
+
+molecules:
+  molecule:
+    filepath: freesolv-mini.smiles
+    openeye:
+      quacpac: am1-bcc
+    antechamber:
+      charge_method: null
```
jchodera: Does `null` just read in the charges from the `OpenEye`-generated mol2 files?

jchodera:
OK, done!  Go ahead and merge after you feel the inline comments are addressed!

examples/hydration-smiles/freesolv.yaml
andrrizzi: The FreeSolv paper doesn't include a protocol per se as it is an aggregation of previous simulations.

That's the exact same protocol used for the 504 database, which is described [here](http://pubs.acs.org.proxy.library.cornell.edu/doi/pdf/10.1021/jp0667442).  It says that the "vacuum calculations used the same parameters" but I can change it if you think that's too many states.

examples/hydration-smiles/freesolv.yaml
andrrizzi: Exactly.

examples/hydration-smiles/freesolv.yaml
jchodera: Huh!  Well, it should be fine. Let's stick with it for now.

andrrizzi:
Updated, thanks! Merging.

Issue #379, opened 2016-05-25
Add feature to allow extraction of replica or state trajectories for visualization


jchodera:
@luirink has written a script to extract trajectories from NetCDF store files into XTC format (either by replica or state) to make them easier to visualize in PyMOL or VMD.  I'd like to add this as a CLI feature via `yank write-trajectory` or some related name.

Any suggestions?

andrrizzi:
An alternative would be to make it an argument of an existing, more flexible command. Maybe something like `yank analyze --writetrajectory`.

andrrizzi:
#309 should be implemented first so that we won't have to rely on a `pdb` file to extract the topology.

Issue #375, opened 2016-05-22
Check for existence of specified YAML file, raising an informative error message if not found


jchodera:
I'm getting the following error with the 0.8.0 conda package:
```
Running simulation via MPI...
Traceback (most recent call last):
  File "/cbio/jclab/home/chodera/miniconda/bin/yank", line 9, in <module>
Traceback (most recent call last):
  File "/cbio/jclab/home/chodera/miniconda/bin/yank", line 9, in <module>
Traceback (most recent call last):
  File "/cbio/jclab/home/chodera/miniconda/bin/yank", line 9, in <module>
Traceback (most recent call last):
  File "/cbio/jclab/home/chodera/miniconda/bin/yank", line 9, in <module>
    load_entry_point('yank==0.9.0', 'console_scripts', 'yank')()
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/yank-0.9.0-py2.7-linux-x86_64.egg/yank/cli.py", line 105, in main
    load_entry_point('yank==0.9.0', 'console_scripts', 'yank')()
    load_entry_point('yank==0.9.0', 'console_scripts', 'yank')()
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/yank-0.9.0-py2.7-linux-x86_64.egg/yank/cli.py", line 105, in main
    load_entry_point('yank==0.9.0', 'console_scripts', 'yank')()
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/yank-0.9.0-py2.7-linux-x86_64.egg/yank/cli.py", line 105, in main
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/yank-0.9.0-py2.7-linux-x86_64.egg/yank/cli.py", line 105, in main
    dispatched = getattr(commands, command).dispatch(args)
    dispatched = getattr(commands, command).dispatch(args)
    dispatched = getattr(commands, command).dispatch(args)
    dispatched = getattr(commands, command).dispatch(args)
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/yank-0.9.0-py2.7-linux-x86_64.egg/yank/commands/script.py", line 33, in dispatch
  File "/cbio/jclab/home/chodera/miniconda/lib/python2.7/site-packages/yank-0.9.0-py2.7-linux-x86_64.egg/yank/commands/script.py", line 33, in dispatch
[...]

PR #428, opened 2016-07-17
Store topology, write trajectory

### LINKED ISSUE #309: Store serialized OpenMM Topology objects in NetCDF store files
It would be helpful for generating visualizations and trajectories to store the OpenMM `Topology` objects corresponding to each phase as serialized objects in the NetCDF store files.

andrrizzi:
Fix #309, #379 and #375.

I couldn't find a way to serialize the OpenMM `Topology` object without pickling, so I saved instead the MDTraj topology. I basically saved the arguments of `mdtraj.Topology.from_dataframe()` needed to recreate the topology into a JSON string. At least the representation will be constrained by the API and it won't break every time someone changes the implementation of `Topology`.


jchodera:
I think mdtraj will serialize the Topology into JSON for you automatically.
It does this for HDF5 storage. See

http://mdtraj.org/latest/hdf5_format.html#topology

andrrizzi:
Yeah, that was my first attempt. Unfortunately, I don't think it does it automatically. That format is implemented manually in the `HDF5TrajectoryFile` class [here](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/hdf5.py#L304-L380) and it stores the representation directly into an HDF5 file. I couldn't find any method exposed by `Topology` object to serialize it automatically.

jchodera:
Ah, OK!

Maybe we should post in the OpenMM issue tracker to see if we can get serialization capability eventually migrated into OpenMM proper?

Thanks!

andrrizzi:
Will do!

andrrizzi:
I've added the command `yank analyze extract-trajectory` based on a function I've implemented some time ago. I tested it on a local netcdf file and it works well.

Feel free to merge if you approve the choice of the command name and tests passed.

Tomorrow I'll cut out a new release.

andrrizzi:
I'll add the commits for the new release here as I want to test these changes too with the fixed-version recipe.

andrrizzi:
Ok! I'll merge this and test a couple of examples with the next `yank-dev` on the cluster before releasing on `conda-recipes`.

PR #429, opened 2016-07-18
Replica exchange bug fixing

andrrizzi:
Fixing few bugs that came up while testing the examples. I'll merge this if tests pass and proceed with the release.


jchodera:
Thanks!

andrrizzi:
Ok everything seems to be working. Merging.

PR #430, opened 2016-07-19
Set back to development config after release

andrrizzi:
Will also test if everything work with updated versions of `ambermini` and `mdtraj`.


andrrizzi:
Looks like everything is working. I've updated to `ambermini15` because `ambermini14` gave me `GLIBC_2.7 Not found` error in omnia-md/conda-recipes#558 . The latest `mdtraj` should have fixed the `segment_id` bug.

PR #439, opened 2016-07-26
[WIP] Revise "Running YANK" documentation

jchodera:
@andrrizzi: I've taken a pass through the `Running YANK` section, but I realize that I am not entirely sure how we currently support some of the tasks described there with the YAML input.

Questions:
* How do we currently specify that we want to run with MPI?
* If MPI is requested but not found, do we revert to serial mode or raise an exception?
* How do we now specify which platform and precision mode we want to use? If we don't have a scheme, should this be done by the YAML file or a command-line argument?
* Should we deprecate many of the existing YANK CLI options, such as `yank prepare`, `yank run`, and many of the additional arguments used to specify things like timestep?
* Should we keep the name `yank script`, or change it to something else (since this really isn't a script, but the recommended way of creating YANK jobs now)?

I can tackle documentation and code changes this week if you can just give me a bit of guidance on these questions!


andrrizzi:
Awesome!
> * How do we currently specify that we want to run with MPI?
* If MPI is requested but not found, do we revert to serial mode or raise an exception?

MPI is automatically detected when we call yank with `mpirun`. Currently there is no option to force behavior but [the initialization function](https://github.com/choderalab/yank/blob/master/Yank/utils.py#L151) should work well for many different MPI implementations (if not all).

> * How do we now specify which platform and precision mode we want to use?

Platform and precision were two options that don't work in YANK (as far as I remember) because they are not connected to anything neither in the class `Yank` nor in `repex`. We need to add this to the code. I thought I opened an issue about this but maybe I forgot.

> * If we don't have a scheme, should this be done by the YAML file or a command-line argument?

It would be very handy to have them on the CLI, I think. The same goes for forcing the `mpi` option. I guess the ideal would be to have them on both (YAML and CLI) with CLI priority?

> * Should we deprecate many of the existing YANK CLI options, such as `yank prepare`, `yank run`, and many of the additional arguments used to specify things like timestep?
* Should we keep the name `yank script`, or change it to something else (since this really isn't a script, but the recommended way of creating YANK jobs now)?

I've been thinking about this a little. I think we should deprecate `yank prepare amber/gromacs` and many arguments that should now specified in the YAML script (I was waiting to convert all the examples to YAML before doing it), but maybe we can recycle `yank prepare` and `yank run` somehow.

We could remove `yank script` as you say, and have a single `yank prepare --yaml=yank.yaml` that creates the phases, and `yank run --yaml=yank.yaml`? Although one may run into inconsistencies if he modifies the script in between the two commands. Or just have a single `yank run --yaml=yank.yaml` that does everything.

andrrizzi:
Just opened an issue on platform/precision #441 .

PR #442, opened 2016-07-26
Fix bug that prevents Yank to run on multiple GPUs with MPI

andrrizzi:
I'm still working on a second problem that showed up and I want to verify, but after I have corrected it, we should probably make a new release.

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -532,7 +532,7 @@ def analyze(source_directory):
 
     # Print energies
     logger.info("")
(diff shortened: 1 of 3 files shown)


andrrizzi:
I've opened a new issue #443 describing the problem I found. There's also another issue with resuming introduced with the fully interacting state that I need to fix, but I think I already know where is the problem there.

andrrizzi:
Merging this. I'll solve the problem in #443 in another PR.

PR #445, opened 2016-07-30
[WIP] Extend analysis to include fully-interacting system

jchodera:
This includes a bugfix for crashes that would occur while writing to the NetCDF file in MPI mode.

### DIFF
--- a/Yank/sampling.py
+++ b/Yank/sampling.py
@@ -789,7 +789,10 @@ class ModifiedHamiltonianExchange(ReplicaExchange):
 
     def _write_iteration_netcdf(self):
         super(ModifiedHamiltonianExchange, self)._write_iteration_netcdf()
-        if self.fully_interacting_state is not None:
+        if self.mpicomm:
+            # Only the root node will write data.
+            if self.mpicomm.rank != 0: return
+        if (self.fully_interacting_state is not None):
             self.ncfile.variables['fully_interacting_energies'][self.iteration, :] = self.u_k[:]
             self.ncfile.sync()
 
--- a/examples/hydration-smiles/freesolv.yaml
+++ b/examples/hydration-smiles/freesolv.yaml
@@ -19,7 +19,7 @@ molecules:
 
 solvents:
   water:
-    nonbonded_method: PME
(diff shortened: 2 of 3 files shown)


examples/hydration-smiles/freesolv.yaml
```diff
@@ -19,7 +19,7 @@ molecules:
 
 solvents:
   water:
-    nonbonded_method: PME
+    nonbonded_method: CutoffPeriodic
```
jchodera: I'm just debugging this at the moment. Will change back to PME later.

Yank/sampling.py
```diff
@@ -789,7 +789,10 @@ def _initialize_netcdf(self):
 
     def _write_iteration_netcdf(self):
         super(ModifiedHamiltonianExchange, self)._write_iteration_netcdf()
-        if self.fully_interacting_state is not None:
+        if self.mpicomm:
+            # Only the root node will write data.
+            if self.mpicomm.rank != 0: return
```
andrrizzi: I've already added this in PR #442 with another couple of bugfixes. I wanted to implement also the detection of `exclusive` vs `shared` mode before merging, but I can do it right now and you can debug from that one so that we'll avoid having to solve conflicts.

jchodera: That would be great! Thanks!

jchodera: Alternatively, autodetection could certainly go into a separate PR.

jchodera:
Closing this to open a new PR.

Yank/sampling.py
andrrizzi: > Alternatively, autodetection could certainly go into a separate PR.

Oh! Sorry, that's what I meant. Now that I read what I wrote above, it's a mess. Anyway now I've merged my PR so you can fetch from upstream and start from there.

Maybe it would be useful to use branches instead of forks internally, so that anyone can modify the PR of somebody else.

jchodera: Agreed! We should probably protect master and switch to using feature branches instead

andrrizzi: > We should probably protect master

Done!

PR #451, opened 2016-08-17
Fix barostat issues for hydration free energy calculations

jchodera:
- [x] Add test that fails if no barostat is defined
- [x] Fix barostat issues for hydration free energy calculations

@andrrizzi : I updated the `testyaml.py:test_run_solvation_experiment()` test so that it creates the proper `solvent1` and `solvent2` phases instead of `complex` and `solvent` phases.


jchodera:
OK, I *think* I've fixed this.

Yank/tests/test_yaml.py
```diff
@@ -1439,7 +1438,6 @@ def test_setup_explicit_solvation_system():
             assert os.path.getsize(inpcrd_path) > 0
             assert found_resnames == expected_resnames[phase]
 
-
```
andrrizzi: This is very minor, but PEP8 suggests to separate top level functions with 2 blank lines: https://www.python.org/dev/peps/pep-0008/#blank-lines

andrrizzi:
Awesome! Thank you for catching this and for adding the example.

PR #455, opened 2016-08-20
More NaN protection / debugging

jchodera:
* Check for NaNs after Monte Carlo proposals, rejecting if energy is NaN
* Check for NaNs at beginning of dynamics propagation


jchodera:
All tests passed, so I'm not sure the build exited with 1.

PR #464, opened 2016-08-28
Adds filename check as tleap does not allow numeric only variables

jlerche:
This is a tleap error that we noticed because we occasionally automatically generated mol2 filenames that were numeric only, ie '4354.mol2'. The filename is then written in the *.leap.in file which when fed into tleap generates a syntax error.

Welcome to LEaP!
(no leaprc in search path)
52432 = loadMol2 ./setup/molecules/52432/52432.gaff.mol2  
ERROR: syntax error

This pull request adds a prototype for a simple check that then raises an exception

### DIFF
--- a/Yank/yamlbuild.py
+++ b/Yank/yamlbuild.py
@@ -712,6 +712,10 @@ class SetupDatabase:
 
         return molecule
 
+    def _check_mol_filename(self, fname):
+        name = os.path.basename(fname)
+        return name.isdigit()
+
     def _setup_molecules(self, *args):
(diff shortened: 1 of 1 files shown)


jchodera:
Thanks for this!

I think the better way to handle this is to ensure that we write an intermediate filename that `tleap` can read, rather than not be able to deal with numeric names.

jchodera:
We'll tackle this issue this week.

Lnaden:
This got fixed in #819 as well, im going to close this PR

Issue #466, opened 2016-08-30
Compute entropies and enthalpies in analyze.py using MBAR


jchodera:
Currently, `analyze` does not use MBAR to compute entropies and enthalpies:

https://github.com/choderalab/yank/blob/master/Yank/analyze.py#L229-L314

We should replace this with a call to [`mbar.computeEntropyAndEnthalpy()`](https://github.com/choderalab/pymbar/blob/master/pymbar/mbar.py#L1096).

Lnaden:
Fixed by #470

PR #470, opened 2016-09-01
Use MBAR for enthalpy and clean up analyze workflow

Lnaden:
Addresses #466 and some todos in the analyze.py script.

* Enthalpy now computed from MBAR instead of just mean energy.
* Changed workflow to reduce duplicate code for decorrelating samples and to prevent having to create multiple MBAR objects.
* Added cascading functions in workflow so if any particular step is forgotten, the pre-requisite steps will be called.
  - Equilibration calculation is not part of cascade yet.

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -110,10 +110,9 @@ def show_mixing_statistics(ncfile, cutoff=0.05, nequil=0):
 
     return
 
-
-def estimate_free_energies(ncfile, ndiscard=0, nuse=None, g=None):
+def extract_ncfile_energies(ncfile, ndiscard=0, nuse=None, g=None):
     """
-    Estimate free energies of all alchemical states.
+    Extract and decorelate energies from the ncfile to gather common data for other functions
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


Lnaden:
I have not been able to test all functionality yet, so there may be a few bugs still.

jchodera:
This looks like a big improvement! 

We should sort out what we want to do with the analysis framework. We've given very little thought to this so far, but will need to make it scale up to analyze a large number of simulations. The current structure isn't very amenable to that.

Just brainstorming now---we can move this to an issue later---but I'd love to see:
* an `Analysis` class that contains all the methods and data, and is initialized with a YANK calculation directory
* a cleaner analysis API---it's OK if this breaks backward-compatibility
* some facility for analyzing a batch of simulations that have been set up in some directory tree

jchodera:
I've just fixed the travis test issue. You should update or rebase from `master` when you get a chance so the travis tests will run.

jchodera:
@Lnaden : I'm going to make some commits to this branch to clean things up a bit.

PR #475, opened 2016-09-07
Fix automatic addition of barostat for non-periodic systems

andrrizzi:
I just prevented `Yank` to add a barostat if the system is not periodic.

**Question:** it looks like OpenMM sets the default box vectors of a newly created system to a cube of side 2.0nm. Are there cases in which we want a non-periodic box? Should we set the vectors to `None` for non-periodic systems?

### DIFF
--- a/Yank/tests/test_yaml.py
+++ b/Yank/tests/test_yaml.py
@@ -1578,6 +1578,7 @@ def test_get_alchemical_path():
     assert len(complex_path) == 7
 
 
+@attr('slow')  # Skip on Travis-CI
 def test_run_experiment_from_amber_files():
     """Test experiment run from prmtop/inpcrd files."""
     complex_path = examples_paths()['bentol-complex']
@@ -1751,21 +1752,21 @@ def test_run_solvation_experiment():
     """Test solvation free energy experiment run."""
     with omt.utils.temporary_directory() as tmp_dir:
         yaml_script = get_template_script(tmp_dir)
+        yaml_script['solvents']['PME']['clearance'] = '14*angstroms'
         yaml_script['systems'] = {
             'system1':
                 {'solute': 'toluene', 'solvent1': 'PME', 'solvent2': 'vacuum',
                  'leap': {'parameters': ['leaprc.gaff', 'leaprc.ff14SB']}}}
-
         protocol = yaml_script['protocols']['absolute-binding']['solvent']
         yaml_script['protocols'] = {
             'hydration-protocol': {
-                'solvent1' : protocol,
-                'solvent2' : protocol
+                'solvent1': protocol,
+                'solvent2': protocol
             }
         }
         yaml_script['experiments'] = {
-            'system' : 'system1',
-            'protocol' : 'hydration-protocol'
+            'system': 'system1',
+            'protocol': 'hydration-protocol'
             }
 
         yaml_builder = YamlBuilder(yaml_script)
@@ -1787,7 +1788,6 @@ def test_run_solvation_experiment():
 
         # Check that solvated phase has a barostat.
         from netCDF4 import Dataset
-        from simtk import openmm
         ncfile = Dataset(os.path.join(output_dir, 'solvent1.nc'), 'r')
         ncgrp_stateinfo = ncfile.groups['thermodynamic_states']
         system = openmm.System()
--- a/Yank/yank.py
+++ b/Yank/yank.py
@@ -331,28 +331,6 @@ class Yank(object):
         atom_indices = alchemical_phase.atom_indices
         alchemical_states = alchemical_phase.protocol
 
-        # If temperature and pressure are specified, make sure MonteCarloBarostat is attached.
-        if thermodynamic_state.temperature and thermodynamic_state.pressure:
-            forces = { reference_system.getForce(index).__class__.__name__ : reference_system.getForce(index) for index in range(reference_system.getNumForces()) }
-
-            if 'MonteCarloAnisotropicBarostat' in forces:
-                raise Exception('MonteCarloAnisotropicBarostat is unsupported.')
-
-            if 'MonteCarloBarostat' in forces:
-                logger.debug('MonteCarloBarostat found: Setting default temperature and pressure.')
-                barostat = forces['MonteCarloBarostat']
-                # Set temperature and pressure.
-                try:
-                    barostat.setDefaultTemperature(thermodynamic_state.temperature)
-                except AttributeError:  # versions previous to OpenMM7.1
-                    barostat.setTemperature(thermodynamic_state.temperature)
-                barostat.setDefaultPressure(state.pressure)
-            else:
-                # Create barostat and add it to the system if it doesn't have one already.
-                logger.debug('MonteCarloBarostat not found: Creating one.')
-                barostat = openmm.MonteCarloBarostat(thermodynamic_state.pressure, thermodynamic_state.temperature)
-                reference_system.addForce(barostat)
-
         # Check the dimensions of positions.
         for index in range(len(positions)):
             n_atoms, _ = (positions[index] / positions[index].unit).shape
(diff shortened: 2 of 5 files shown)


andrrizzi:
I've discovered and fixed some more bugs:

1. equilibration timestep was ignored
2. there was a bug in `_is_periodic()` function that always made return `False`. I've ditched the function and used OpenMM `System.usesPeriodicBoundaryConditions()` now which (correct me if I'm wrong) should test the exact same condition.

As a consequence of 2), we have never changed the cutoff of the fully interacting system and now that is causing problems. I'll keep working on this.

Yank/yank.py
```diff
@@ -365,7 +365,7 @@ def _create_phase(self, thermodynamic_state, alchemical_phase):
             # TODO: Should we warn if cutoff can't be extended enough?
             # TODO: Should we extend to some minimum cutoff rather than the maximum allowed?
             box_vectors = fully_interacting_system.getDefaultPeriodicBoxVectors()
-            max_allowed_cutoff = 0.499 * max([ max(vector) for vector in box_vectors ]) # TODO: Correct this for non-rectangular boxes
+            max_allowed_cutoff = 0.489 * max([max(vector) for vector in box_vectors])  # TODO: Correct this for non-rectangular boxes
```
jchodera: Should this maybe be 
```python
min([max(vector) for vector in box_vectors])
```
in case one box edge is shorter than the others?

andrrizzi:
Yep! I realized it after I pushed. I'm correcting now, thanks!

jchodera:
> As a consequence of 2), we have never changed the cutoff of the fully interacting system, and now I'm having new problems related to that. I'll keep working on this.

Is it the fact that the box shrinks during equilibration with the barostat? If so, there are a few possibilities:
1. Set the scaling factor (currently `0.489`) to be small enough to safely not run into an issue
2. Set (or reset) the scaling factor after an "equilibration" phase
3. Adapt the maximum box size each iteration (which would likely cause problems with MBAR)
4. Fix the cutoff at something sufficiently large, like 16 A, and raise an exception if the box size isn't big enough to accommodate.

The relevant work quantifying the error is [this paper](http://www.choderalab.org/publications/2014/4/26/accurate-and-efficient-corrections-for-missing-dispersion-interactions-in-molecular-simulations).

jchodera:
I'm thinking the last option may be best:

Fix the large cutoff to 16 A by default (where the correction is negligble beyond this), and maybe make this an option in the YAML file?

andrrizzi:
> Is it the fact that the box shrinks during equilibration with the barostat?

No, I was referring to [this error](https://travis-ci.org/choderalab/yank/builds/158527850#L2284). It was just that OpenMM raised an exception on context creation because the cutoff was larger than half the box side. Changing `min` to `max` mostly fixed it.

Now the only error I get is `Abort trap 6` on the test starting with GROMACS files. The error disappears if I don't change the cutoff of the forces in the fully interacting system. Still trying to figure this out.

andrrizzi:
More specifically this test: `nosetests Yank/tests/test_yaml.py:test_run_experiment_from_gromacs_files`

jchodera:
> No, I was referring to this error. It was just that OpenMM raised an exception on context creation because the cutoff was larger than half the box side. Changing min to max mostly fixed it.
> Now the only error I get is Abort trap 6 on the test starting with GROMACS files. The error disappears if I don't change the cutoff of the forces in the fully interacting system. Still trying to figure this out.

When running with a barostat, an exception, segfault, or abort can happen when the box becomes smaller than twice the cutoff because of the barostat. I think we just need to make sure the cutoff is large enough to minimize the artifacts (16A or greater) and the box is large enough that this isn't a problem (e.g. 36A per side or larger).

andrrizzi:
I see! Makes sense. I'll set the max cutoff to 16A and make Yank raise an exception if the minimum box side is smaller than 3 times the maximum cutoff. I'll open an issue to remind us to add the option to control the maximum cutoff.

andrrizzi:
Sorry, I meant 2 times the maximum cutoff.

jchodera:
Maybe 2.5 times? If it's only 2 times larger, if the box shrinks during equilibration, it will still throw an exception or segfault.

andrrizzi:
The new limitation on the box size caused a couple of test starting from amber/gromacs files to fail. We have to update the examples data, but I would do that later, when we take a pass at all of them. I don't want to spend time on an example that we can decide to substitute/modify.

Is it ok if I mark those two tests as `slow` for now and merge? I'd like to restart the hydration free energy calculation tests before the weekend.

jchodera:
> The new limitation on the box size caused a couple of test starting from amber/gromacs files to fail. We have to update the examples data, but I would do that later, when we take a pass at all of them. I don't want to spend time on an example that we can decide to substitute/modify.

The easiest fix here is probably to manually specify the larger box cutoff in the YAML file for these systems once we take a pass through updating them.

> Is it ok if I mark those two tests as slow for now and merge? I'd like to restart the hydration free energy calculation tests before the weekend.

Sounds great!

jchodera:
>  I'd like to restart the hydration free energy calculation tests before the weekend.

The hydration free energies will still be wrong until I finish https://github.com/choderalab/alchemy/pull/39 so you might actually want to hold off on that test.

andrrizzi:
Ok, tests pass! Feel free to merge if you have already reviewed.

Yank/tests/test_yaml.py
```diff
@@ -1751,21 +1752,21 @@ def test_run_solvation_experiment():
     """Test solvation free energy experiment run."""
     with omt.utils.temporary_directory() as tmp_dir:
         yaml_script = get_template_script(tmp_dir)
+        yaml_script['solvents']['PME']['clearance'] = '14*angstroms'
```
jchodera: Is `clearance` the size of the box buffer when constructing a box, or the size of the cutoff used for post-processing?

Yank/yank.py
```diff
@@ -346,48 +346,35 @@ def _create_phase(self, thermodynamic_state, alchemical_phase):
...
             box_vectors = fully_interacting_system.getDefaultPeriodicBoxVectors()
-            max_allowed_cutoff = 0.499 * max([ max(vector) for vector in box_vectors ]) # TODO: Correct this for non-rectangular boxes
+            min_box_dimension = min([max(vector) for vector in box_vectors])
+
+            # Expand cutoff to minimize artifact and verify that box is big enough.
+            # If we use a barostat we leave more room for volume fluctuations or
+            # we risk fatal errors. If we don't use a barostat, OpenMM will raise
+            # the appropriate exception on context creation.
+            # TODO: Make max_allowed_cutoff an option
+            max_allowed_cutoff = 16 * unit.angstroms
```
jchodera: Can you create an issue to make this an option in the YAML file?

Yank/yank.py
```diff
@@ -346,48 +346,35 @@ def _create_phase(self, thermodynamic_state, alchemical_phase):
...
+            if thermodynamic_state.pressure and min_box_dimension < 2.25 * max_allowed_cutoff:
+                raise RuntimeError('Barostated box sides must be at least 36 Angstroms '
```
jchodera: This warning should probably compute the `2.25 * max_allowed_cutoff` value and warn about that so we don't forget to change it when we make `max_allowed_cutoff` an option.

jchodera:
Just reviewed! Feel free to merge once comments are addressed to your satisfaction.

Yank/tests/test_yaml.py
andrrizzi: It is the size of the box buffer when constructing a box.

Yank/yank.py
andrrizzi: Sure!

Yank/yank.py
andrrizzi: Yeah, I'll add this to the issue to make it an option.

Issue #479, opened 2016-09-08
Document how to test working platforms


jchodera:
Add section on checking which platforms are installed and working with `yank platform`.

Issue #476, opened 2016-09-08
Document how to configure YANK with nonstandard CUDA installations


jchodera:
We found we needed to use:
```
export OPENMM_CUDA_COMPILER=$CUDA_HOME/bin/nvcc
export LD_LIBRARY_PATH=$CUDA_HOME/lib64:$LD_LIBRARY_PATH
```

Lnaden:
Documented by #484

Issue #467, opened 2016-08-31
Add instructions to installation docs to test optional OpenEye install


Lnaden:
We should add a section to the OpenEye installation docs to have the user check if they were installed correctly, should they install them.

jchodera:
Better yet, perhaps `yank platforms` or `yank --version` or `yank status` or something should report on whether the OpenEye tools are installed and working correctly.

Lnaden:
I added the check to the `yank selftests` command and online documentation in #484, hopefully this is sufficient tests.

PR #484, opened 2016-09-13
CLI Docs Overhaul and Numerous Doc Updates

Lnaden:
This PR is a series of small changes to the online documentation and a large change to how the command line interface (CLI) help docs are handled.

* CLI now uses a nested series of help string and docs to provide the user with command specific instructions and options instead of just dumping all commands and options at once. Similar to github help docs in that you can do `yank help COMMAND` to get help on the command.
 * Moved individual command documentation into that command's .py file to keep doc and command in same place, usage doc is auto read by the `yank help` file as needed.
* Improved the cli.py file to automatically detect what commands are available by inspecting the available modules inside the commands folder. Any future commands which follow the same structure can be added and should automatically be handled by the main cli.py file
* Added option to selftest for OpenEye check, online documentation has also been updated to explain how you can test the install should fully address #467 
* Added instructions for non-standard CUDA installations to the online instructions, addresses #476
* Added instructions for testing working platforms and what to do if you do not see the platforms you expect, addresses #479 
* **NEW!** Added online documentation which points directly at the `examples/yank-yaml-cookbook/all-options.yaml` file and renders it for online documentation. It is only a raw dump of the file, but it will update in real time as we push changes and we dont have to create a custom .rst file for it. We might want to improve appearance in the future, but its a start for now.


docs/yaml.rst
```diff
@@ -0,0 +1,11 @@
+.. _yaml-head:
+
+YAML Files for Simulation
+*************************
+
+YANK now supports setting up entire simulations with easy to read and write `YAML files <http://www.yaml.org/about.html>`. No longer do you need to remember long command line arguments with many ``--`` flags, you can just write a ``yank.yaml`` file and call ``yank script --yaml=yank.yaml`` to carry out all your preparation and simulations. Below are the valid YAML options. We also have example YANK simulations which use YAML files to run.
+
+.. todo:: This section is a work in progress and not all options are shown yet
+
+.. literalinclude:: ../examples/yank-yaml-cookbook/all-options.yaml
```
andrrizzi: Maybe it's worth linking also `../examples/yank-yaml-cookbook/combinatorial-experiment.yaml`?

andrrizzi:
> Moved individual command documentation into that command's .py

I'm not sure `docopt` supports this since it uses the description to check that the user syntax is correct. It could be that tests are failing because of that.

The rest looks great so far.

Lnaden:
Looks like you are probably right, I need the base version to parse all "--options" as base strings after the fact. I can rework the code to not parse the initial input and rely instead on the `dispatch` variable to see if it failed.

Lnaden:
Might have found a `docopt` way to fix it, we'll see after this new push.

docs/yaml.rst
Lnaden: Just added in most recent push

Lnaden:
These tests are passing now. I made a few small changes based on suggestions, if it looks good though, I'll merge it.

andrrizzi:
Looks good! Just to understand, does docopt still raise an exception when the CLI syntax is wrong with this implementation (I don't think we have a test for this)?

Lnaden:
So it works by checking the basic syntax of `yank COMMAND everything else`, then parses the `COMMAND` to see if its part of the commands in the `commands` folder. If it is a command, it [checks that command's individual syntax](https://github.com/choderalab/yank/blob/551181b58f9b1a32864c641f3c4949fa93dd8d05/Yank/cli.py#L68-L74) and will reject any thing it doesn't expect, telling you how to usue that specific command. Its why I moved each command's individual usage variable to each command's individual file. If it didn't find anything to do, the `dispatch` variable is still False and the default usage message is thrown. 

I tried to model it a bit on how command line git does it errors. It also makes it a bit more tenable moving forward as we can create new commands as needed and drop them in the commands folder with the same structure as other the other <command>.py files.

andrrizzi:
Oh, I see! Nice! Feel free to merge.

PR #488, opened 2016-09-14
[WIP] Python 3.X compatability

Lnaden:
This PR attempts to fix the code so that it compiles and can test on python 3.X. There are no new features in this PR beyond fixing the `print` functions. 

There will probably need to be some changes to the travis tests to get it working correctly.

### DIFF
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -1179,13 +1179,13 @@ class ReplicaExchange(object):
         mbar_citations = """\
         Shirts MR and Chodera JD. Statistically optimal analysis of samples from multiple equilibrium states. J. Chem. Phys. 129:124105, 2008. DOI: 10.1063/1.2978177"""
 
-        print "Please cite the following:"
-        print ""
-        print openmm_citations
+        print("Please cite the following:")
+        print("")
+        print(openmm_citations)
         if self.replica_mixing_scheme == 'swap-all':
-            print gibbs_citations
+            print(gibbs_citations)
         if self.online_analysis:
-            print mbar_citations
+            print(mbar_citations)
 
         return
 
@@ -1519,7 +1519,7 @@ class ReplicaExchange(object):
             # Compute log probability of swap.
             log_P_accept = - (self.u_kl[i,jstate] + self.u_kl[j,istate]) + (self.u_kl[i,istate] + self.u_kl[j,jstate])
 
-            #print "replica (%3d,%3d) states (%3d,%3d) energies (%8.1f,%8.1f) %8.1f -> (%8.1f,%8.1f) %8.1f : log_P_accept %8.1f" % (i,j,istate,jstate,self.u_kl[i,istate],self.u_kl[j,jstate],self.u_kl[i,istate]+self.u_kl[j,jstate],self.u_kl[i,jstate],self.u_kl[j,istate],self.u_kl[i,jstate]+self.u_kl[j,istate],log_P_accept)
+            #print("replica (%3d,%3d) states (%3d,%3d) energies (%8.1f,%8.1f) %8.1f -> (%8.1f,%8.1f) %8.1f : log_P_accept %8.1f" % (i,j,istate,jstate,self.u_kl[i,istate],self.u_kl[j,jstate],self.u_kl[i,istate]+self.u_kl[j,jstate],self.u_kl[i,jstate],self.u_kl[j,istate],self.u_kl[i,jstate]+self.u_kl[j,istate],log_P_accept))
 
             # Record that this move has been proposed.
             self.Nij_proposed[istate,jstate] += 1
@@ -1586,7 +1586,7 @@ class ReplicaExchange(object):
             # Compute log probability of swap.
             log_P_accept = - (self.u_kl[i,jstate] + self.u_kl[j,istate]) + (self.u_kl[i,istate] + self.u_kl[j,jstate])
 
-            #print "replica (%3d,%3d) states (%3d,%3d) energies (%8.1f,%8.1f) %8.1f -> (%8.1f,%8.1f) %8.1f : log_P_accept %8.1f" % (i,j,istate,jstate,self.u_kl[i,istate],self.u_kl[j,jstate],self.u_kl[i,istate]+self.u_kl[j,jstate],self.u_kl[i,jstate],self.u_kl[j,istate],self.u_kl[i,jstate]+self.u_kl[j,istate],log_P_accept)
+            #print("replica (%3d,%3d) states (%3d,%3d) energies (%8.1f,%8.1f) %8.1f -> (%8.1f,%8.1f) %8.1f : log_P_accept %8.1f" % (i,j,istate,jstate,self.u_kl[i,istate],self.u_kl[j,jstate],self.u_kl[i,istate]+self.u_kl[j,jstate],self.u_kl[i,jstate],self.u_kl[j,istate],self.u_kl[i,jstate]+self.u_kl[j,istate],log_P_accept))
 
             # Record that this move has been proposed.
             self.Nij_proposed[istate,jstate] += 1
@@ -1626,7 +1626,7 @@ class ReplicaExchange(object):
         if self.replica_mixing_scheme == 'swap-neighbors':
             self._mix_neighboring_replicas()
         elif self.replica_mixing_scheme == 'swap-all':
-            # Try to use weave-accelerated mixing code if possible, otherwise fall back to Python-accelerated code.
+            # Try to use cython-accelerated mixing code if possible, otherwise fall back to Python-accelerated code.
             try:
                 self._mix_all_replicas_cython()
             except ValueError as e:
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -71,7 +71,7 @@ import numpy as np
 import mdtraj as md
 import netCDF4 as netcdf
 
-from utils import is_terminal_verbose, delayed_termination
+from .utils import is_terminal_verbose, delayed_termination
 
 #=============================================================================================
 # MODULE CONSTANTS
@@ -2019,7 +2019,7 @@ class ReplicaExchange(object):
             The dict to store.
 
         """
-        from utils import typename
+        from .utils import typename
         import collections
         for option_name in options.keys():
             # Get option value.
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -1540,8 +1540,7 @@ class ReplicaExchange(object):
         Attempt to exchange all replicas to enhance mixing, calling code written in Cython.
         """
 
-        from . import mixing
-        from mixing._mix_replicas import _mix_replicas_cython
+        from .mixing._mix_replicas import _mix_replicas_cython
 
         replica_states = md.utils.ensure_type(self.replica_states, np.int64, 1, "Replica States")
         u_kl = md.utils.ensure_type(self.u_kl, np.float64, 2, "Reduced Potentials")
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -2118,32 +2118,29 @@ class ReplicaExchange(object):
         options = dict()
 
         import numpy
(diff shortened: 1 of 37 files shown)


Yank/tests/test_repex.py
```diff
@@ -111,7 +111,7 @@ def test_replica_exchange(mpicomm=None, verbose=True):
 
     """
 
-    if verbose and ((not mpicomm) or (mpicomm.rank==0)): print "Testing replica exchange facility with harmonic oscillators: ",
+    if verbose and ((not mpicomm) or (mpicomm.rank==0)): print("Testing replica exchange facility with harmonic oscillators: ", end = " ")
```
andrrizzi: Would the `end` argument work in python 2?

Lnaden: No, the syntax to supress the new line character changes between 2.X and 3.X, so we way not get it on this one line in either. I'll fix this on the next update.

jchodera: Do we want to migrate those print statements to logging statements?

jchodera: Alternatively, we can just use this idiom for the meantime:
```python
sys.stdout.write(string)
```
which doesn't include a newline by default.

jchodera:
Looks good to me otherwise!

Lnaden:
There appear to be some issues loading CUDA and OpenCL plugins in py 3.X. Anyone know whats going on there? Both the OpenMM and OpenEye tools fail to fetch the correct shared object libs.

Otherwise this is almost ready to go in.

jchodera:
We don't have problems with any of our other tools. Can you point to the specific failures?

Lnaden:
Failure is here:
https://travis-ci.org/choderalab/yank/jobs/161445267#L1925-L2013

jchodera:
The errors are all failing to load `CUDA` and `OpenCL` on travis:
```
Error loading library /home/travis/miniconda/envs/test/lib/plugins/libOpenMMCUDA.so: libcuda.so.1: cannot open shared object file: No such file or directory
Error loading library /home/travis/miniconda/envs/test/lib/plugins/libOpenMMOpenCL.so: libOpenCL.so.1: cannot open shared object file: No such file or directory
Error loading library /home/travis/miniconda/envs/test/lib/plugins/libOpenMMRPMDCUDA.so: libOpenMMCUDA.so: cannot open shared object file: No such file or directory
Error loading library /home/travis/miniconda/envs/test/lib/plugins/libOpenMMDrudeCUDA.so: libOpenMMCUDA.so: cannot open shared object file: No such file or directory
Error loading library /home/travis/miniconda/envs/test/lib/plugins/libOpenMMAmoebaCUDA.so: libOpenMMCUDA.so: cannot open shared object file: No such file or directory
Error loading library /home/travis/miniconda/envs/test/lib/plugins/libOpenMMRPMDOpenCL.so: libOpenCL.so.1: cannot open shared object file: No such file or directory
Error loading library /home/travis/miniconda/envs/test/lib/plugins/libOpenMMDrudeOpenCL.so: libOpenCL.so.1: cannot open shared object file: No such file or directory
```
Travis doesn't have CUDA or OpenCL libraries installed, so this is expected behavior.

Yank/repex.py
```diff
@@ -2089,31 +2118,29 @@ def _restore_dict_from_netcdf(self, ncgrp):
...
+                    option_value = option_type(option_ncvar.getValue())
+                elif (option_ncvar.shape[0] >= 0):
+                    # Handle array types
+                    option_value = np.array(option_ncvar[:], option_type)
+                    # TODO: Deal with values that are actually scalar constants.
+                    # TODO: Cast to appropriate type
+                else:
+                    # Handle iterable types?
+                    # TODO: Figure out what is actually cast here
+                    option_value = option_type(option_ncvar[0])
```
Lnaden: Changes made to ncvar type inputs have been mostly rolled into the function `_convert_netcdf_store_type`. I just want to make sure there is not a type handled by the old `elif` format that would be handled unexpectedly from the new function.

Yank/tests/test_utils.py
```diff
@@ -84,32 +89,35 @@ def test_expand_tree():
...
 
-    # Test named_combinations generator
-    expected_names = set(['2_a', '3_a', '4_a', '2_b', '3_b', '4_b', '2_c', '3_c', '4_c'])
-    assert expected_names == set([name for name, _ in simple_tree.named_combinations(
-                                                       separator='_', max_name_length=3)])
+    # Test named_combinations generator using either order to account for generator randomness
+    expected_names_A = set(['2_a', '3_a', '4_a', '2_b', '3_b', '4_b', '2_c', '3_c', '4_c'])
+    expected_names_B = set(['a_2', 'a_3', 'a_4', 'b_2', 'b_3', 'b_4', 'c_2', 'c_3', 'c_4'])
+    expected_names = [expected_names_A, expected_names_B]
+    assert set([name for name, _ in simple_tree.named_combinations(separator='_', max_name_length=3)]) in expected_names
```
Lnaden: Will reordered string by flipping along the separator cause any issues when used in practice?

Because the Py3 generators return unordered combinations, the string structure in the expected names can be reversed, This `assert` was changed to check for both. 

There are couple more `asserts` in this test which do the same thing, marking them with "string re-order"

Yank/tests/test_utils.py
```diff
@@ -84,32 +89,35 @@ def test_expand_tree():
...
     long_tree = CombinatorialTree({'key1': CombinatorialLeaf(['th#*&^isnameistoolong1',
                                                               'th#*&^isnameistoolong2']),
                                    'key2': CombinatorialLeaf(['test1', 'test2'])})
-    expected_names = set(['test-thisn', 'test-thisn-2', 'test-thisn-3', 'test-thisn-4'])
-    assert expected_names == set([name for name, _ in long_tree.named_combinations(
-                                                       separator='-', max_name_length=10)])
+    expected_names_A = set(['test-thisn', 'test-thisn-2', 'test-thisn-3', 'test-thisn-4'])
+    expected_names_B = set(['thisn-test', 'thisn-test-2', 'thisn-test-3', 'thisn-test-4'])
+    expected_names = [expected_names_A, expected_names_B]
+    assert set([name for name, _ in long_tree.named_combinations(separator='-', max_name_length=10)]) in expected_names
```
Lnaden: String re-order

Yank/tests/test_utils.py
```diff
@@ -84,32 +89,35 @@ def test_expand_tree():
...
-    expected_names = set(['benzene-2HYYpdbfixer', 'benzene-2HYYpdbfixer-2', 'notapath-2HYYpdbfixer',
+    expected_names_A = set(['benzene-2HYYpdbfixer', 'benzene-2HYYpdbfixer-2', 'notapath-2HYYpdbfixer',
                           'benzene-benzene', 'benzene-benzene-2', 'notapath-benzene'])
-    assert expected_names == set([name for name, _ in long_tree.named_combinations(
-                                                       separator='-', max_name_length=25)])
-
+    expected_names_B = set(['2HYYpdbfixer-benzene', '2HYYpdbfixer-benzene-2', '2HYYpdbfixer-notapath',
+                          'benzene-benzene', 'benzene-benzene-2', 'benzene-notapath'])
+    expected_names = [expected_names_A, expected_names_B]
+    assert set([name for name, _ in long_tree.named_combinations(separator='-', max_name_length=25)]) in expected_names
```
Lnaden: String re-order

Yank/utils.py
```diff
@@ -235,6 +235,9 @@ class CombinatorialLeaf(list):
     """List type that can be expanded combinatorially in CombinatorialTree."""
     def __repr__(self):
         return "Combinatorial({})".format(super(CombinatorialLeaf, self).__repr__())
+    def __eq__(self, other):
+        # Order does not mater, so comparisons should not rely on them
+        return collections.Counter(self) == collections.Counter(other)
```
Lnaden: Does the order of the `CombinatorialLeaf` matter when making equality checks?

`CombinatorialLeaf`s generated from `CombinatorialTree`s can have lists pop out in any order, causing equality tests to sometimes fail. I overloaded the equality operator for the `CombinatorialLeaf` instead of making it a list, it makes a type of `set`. (`set` might be a simpler cast here over `collections.Counter`, but it works for now).

Yank/utils.py
```diff
@@ -489,7 +492,14 @@ def _resolve_path(d, path):
...
         """
-        return reduce(lambda d,k: d[k], path, d)
+        # TODO: Make sure this is working
+        # Converted from old py 2.X `reduce` function
+        function = lambda d,k: d[k]
+        it = iter(path)
+        accum_value = d
+        for x in it:
+            accum_value = function(accum_value, x)
+        return accum_value
```
Lnaden: Will this function as originally intended?

`reduce` was removed in Py3, the code here is the what the Py2 docs call the "roughly equivalent" code

Yank/utils.py
```diff
@@ -845,6 +855,170 @@ def camelcase_to_underscore(camelcase_str):
     underscore_str = re.sub(r'([A-Z])', '_\g<1>', camelcase_str)
     return underscore_str.lower()
 
+def quantity_from_string(quantity_str):
```
Lnaden: Should I write explicit tests for this parser and/or make a simpler parser?

This is a powerful converter that can take arbitrary strings and covert them to `units.Quantity`s based on a number of valid operators. Its automatically (although implicitly) tested by a number of the other tests. It can parse more complex strings than any of the other tests or use cases call for right now though.

Lnaden:
All tests are now passing but there are some functionality changes I have made to parts of the code and I want to make sure the changes won't break anything. I would appreciate someone looking over the code review comments I made, and if possible check the rest of the code as well.

Yank/tests/test_utils.py
andrrizzi: Ah! I just remembered something. This could cause problems with resuming if a combinatorial experiment is created in a folder, and then the inverted name is used to resume the experiment later. I'll found a way to fix this.

andrrizzi: Reordering the strings by flipping along the separator would solve it.

Yank/utils.py
andrrizzi: Sounds good. Order should not matter.

Yank/utils.py
andrrizzi: It looks correct to me!

Yank/utils.py
andrrizzi: Yeah, it is probably worth to write a simple test in which you throw few expressions at it in case anybody will end up using its advanced functionalities.

andrrizzi:
Went through the whole changes. Looks great to me!

Yank/tests/test_utils.py
Lnaden: I'm not sure where the re-order would take place. Would it be in the file name check itself? It may be possible to fix this by setting the internal dictionaries of the `CombinatorialTree` as `collections.OrderedDict`, but I'm not sure where to set those in the current code.

andrrizzi: Yeah, probably transforming it into an `OrderedDict` is the fastest option, although it depends on how it is constructed since the YAML script will be originally loaded as a normal `dict` (and I guess the order will be different each time) and then converted.

Another possibility that comes to mind is to order combinatorial leafs in alphabetical order (of the key) and generate names always based on that order. I can open an issue and tackle it later. Resuming in Python 2 will still work for now.

Lnaden: I don't think sorting the leaf itself maters, the problem is the order in which the leaves themselves are returned. Is it possible to tell yaml to use only ordered dictionaries for its construction (I can see that you kind of can in the `yamlbuild.py` file.) Or should we come up with a solution to resume that does not depend on the order of the objects?

Yank/utils.py
Lnaden: I'll make some tests for this in a new PR

Issue #80, opened 2014-07-21
Testing optional dependencies, like OpenEye tools


jchodera:
For now, we'll need to find a way to skip tests for optional dependencies (like OpenEye tools) on travis-ci.

In PR #79, I've added a method to check for OpenEye tool availability and skip the test if not found, but we need to find a way to skip doctests that require optional dependencies that are not installed.

@kyleabeauchamp suggests the SKIP feature (https://docs.python.org/2/library/doctest.html#doctest-options) might be metaprogrammed.

Sage seems to have a more elaborate doctest scheme:
http://www.sagemath.org/doc/developer/doctesting.html

See especially optional tests:
http://www.sagemath.org/doc/developer/doctesting.html#run-optional-tests
which seems to be our use case.

Continuing to investigate.

pgrinaway:
If we really want to test code that includes the optional stuff, we can probably use EC2 or Jenkins, but IMO this is low priority.

Lnaden:
I added a flag to the `yank selftest` command which runs the tests on the OpenEye install that things like `nosetests` won't run them, but the user can. Is this a sufficient solution to this problem?

jchodera:
> I added a flag to the yank selftest command which runs the tests on the OpenEye install that things like nosetests won't run them, but the user can. Is this a sufficient solution to this problem?

What do you think about having `yank selftest` have two modes:

### Default mode (no command-line options specified)
Report on as many things of value as we can quickly test.
* Report YANK version
* Report which OpenMM platforms are available and any plugin load failures
* Report on whether OpenEye tools are installed (if so, which version) and whether license is available
* As we find more things that may be useful in debugging installations onsite, we can add those here.

### Extensive testing mode
If the user specifies `--nosetests`, run all `nosetests` to make sure everything is working all right in this installation.

I don't think the doctests are necessary to run, but we could also have a `--doctests` flag to run those.

Lnaden:
Fixed by #489

PR #489, opened 2016-09-14
[WIP] Make selftest a more helpful

Lnaden:
This PR tries to make the `yank selftest` command more helpful to the user. It runs a few general checks and alerts the users when things are not expected. Changes are based on suggestion in #80.

Options available for running all nosetests and/or doctests. These should not cause an infinite loop since they must be explicitly called when the command is issued and the tests themselves do not do so.

I have not had a chance to test this (yet) but wanted to upload to get feedback on if we want selftest to do more or less while we're making changes to it anyways

### DIFF
--- a/Yank/commands/selftest.py
+++ b/Yank/commands/selftest.py
@@ -17,7 +17,9 @@ import doctest
 import pkgutil
 import sys
 
-from yank import utils
+from yank import utils, version
+from yank.commands import platforms
+import simtk.openmm as mm
 
 
 #=============================================================================================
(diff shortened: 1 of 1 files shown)


Yank/commands/selftest.py
```diff
@@ -27,35 +29,77 @@
...
+        print{  "************************************************")
+        print("\nWarning! There were OpenMM Platform Load Errors!")
+        print{  "************************************************")
+        for e in platform_errors:
+            print(e)
+        print{  "************************************************")
+        print{  "************************************************")
+    # OpenEye checks
+    try:
+        print("OpenEye install Found! Checking install...")
```
jchodera: Can we report the OpenEye version?

Lnaden: Got it!

Lnaden:
OpenEye version check in.

Do we know of a way to force OpenMM to throw Plugin load errors to see if they are being caught correctly?

PR #497, opened 2016-09-22
Added ability for analyze to use the fully interacting state if present

### LINKED ISSUE #495: Make analysis.py reweight to the fully interacting system

Lnaden:
Assumes the fully interacting state is the "true" lambda=1 value.

It's pretty simple, is this all we want to do with the fully interacting state?

Fixes #495 

Also fixes bug with enthalpy calculation now that its using MBAR to get deltaH instead.


andrrizzi:
Awesome!

PR #499, opened 2016-09-24
Add Dockerfile

jchodera:
This is just a little experimental test for adding a Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,19 @@
+# Start with CUDA base image
+FROM kaixhin/cuda
+MAINTAINER John Chodera <john.chodera@choderalab.org>
+
+# Install curl and dependencies for FAH
+RUN apt-get update && apt-get install -y \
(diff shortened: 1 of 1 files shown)


jchodera:
This should be ready for review. After merging, I'd create an experimental docker hub project so that users could pull down `yank` images via
```bash
docker run -i -t jchodera/yank
```
Again, this is experimental, but should work for now.

andrrizzi:
I don't know much about docker, but this looks good to me!

jchodera:
Let me add some notes about using the docker version in a separate PR after
more testing.

andrrizzi:
Sure! Feel free to merge if ready.

jchodera:
Oh! This was never merged.

Is it too late to get this into 0.11.0?

andrrizzi:
Still in time I think! I'm merging this now.

Issue #348, opened 2016-02-26
Generalizing restraints code


jchodera:
@luirink and @davidlmobley both want to contribute code to add new restraint types.  @luirink would like to use `CustomCompoundBondForce` to allow restraints centered in a binding site defined by multiple residues, while @davidlmobley would like to add Boresch-style restraints involving torsions, angles, and bond lengths.

It would be great to get both in the main YANK code, but to make it easier for people to add new restraints, it would be great to modify the `restraints.py` code a bit to make things general.

We already have a `ReceptorLigandRestraint` base class from which other restraint classes are derived, but there are some things that prevent easy generalization:
- These classes are imported by name in [yank.py](https://github.com/choderalab/yank/blob/master/Yank/yank.py#L34)
- The text `restraint_type` selection selects the restraint type in (an explicit `if/elif` structure](https://github.com/choderalab/yank/blob/master/Yank/yank.py#L324-L330)

Instead, what if we simply [found all subclasses of the `ReceptorLigandRestraint` class using Python introspection](http://stackoverflow.com/questions/3862310/how-can-i-find-all-subclasses-of-a-given-class-in-python)?  Each subclass could be forced to define a `getName()` method or `.name` property (or something similar), and we could simply use this to select which class we want.  That way, all someone needs to do is to create a new subclass and give it a unique name and they've succeeded in adding a new restraint type without changing any other code.

Also CCing #305, where we were thinking about standard state issues.

andrrizzi:
> what if we simply found all subclasses of the ReceptorLigandRestraint class using Python introspection?

I guess an alternative could be something like (with appropriate exception handling)
```python
import restraints
...
# the restraint_type option becomes the name of the class instead of 'flat-bottom', 'harmonic', etc.
# or maybe just the first part of the name (i.e. 'Harmonic' for 'HarmonicReceptorLigandRestraint')
LigandRestraintsCls = getattr(restraints, self._restraint_type + 'ReceptorLigandRestraint')
restraints = LigandRestraintsCls(thermodynamic_state, reference_system, ...)
```
we could even use introspection with ``inspect.getargspec(LigandRestraintsCls.__init__)`` to figure out if initialization requires different arguments.

davidlmobley:
@nividic should see this too. Gaetano, this would be relevant to you.

jchodera:
[...]

PR #500, opened 2016-09-26
Initial refactoring of restraints

jchodera:
This is a very crude initial stab at refactoring restraints to address #348.

The basic plan is to have a class hierarchy:
```
ProteinLigandRestraints (abstract)
+- RadiallySymmetricProteinLigandRestraints (abstract)
   +- HarmonicProteinLigandRestraints
   +- FlatBottomHarmonicLigandRestraints
```
Each class has a `name` attribute that corresponds to the name the user should employ in the YAML file (`harmonic` or `flat-bottom` for now).

When a restraint is specified, all subclasses of `ProteinLigandRestraints` are identified in an index and the appropriate subclass is created by `yank.restraints.createRestraints()`. This allows users to specify their own restraint subclasses if they so desire, and this should automatically extend the restraint capabilities provided the class is imported. (We'll figure out later how this would work with the command-line version of YANK, but this will at least make the API version more extensible.)

Some design decisions need to be made:
* Do we want to keep clunky names like `ProteinLigandRestraint` or trim them down a bit? The original motiviation was that we could later add things like intra-ligand or intra-protein restraints used at intermediate lambda values that would reduce the effort needed to converge free energy calculations.
* What public fields do we really want to expose for `ProteinLigandRestraint` objects? We have a lot of these right now:
```
    Public class fields
    -------------------
    name : str
        The name by which this restraint is selected in YAML file.
    state : simtk.openmm.State
        The State object used during initialization
    system : simtk.openmm.System
        The System object used during initialization
    coordinates : simtk.unit.Quantity of dimension (nparticles,3) with units compatible with angstroms
        The system coordinates used during initialization to define the restraint
        TODO: Should this be optional?
    receptor_atoms : list of int
        List of atoms declared to be part of the receptor
    ligand_atoms : list of int
        List of atoms declared to be part of the ligand to be restrained to the receptor
    temperature : simtk.unit.Quantity
        The temperature specified during initialization.
    kT : simtk.unit.Quantity
        Thermal specified during initialization.
    beta : simtk.unit.Quantity
        The inverse temperature specified during initialization.
```
* What public methods do we want to support? Right now we have:
```
getRestraintForce
getRestrainedSystemCopy
getStandardStateCorrection
```
* Should a `ProteinLigandRestraints` object be immutable, and tied to a specific `System`? If so, this simplifies the internals a bit.


jchodera:
(Note that the refactoring is probably very buggy right now.)

I've run into an issue trying to test this locally. After uninstalling yank and then `pip install .`, I see:
```
[LSKI1497:choderalab/yank/yank.choderalab] choderaj% nosetests --nocapture -vv Yank/tests/test_yaml.py 
nose.config: INFO: Ignoring files matching ['^\\.', '^_', '^setup\\.py$']
Failure: ImportError (cannot import name version) ... ERROR

ERROR: Failure: ImportError (cannot import name version)
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/Users/choderaj/github/choderalab/yank/yank.choderalab/Yank/__init__.py", line 9, in <module>
    from . import version #Needed for yank 3.X. TODO: should be from yank import version for clarity.
ImportError: cannot import name version

Ran 1 test in 0.001s

FAILED (errors=1)
```

jchodera:
New restraints to be implemented:
- [ ] `BoreschOrientationalProteinLigandRestraints`
- [ ] `CenterOfMassHarmonicProteinLigandRestraints` (which could just be an option of `HarmonicProteinLigandRestraints` if we had some way to provide parameters)
- [ ] `CenterOfMassFlatBottomProteinLigandRestraints` (which could just be an option of `FlatBottomProteinLigandRestraints` if we had some way to provide parameters)

It might simplify things greatly if we had an easy way to pass extra optional arguments for the restraints class along. Is there some provision for specifying this in the YAML file? What if we changed the syntax from 
```YAML
options:
  restraint_type: flat-bottom                   # Apply a restrain to the ligand to keep it close to the
```
to something that supported optional additional arguments to pass along, like:
```
restraints:
   type: flat-bottom
   r0: 10*angstroms
```

andrrizzi:
Awesome! Thanks for tackling this!

> Each class has a `name` attribute

Just to throw an alternative on the table. We could just let the user specify the name of the class to make it more transparent. Something like:
```yaml
restraint:
  class: HarmonicProteinLigandRestraint
```
or for user custom restraints
```yaml
restraint:
  class: mymodule.MyRestraintClass
```
the advantage of this is that the user can take a look at the API docs of the specific class to understand what it does and how to configure it, especially if we will move to a more complicated YAML syntax.

> Do we want to keep clunky names like `ProteinLigandRestraint` or trim them down a bit?

If we are still plan to add those ligand-ligand restraint I would keep them like this, unless there are alternatives that are as clear as this one.

> What public fields do we really want to expose for ProteinLigandRestraint objects? What public methods do we want to support?

Those you listed look good to me! We can still add more stuff later

> Should a `ProteinLigandRestraints` object be immutable, and tied to a specific System? If so, this simplifies the internals a bit.

+1

> Is there some provision for specifying this in the YAML file?

I like the design you proposed. And maybe we could move the restraint specification from the `options:` to the `systems:` level? I wrote a function in `utils` that convert a function signature into a `schema` specification, so that can be easily integrated.

Lnaden:
> I've run into an issue trying to test this locally. After uninstalling yank and then pip install ., I see:

Odd.. I'm not sure what is happening there but it looks like your nosetests are trying to read the non-installed version of Yank from the local directory instead of an installed version. The version.py file is not built in the downloaded Yank folder when you just do `pip install .` which is probably why its failing.

jchodera:
> The version.py file is not built in the downloaded Yank folder when you just do pip install . which is probably why its failing.

When does `version.py` it get built?

jchodera:
Also, it doesn't look like we have a `Yank/tests/test_restraints.py` unit test for restraints. I guess it would be useful to add this!

Lnaden:
When you run `pip install .` it runs the setup.py file and installs everything in the `pythonX.Y/site-packages/yank` folder, including the `version.py` file. When nostests run on that test file, they should try to do `import yank` which _should_ pull from the site-packages folder, and not the local download folder. I'm not sure why your nosetests are trying to use the download folder, unless pip didn't install yank?

jchodera:
> I like the design you proposed. And maybe we could move the restraint specification from the options: to the systems: level? I wrote a function in utils that convert a function signature into a schema specification, so that can be easily integrated.

That works for me---that's where we specify what goes into a complex, right?

One experiment that may be of interest, for example, is making sure that we can study the dependence of the computed free energy on different parameter choices for restraints.

andrrizzi:
> That works for me---that's where we specify what goes into a complex, right?

Yes. The only problem with this, now that I think about it, is that if you want to specify a limited number of receptor/ligand atoms for the restraints in a combinatorial experiment, you will have to create one system for each combination because the DSL strings will be different for each pair of receptor-ligand.

The only way to avoid this that I can see, is to specify the DSL expression for the atoms to restraints in the `molecules:` section. Other ideas?

jchodera:
Questions:
* Did we decide to change from using restraint names like `harmonic` and `flat-bottom` to the class names?
* Did we want to simplify the class names from 
```
ProteinLigandRestraints (abstract)
+- RadiallySymmetricProteinLigandRestraints (abstract)
   +- HarmonicProteinLigandRestraints
   +- FlatBottomHarmonicLigandRestraints
```
to something shorter?

andrrizzi:
> Did we decide to change from using restraint names like harmonic and flat-bottom to the class names?

It seems to be the best way to generalize it to user-defined restraints.
> Did we want to simplify the class names to something shorter?

That would be great, especially considering the point above. Not sure how though. I don't think we should remove anything from the names that can create ambiguities in future restraints that we could implement.
> What kinds of tests should I add?

Maybe we could leave the test you proposed as some sort of validation test, or slow test. Something to run only once, or rarely, and then have a quick Travis check for a very simple system to make sure we don't break anything while changing the code.

jchodera:
> It seems to be the best way to generalize it to user-defined restraints.

OK, this will simplify things!

How about this for class names:
```
ProteinLigandRestraints (abstract)
+- RadiallySymmetricRestraints (abstract)
   +- Harmonic (with `centroid` or `single-atom` options)
   +- FlatBottomHarmonic (with `centroid` or `single-atom` options)
+- OrientationalRestraints (abstract)
   +- Boresch
```
This way, we can use in the YAML
```YAML
restraints: !Combinatorial [Harmonic, FlatBottomHarmonic, Boresch]
```
> Maybe we could leave the test you proposed as some sort of validation test, or slow test. Something to run only once, or rarely, and then have a quick Travis check for a very simple system to make sure we don't break anything while changing the code.

I can add a more elaborate test and mark it `slow` so travis won't run it.

I won't implement the `centroid` versions now, but will wait for @andrrizzi to add the YAML syntax to support passing options to the restraints class in a future PR.

andrrizzi:
Sounds all great to me!

jchodera:
Looks like I just have to update the restraint terms used in the tests and examples now.

andrrizzi:
Cool! I can do it in this same branch if you're short on time. I have to get familiar with the new restraints anyway.

jchodera:
Let me take a quick pass through this first!

jchodera:
I don't think there's a test to programmatically set up a YANK simulation from a simple YAML file that includes restraints. If I wanted to create one---say for a host-guest system in vacuum or implicit solvent---how would I set that up?

andrrizzi:
With the current YAML syntax, you can only do something like
```python
from yamlbuild import YamlBuilder

yaml_script = """
options:
  restraints: harmonic
...  # molecules, solvents, protocols, experiments
"""
yaml_builder = YamlBuilder(yaml_script)
yaml_builder.build_experiment()  # run both setup and experiments
```
There is currently no way to go through YamlBuilder with a system picked from `openmmtools.testsystems` though. Setting up the simulation purely from the API could use some work.

Maybe it makes more sense to add these once the new YAML syntax for the restraints will be in place. I'll surely add some integration tests in `test_yaml.py` anyway.

jchodera:
> There is currently no way to go through YamlBuilder with a system picked from openmmtools.testsystems though. Setting up the simulation purely from the API could use some work.

Hm, I guess we don't have any testsystems specifically in YANK right now (e.g. in `Yank/tests/systems/`)---only the `examples/`, which we're hoping to migrate out to a separate package.

What about using the paths from where `openmmtools.testsystems` examples are stored as PDB and `mol2` files? I could encode them into the YAML file, like this:
```python
from yamlbuild import YamlBuilder
from openmmtools.testsystems import get_data_filename
filepaths = {
   'benzene_mol2_filepath' : get_data_filename('data/benzene-toluene-implicit/benzene.mol2'),
   'toluene_mol2_filepath' : get_data_filename('data/benzene-toluene-implicit/toluene.mol2'),
   }
yaml_script = """
---
options:
  minimize: no
  verbose: no
  output_dir: .
  number_of_iterations: 2
  restraint_type: !Combinatorial [Harmonic, FlatBottom]
  temperature: 300*kelvin
  softcore_beta: 0.0

molecules:
  benzene:
    filepath: %(benzene_mol2_filepath)s
    antechamber:
      charge_method: bcc
  toluene:
    filepath: %(toluene_mol2_filepath)s
    antechamber:
      charge_method: bcc

solvents:
  OBC2:
    nonbonded_method: NoCutoff
    implicit_solvent: OBC2

systems:
  benzene-toluene:
    receptor: toluene
    ligand: benzene
    solvent: OBC2
    leap:
      parameters: [leaprc.ff14SB, leaprc.gaff]

protocols:
  absolute-binding:
    complex:
      alchemical_path:
        lambda_electrostatics: [1.0, 0.0, 0.0]
        lambda_sterics:        [1.0, 1.0, 0.0]
    solvent:
      alchemical_path:
        lambda_electrostatics: [1.0, 0.0, 0.0]
        lambda_sterics:        [1.0, 1.0, 0.0]

experiments:
  system: benzene-toluene
  protocol: absolute-binding
""" % filepaths
yaml_builder = YamlBuilder(yaml_script)
yaml_builder.build_experiment()  # run both setup and experiments
```
Maybe you could help me figure out how much of that YAML file I can trim out because the default options are fine?

Note that the benzene and toluene `mol2` files are not at the paths specified above. For some reason, those didn't get checked into `openmmtools`, so I might add them to `yank/tests/systems/` instead.

jchodera:
I've not implemented a test like this, where I set up a combinatorial experiment, but I think I have the YAML syntax wrong, since I get the error:
```
ERROR: Test the restraints in a protein:ligand system.
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/choderaj/github/choderalab/yank/yank.choderalab/Yank/tests/test_restraints.py", line 156, in test_protein_ligand_restraints
    yaml_builder = YamlBuilder(yaml_script)
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/yank/yamlbuild.py", line 1200, in __init__
    self.parse(yaml_source)
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/yank/yamlbuild.py", line 1248, in parse
    yaml_content.get('metadata', {}))))
  File "/Users/choderaj/miniconda/lib/python2.7/site-packages/yank/yamlbuild.py", line 1485, in _validate_options
    raise YamlParseError(str(e))
YamlParseError: parameter restraint_type=Combinatorial(['FlatBottom', 'Harmonic']) is incompatible with FlatBottom
yank.yamlbuild: ERROR: parameter restraint_type=Combinatorial(['FlatBottom', 'Harmonic']) is incompatible with FlatBottom
--------------------- >> end captured logging << ---------------------
```
The YAML file is [here](https://github.com/choderalab/yank/blob/07a560f31daf8930d59f6657cbbf86cd7b21480e/Yank/tests/test_restraints.py#L109-L155).

andrrizzi:
I think that's a bug I introduced when I updated the YAML syntax validation! I can fix it, but probably better in a separate PR? It should go away if you remove the `Combinatorial` option for now.

jchodera:
@andrrizzi and @Lnaden : What would you think about also allowing the `Topology` to be passed to the restraints class? Would that be easy to do?

jchodera:
It looks like I have to switch to a smaller protein-ligand test system to run on travis.

jchodera:
@andrrizzi : I'd like to set up a host-guest system for the tests, but I think both host and guest would start as a Tripos mol2 file (with charges). Can you help me set up a YAML file for that?

Lnaden:
What advantage would passing the Topology object in be? We also don't necessarily have a proper .top object, or at least won't if we ever move off Amber file

jchodera:
The `Topology` is immediately necessary if we want to use MDTraj to measure equilibrium bond lengths, angles, and torsions instead of writing our own functions to measure these. More generally, though, it could be used to more intelligently select atoms for restraints.

It looks like the `Yank._create_phase()` function does have `alchemical_phase.reference_topology` available, so I am now passing that.

jchodera:
I now have what appears to be a working draft of the Boresch restraints coded up.

I would still like to:
- [ ] switch from T4 lysozyme to a host-guest system, if this could be made efficient
- [ ] implement a slow calculation to confirm the standard state correction is computed correctly

jchodera:
I'm happy to punt those additional items to another PR if switching the T4 test system to vacuum makes it fast enough to pass travis tests.

jchodera:
OK, should be good to review and merge. I can do the extra steps in a follow-up PR:
* switch from T4 lysozyme to a host-guest system, if this could be made efficient
* implement a slow calculation to confirm the standard state correction is computed correctly

jchodera:
This seems to hang when I'm setting up `abl-imatinib-explicit` to use Boresch restraints, but the only output I get is:
```
[chodera@gpu-3-8 abl-imatinib-explicit]$ ./run.sh
Running simulation...
2016-10-17 00:16:44,795: MPI disabled.
2016-10-17 00:16:44,807: Setting up the systems for Abl, STI and pm
```
Is there more debug output generated somewhere during setup?

jchodera:
Looks like the default `antechamber` `sqm` AM1-BCC charge calculation for imatinib was taking forever. I switched it to OpenEye and it took much less time. Maybe we should have some logging output for progress through setup?

jchodera:
Just found a problem with the Boresch restraints that I want to fix.

jchodera:
Phew, done!

jchodera:
It's failing on py 3.x because I think there may be some weird differences with how abstract subclasses are identified. Will debug locally.

jchodera:
Looks like py2 and py3 handle `ABCMeta` in different ways:
http://stackoverflow.com/questions/35673474/using-abc-abcmeta-in-a-way-it-is-compatible-both-with-python-2-7-and-python-3-5

Addressing this now.

jchodera:
I think I've fixed it now.

jchodera:
Tests pass! Go ahead with review!

Yank/restraints.py
```diff
@@ -174,71 +363,77 @@ def _computeRadiusOfGyration(self, coordinates):
...
-        mm : simtk.openmm compliant interface, optional, default=None
-           If specified, use an alternative OpenMM API implementation.
-           Otherwise, use simtk.openmm.
+           Index of second atom in restraint
 
         Returns
         -------
-        force : simtk.openmm.CustomBondForce
-           A restraint force object
+        force : simtk.openmm.HarmonicBondForce
```
andrrizzi: This says an `HarmonicBondForce` is returned, but it's a `CustomBondForce`.

Yank/restraints.py
```diff
@@ -174,71 +363,77 @@ def _computeRadiusOfGyration(self, coordinates):
...
+    def getRestraintForce(self):
         """
-        Compute the standard state correction for the arbitrary restraint energy function.
+        Returns a new Force object that imposes the receptor-ligand restraint.
 
         Returns
         -------
-        DeltaG : float
-           Computed standard-state correction in dimensionless units (kT)
+        force : simtk.openmm.HarmonicBondForce
```
andrrizzi: Same here.

Yank/restraints.py
```diff
@@ -257,99 +452,52 @@ def integrand(r):
...
+
         final_time = time.time()
         elapsed_time = final_time - initial_time
         logger.debug("restraints: _computeStandardStateCorrection: %.3f s elapsed" % elapsed_time)
 
         # Return standard state correction (in kT).
         return DeltaG
 
-    def getRestraintForce(self, mm=None):
+    def getRestraintForce(self):
```
andrrizzi: Duplicate method.

Yank/restraints.py
```diff
@@ -551,20 +698,295 @@ def _determineBondParameters(self):
...
+        """
+        super(Boresch, self).__init__(topology, state, system, positions, receptor_atoms, ligand_atoms)
+
+        # Select atoms to be used in restraint.
+        self.restraint_atoms = self._select_restraint_atoms(positions, receptor_atoms, ligand_atoms)
+
+        # Determine restraint parameters
+        self._determine_restraint_parameters(state, system, positions)
+
+    def _is_collinear(self, positions, atoms, THRESHOLD=0.9):
```
andrrizzi: Lower case `THRESHOLD`?

Yank/restraints.py
```diff
@@ -551,20 +698,295 @@ def _determineBondParameters(self):
...
+        """Report whether any sequential vectors in a sequence of atoms are collinear to within a given dot product threshold.
+
+        Parameters
+        ----------
+        positions : simtk.unit.Quantity of natoms x 3 with units compatible with nanometers
+            Reference positions to use for imposing restraints
+        atoms : array-like, dtype:int, length:natoms
+            A seris of indices of atoms
+        THRESHOLD : float, optional, default=0.9
+            Atoms are not colinear if their sequential vector separation dot products are less than THRESHOLD
```
andrrizzi: Typo `colinear`

Yank/restraints.py
```diff
@@ -551,20 +698,295 @@ def _determineBondParameters(self):
...
+        -------
+        result : bool
+            Returns True if any sequential pair of vectors is collinear; False otherwise.
+
+        """
+        result = False
+        for i in range(len(atoms)-2):
+            v1 = positions[atoms[i+1],:] - positions[atoms[i],:]
+            v2 = positions[atoms[i+2],:] - positions[atoms[i+1],:]
+            innerprod = np.dot(v1, v2)
```
andrrizzi: Do we need to make `v1` and `v2` unit vectors before taking the inner product? Otherwise the threshold is relative to the distances between atoms.

Yank/restraints.py
```diff
@@ -551,20 +698,295 @@ def _determineBondParameters(self):
...
+
+        Notes
+        -----
+        The current algorithm simply selects random subsets of receptor and ligand atoms
+        and rejects those that are too close to collinear. Future updates can further refine
+        this algorithm.
+
+        """
+        if (len(receptor_atoms) < 3) or (len(ligand_atoms) < 3):
+            raise Exception('Both receptor_atoms (len %d) and ligand_atoms (len %d) must contain at least three atoms.' % (len(receptor_atoms), len(ligand_atoms)))
```
andrrizzi: Too general `Exception`.

Yank/restraints.py
```diff
@@ -551,20 +698,295 @@ def _determineBondParameters(self):
...
+        accepted = False
+        while (not accepted):
+            # Select three random atoms from the receptor and ligand
+            restraint_atoms = random.sample(receptor_atoms, 3) + random.sample(ligand_atoms, 3)
+            # Reject collinear sets of atoms.
+            accepted = not self._is_collinear(positions, restraint_atoms)
+
+        return restraint_atoms
+
+    def _determine_restraint_parameters(self, state, system, positions):
```
andrrizzi: These three arguments `state`, `system`, and `positions` are not used.

Yank/restraints.py
```diff
@@ -551,20 +698,295 @@ def _determineBondParameters(self):
...
+class OrientationDependentRestraint(ReceptorLigandRestraint):
+    """
+    Impose an orientation-dependent restraint between ligand and protein to prevent ligand from
+    changing to a different binding mode.
+
+    This restraint strength is controlled by a global context parameter called 'lambda_restraints'.
+
+    """
+    energy_function = ''  # energy function to use in computation of restraint
+    bond_parameter_names = []  # list of bond parameters that appear in energy function above
```
andrrizzi: Do we need `bond_parameter_names` in orientation restraints?

examples/abl-imatinib-explicit/yank.yaml
```diff
@@ -4,38 +4,50 @@ options:
...
 molecules:
   Abl:
     filepath: input/2HYY-pdbfixer.pdb
   STI:
     filepath: input/STI02.mol2
+    openeye:
+      quacpac: am1-bcc
     antechamber:
-      charge_method: bcc
+      charge_method: null
```
andrrizzi: Do we want to have main examples be dependent on optional packages (OpenEye)?

jchodera:
> In general, we should change all the camelCase methods in restraints.py. I think it's the only module that uses them. But this can be done in a separate PR.

I'm happy to do this now since we're changing everything anyway!

Yank/restraints.py
jchodera: Right now, only `positions` are used. Later on, we will have an option to run a short equilibrium simulation and use this to derive restraint parameters. But I figured I should at least get the basic implementation in first.

Yank/restraints.py
jchodera: Good point!

Yank/restraints.py
jchodera: No, I've removed this.

We *do* still need a way to write these parameters to the NetCDF file.

Yank/restraints.py
jchodera: Fixed!

Yank/restraints.py
andrrizzi: I see, thanks for clarifying. `positions` is not used anyway because you already have `self.positions`.

Yank/restraints.py
```diff
@@ -50,41 +140,58 @@ class ReceptorLigandRestraint(object):
...
+    ligand_atoms : list of int
+        List of atoms declared to be part of the ligand to be restrained to the receptor
+    temperature : simtk.unit.Quantity
+        The temperature specified during initialization.
+    kT : simtk.unit.Quantity
+        Thermal specified during initialization.
+    beta : simtk.unit.Quantity
+        The inverse temperature specified during initialization.
+
+    TODO: Should these all really be public data fields?
```
jchodera: Good catch!

I think we need to figure out if the restraint parameters should be publicly accessible, and if so, how. Should they be accessible like `restraint.K`, or maybe a dict like `restraint.parameters`, or via getters and setters?

andrrizzi: Good question. It sounds reasonable to allow the control of these parameters to avoid catches (for example, the current scheme to compute `K` automatically in the Harmonic restraints fails when there are only 2 atoms).

I would prefer having them accessible as public attributes `restraint.K`. It's just cleaner than a dictionary, and they can be substituted by properties in case we need to manipulate them on get/set. We can still have a read-only `restraint.parameters` property that builds a dictionary from the parameters in case that's handy to store these in the `nc` file.

jchodera:
OK, I think I've addressed everything!

andrrizzi:
Awesome! I'll merge as soon as tests pass.

Issue #493, opened 2016-09-21
Design a better test for test_order_phases in test_yaml.py


Lnaden:
Because of random ordering:

`assert list(parsed['absolute-binding'].keys()) != ordered_phases`

can fail because the two lists are in the same order. We need to come up with a better control test.

andrrizzi:
This should be fixed when #491 is implemented. I'm tackling it now.

andrrizzi:
Implemented in #503 .

Issue #491, opened 2016-09-21
Force CombinatorialTree to generated names for combinations in predetermined order


andrrizzi:
Follows from discussion in #488 .

We need to modify the method `CombinatorialTree.named_combinations()` to return names based on the alphabetical order of the `CombinatorialLeaf` key so that experiment names will be consistent on resuming.

andrrizzi:
This should also make many tests simpler and resolve #490 .

andrrizzi:
Implemented in #503 .

Issue #496, opened 2016-09-22
Add more complete documentation for every options/key that can be set in the `.yaml` files.


Lnaden:

Lnaden:
Possibly template off the [conda meta.yaml documentation](http://conda.pydata.org/docs/building/meta-yaml.html) so not a single feature is missed.

jchodera:
I think it would be awesome if we could hyperlink the cookbook recipe. I don't think we can do that in github-flavored markdown, but sphinx may have a way to do that.

Lnaden:
Well right now the cookbook recipe is just a pointer at the file which just loads a raw dump of the text in the file. Since the cookbook is not formatted in any kind of sphinx way, I'm not sure what we would be pointing at besides the literal file.

jchodera:
Obviously, we'd have to change that and format the links by hand, but this is an alternative to the conda meta.yaml-style documentation, which I'm not actually a huge fan of.

Lnaden:
What don't you like about the conda meta.yaml style? Are you suggesting doing something like this:

One page has the links, each category is its own page which is a subpage of the main one.

![markup](http://i.imgur.com/tnKx0CA.png)

Lnaden:
@jchodera would this work instead of having everything on one page?

jchodera:
Sure!

Lnaden:
Reopening due to typo in pr which closed it on accident

PR #503, opened 2016-09-26
Ordered combinatorial tree

### LINKED ISSUE #490: Fix test_yaml.test_expand_molecules test
Current test relies on literal strings to compare molecule names in YAML structure. Py3 however uses generators to get keys from dictionaries, these result in non-regular name patterns that make it hard to compare nested dictionaries as the test is currently structured. 

The test is disabled for now (called `notest_expand_molecules test` in the file) until we come up with a better test.

andrrizzi:
Fix #490, #491, and #493.

Combinatorial leaves are iterated always in the same order (alphabetical order of the string tree path), so that the automatically generated name of the combination is the same. This fixes resuming simulations (since the name of the experiment folder could be different for different runs) and simplifies tests.

### DIFF
--- a/Yank/tests/test_utils.py
+++ b/Yank/tests/test_utils.py
@@ -12,7 +12,6 @@ Test various utility functions.
 import textwrap
 
 import openmoltools as omt
-import simtk.unit as unit
 from schema import Schema
 from openmmtools import testsystems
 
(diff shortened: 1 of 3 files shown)


andrrizzi:
Whops! I've just noticed I included the wrong issue number in the last commit so #496 will be closed automatically when this is merged. I don't want to force push a commit amend because I think that breaks history. I'll just reopen the issue when this is merged.

andrrizzi:
Tests pass! This is ready to be merged.

Lnaden:
Did you switch the tests back to how they were or is this a fragment of unmerged commits from the master?

If you changed them back, awesome! Then this looks good to pull

andrrizzi:
I changed them back! Thanks for reviewing! Merging.

PR #506, opened 2016-09-27
Fix issue with fully interacting state in analysis

Lnaden:
Fix array size recasting and deconvolution of the fully interacting state.


jchodera:
Can we print out some more information about which state is fully
interacting and fix the numbering in the other states so they match the
actual alchemical state definitions? We still use that matrix to optimize
spacing of alchemical states right now.

Lnaden:
Right now only `analyze` re-orders the states, and only in that function. Is there something post-analyze which uses the output `u_kln` from the analyze file?

jchodera:
I thought there were some matrices that get printed out, but I think we don't number the `Deltaf_ij` matrix anymore, so maybe this isn't a real concern.

Lnaden:
Only the stochastic tests (fixed in another PR) failed. @andrrizzi if this looks good to you, merge it so we can overhaul the analyze script in general.

PR #511, opened 2016-10-05
Fully interacting state is now only expanding the LJ force cutoff.

Lnaden:
The fully_interacting_state has been replaced by
reference_LJ_state and reference_LJ_expanded_state.

fully_interacting_energy =
  reference_energy -
  reference_LJ_energy +
  reference_LJ_expanded_energy

This will break backwards compatiblity

### DIFF
--- a/Yank/tests/test_yank.py
+++ b/Yank/tests/test_yank.py
@@ -23,6 +23,7 @@ This code is licensed under the latest available version of the GNU General Publ
 
 
 from openmmtools import testsystems
+import mdtraj
 from mdtraj.utils import enter_temp_directory
 
 from nose import tools
@@ -30,6 +31,7 @@ import netCDF4 as netcdf
 
 from yank.repex import ThermodynamicState
 from yank.pipeline import find_components
+
 
 from yank.yank import *
 
--- a/Yank/yank.py
+++ b/Yank/yank.py
@@ -409,12 +409,12 @@ class Yank(object):
                     pass
 
         # Construct thermodynamic states
+        reference_state = copy.deepcopy(thermodynamic_state)
+        reference_state.system = reference_system
         reference_LJ_state = copy.deepcopy(thermodynamic_state)
         reference_LJ_expanded_state = copy.deepcopy(thermodynamic_state)
         reference_LJ_state.system = reference_system_LJ
         reference_LJ_expanded_state.system = reference_system_LJ_expanded
-        # Assign reference_system only after we are done copy thermodynamic state
-        thermodynamic_state.system = reference_system
 
         # Compute standard state corrections for complex phase.
         metadata['standard_state_correction'] = 0.0
@@ -501,6 +501,7 @@ class Yank(object):
         simulation.create(thermodynamic_state, alchemical_states, positions,
                           displacement_sigma=self._mc_displacement_sigma, mc_atoms=mc_atoms,
                           options=repex_parameters, metadata=metadata,
+                          reference_system = reference_system,
                           reference_LJ_state = reference_LJ_state,
                           reference_LJ_expanded_state = reference_LJ_expanded_state)
 
--- a/Yank/yank.py
+++ b/Yank/yank.py
@@ -365,7 +365,7 @@ class Yank(object):
                 for particle in range(force.getNumParticles()):
                     q, sigma, epsilon = force.getParticleParameters(particle)
                     force.setParticleParameters(particle, 0, sigma, epsilon)
-                for exception in force.getNumExceptions():
+                for exception in range(force.getNumExceptions()):
                     particle1, particle2, chargeprod, epsilon, sigma = force.getExceptionParameters(exception)
                     force.setExceptionParameters(particle1, particle2, 0, sigma, epsilon)
             else:
--- a/Yank/yank.py
+++ b/Yank/yank.py
@@ -1,8 +1,8 @@
 #!/usr/local/bin/env python
 
-#=============================================================================================
+# =============================================================================================
 # MODULE DOCSTRING
-#=============================================================================================
+# =============================================================================================
 
 """
 Yank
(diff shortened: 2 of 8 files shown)


jchodera:
Looks like you have a failure:
```
  File "/home/travis/miniconda/envs/test/bin/yank", line 11, in <module>
    load_entry_point('yank==0.11.0', 'console_scripts', 'yank')()
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.11.0-py2.7-linux-x86_64.egg/yank/cli.py", line 74, in main
    dispatched = getattr(commands, command).dispatch(command_args)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.11.0-py2.7-linux-x86_64.egg/yank/commands/script.py", line 59, in dispatch
    yaml_builder.build_experiments()
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.11.0-py2.7-linux-x86_64.egg/yank/yamlbuild.py", line 1281, in build_experiments
    self._run_experiment(combination, output_dir)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.11.0-py2.7-linux-x86_64.egg/yank/yamlbuild.py", line 2249, in _run_experiment
    yank.create(thermodynamic_state, *phases)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.11.0-py2.7-linux-x86_64.egg/yank/yank.py", line 271, in create
    self._create_phase(thermodynamic_state, phase)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.11.0-py2.7-linux-x86_64.egg/yank/yank.py", line 362, in _create_phase
    force = reference_system_LJ.getForce(forceIndex)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/simtk/openmm/openmm.py", line 7290, in getForce
    return _openmm.System_getForce(self, *args)
Exception: Assertion failure at System.cpp:108.  Index out of range
```

Yank/tests/test_yank.py
```diff
@@ -31,6 +32,7 @@
 from yank.repex import ThermodynamicState
 from yank.pipeline import find_components
 
+
 from yank.yank import *
```
jchodera: Can we avoid using `from ... import *` anywhere in our codebase?

Lnaden: I can work to fix that, I think its the only one and its been in since before I started working on the code. I wanted to make sure the changes I made for functionality changes worked before I did code cleanup.

If we want a full code cleanup, pycharm is more than happy to tell me all our PEP 8 violations 😈

andrrizzi: Does it cause problems with python 3? Because I find it very convenient to use it (only) in tests since in `test_yank.py` we want to test every single public element contained in `yank.py`.

jchodera:
Hm, good point.

Let's worry about that later, I guess.

Lnaden:
Once these tests pass and code review is done, this can be merged.

Just a note, this will break backwards compatibility with anything that used the old `fully_interacting_state` to store energy. Analyze will not be affected since the name of the energy variable is the same, but for debugging and resuming old simulations, this will break reading those .nc files

andrrizzi:
Ah! The parameter name in `ModifiedHamiltonianExchange.create(reference_state, ...)` is incredibly misleading! It's called `reference_state` but it's really the `alchemical_state` (see [here](https://github.com/choderalab/yank/blob/master/Yank/yank.py#L424) and [here](https://github.com/choderalab/yank/blob/master/Yank/yank.py#L472)), so if I understand the current implementation, `fully_interacting_energy` doesn't include the reciprocal space energy.

It's probably a very good idea to rename the parameter.

Lnaden:
Ah, okay, I think I see the problem, let me make another pass at it

jchodera:
> Just a note, this will break backwards compatibility with anything that used the old fully_interacting_state to store energy. Analyze will not be affected since the name of the energy variable is the same, but for debugging and resuming old simulations, this will break reading those .nc files

Can you elaborate? Which part has changed that will break compatibility?

I don't think backwards-compatibility is a big concern pre-1.0.

Lnaden:
The old .nc files saved serialized systems called "fully_interacting_system" and the properties derived from it. Since there is not a single system which is used to compute the fully interacting energy, the data saved in files (and read back in) are slightly different now is all. This will only really affect simulations that are resumed.

Lnaden:
Okay, tests are passing. I changed a fair amount so it will need another review. @andrrizzi double check to make sure I have changed everything as expected.

andrrizzi:
I'm on it!

Yank/yank.py
```diff
@@ -359,33 +355,66 @@ def _create_phase(self, thermodynamic_state, alchemical_phase):
...
+                try:
+                    if force.getCutoffDistance() < max_allowed_cutoff:
+                        force.setCutoffDistance(max_allowed_cutoff)
+                except:
+                    pass
+                try:
+                    if force.getCutoff() < max_allowed_cutoff:
+                        force.setCutoff(max_allowed_cutoff)
+                except:
+                    pass
```
andrrizzi: Are these try/except clauses due to different OpenMM versions?

This is implementation works perfectly, but now that we only have `NonbondedForce`s at this point, we could even do something like
```python
for force in reference_system_LJ)expanded.getForces():
  try:
    if force.getCutoffDistance() < max_allowed_cutoff:
      force.setCutoffDistance(max_allowed_cutoff)
  except AttributeError:
    if force.getCutoff() < max_allowed_cutoff:
      force.setCutoff(max_allowed_cutoff)
```
This way if we messed something above this code we can catch it.

Yank/yank.py
```diff
@@ -431,7 +460,7 @@ def _create_phase(self, thermodynamic_state, alchemical_phase):
             integrator = openmm.VerletIntegrator(1.0 * unit.femtosecond)
             context = openmm.Context(alchemical_system, integrator)
             context.setPositions(positions[0])
-            for alchemical_state in alchemical_states:
+            for index, alchemical_state in enumerate(alchemical_states):
```
andrrizzi: `index` not used?

Yank/yank.py
Lnaden: Its not for different versions, but they are for different forces. I wrote it in 2 separate try's to kind of generalize it. Future proofing and whatnot, also incase we want to treat each force differently. Secondly, if we change what we call "The fully interacting state" at some point, it gives a bit less coding later

Yank/yank.py
Lnaden: The index only appears in the `raise` statement, the original code actually never assigned something to `index`. It would have been a wash in the end since an exception got raised, but lets raise the one we actually want to.

PR #512, opened 2016-10-07
Fixing the doc build pointer

Lnaden:
This should now point to the correct version of the index.html. I hope

This may require more fixes to figure out how to get s3 to actually build the file. 

Do we know of a way to get travis to push to s3 to test without a merge? Better question: do we have a test page that isn't public, or at least not linked to on s3 based on PRs for us to test?


jchodera:
No on an automated way, though I can give you access to S3 so you can try to push directly from your machine.

Lnaden:
Hmm... that might work temporarily, but then travis's nightly build will overwrite it. I would also be weary of pushing test pages directly too often in so it doesn't get cached by search engines in a broken state.

PR #517, opened 2016-10-11
Expand the switching distance to match the new expanded cutoff.

### LINKED ISSUE #515: Ensure we enlarge the switch distance as well as the cutoff
Currently, we [enlarge the cutoff](https://github.com/choderalab/yank/blob/master/Yank/yank.py#L401-L407), but we don't enlarge the switch distance at the same time. We should also use [`setSwitchingDistance()`](http://docs.openmm.org/7.0.0/api-python/generated/simtk.openmm.openmm.NonbondedForce.html#simtk.openmm.openmm.NonbondedForce.setSwitchingDistance) to a larger value.

Lnaden:
Fixes #515


andrrizzi:
Looks good to me!

Issue #481, opened 2016-09-08
Have YANK print out which platform is being used


jchodera:
@Lnaden and I realized YANK doesn't seem to print out which platform is *actually* being used during execution. This would be very helpful to know for debugging purposes.

andrrizzi:
I'll take care of this too in the same PR I'm working on for fixing #441 .

PR #518, opened 2016-10-11
Add platform options

### LINKED ISSUE #441: Fix platform and precision options in Yank
Currently they can be passed but they are not attached to anything.

andrrizzi:
Fix #441 and #481 . Should be ready for review if tests pass.

- Introduce the ability to specify the platform for the calculation through the `yank script` command and `Yank` API.
- Add `precision` option in YAML.
- Set the platform default precision mode to `mixed` when possible. Note that some devices do not support double or mixed precision for the OpenCL platform. In that case the default value is `single`.

Currently I've implemented the system so that the driver class takes care of the platform configuration (i.e. correcting the default precision to `mixed`), and passes a `Platform` object to `Yank`. If nothing is passed, the OpenMM default is used.

### DIFF
--- a/Yank/commands/script.py
+++ b/Yank/commands/script.py
@@ -33,9 +33,6 @@ Description:
 Required Arguments:
   -y, --yaml=FILEPATH           Path to the YAML script specifying options and/or how to set up and run the experiment.
 
-General Options:
-  --platform=PLATFORM           OpenMM Platform to use (Reference, CPU, OpenCL, CUDA)
-
 """
 
 #=============================================================================================
@@ -54,12 +51,11 @@ def dispatch(args):
     """
     if args['--yaml']:
         yaml_path = args['--yaml']
-        platform_name = args['--platform']
 
         if not os.path.isfile(yaml_path):
             raise ValueError('Cannot find YAML script "{}"'.format(yaml_path))
 
-        yaml_builder = YamlBuilder(yaml_source=yaml_path, platform_name=platform_name)
+        yaml_builder = YamlBuilder(yaml_source=yaml_path)
         yaml_builder.build_experiments()
         return True
 
--- a/Yank/commands/script.py
+++ b/Yank/commands/script.py
@@ -25,13 +25,16 @@ usage = """
 YANK script
 
 Usage:
-  yank script (-y=FILEPATH | --yaml=FILEPATH)
+  yank script (-y=FILEPATH | --yaml=FILEPATH) [--platform=PLATFORM]
 
 Description:
   Set up and run free energy calculations from a YAML scrpit. All options can be specified in the YAML script. 
 
 Required Arguments:
   -y, --yaml=FILEPATH           Path to the YAML script specifying options and/or how to set up and run the experiment.
+
+General Options:
+  --platform=PLATFORM           OpenMM Platform to use (Reference, CPU, OpenCL, CUDA)
 
 """
 
@@ -51,11 +54,12 @@ def dispatch(args):
     """
     if args['--yaml']:
         yaml_path = args['--yaml']
+        platform_name = args['--platform']
 
         if not os.path.isfile(yaml_path):
             raise ValueError('Cannot find YAML script "{}"'.format(yaml_path))
 
-        yaml_builder = YamlBuilder(yaml_source=yaml_path)
+        yaml_builder = YamlBuilder(yaml_source=yaml_path, platform_name=platform_name)
         yaml_builder.build_experiments()
         return True
 
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -604,7 +604,6 @@ class ReplicaExchange(object):
         # Set default options.
         # These can be changed externally until object is initialized.
         self.platform = platform
-        self.platform_name = None
         self.integrator = None # OpenMM integrator to use for propagating dynamics
 
         # Initialize keywords parameters and check for unknown keywords parameters
@@ -902,28 +901,6 @@ class ReplicaExchange(object):
 
         return
 
-    def _determine_fastest_platform(self, system):
-        """
-        Determine fastest OpenMM platform for given system.
-
-        Parameters
-        ----------
-        system : simtk.openmm.System
-           The system for which the fastest OpenMM Platform object is to be determined.
-
-        Returns
-        -------
-        platform : simtk.openmm.Platform
-           The fastest OpenMM Platform for the specified system.
-
-        """
-        timestep = 1.0 * unit.femtoseconds
-        integrator = openmm.VerletIntegrator(timestep)
-        context = openmm.Context(system, integrator)
-        platform = context.getPlatform()
-        del context, integrator
-        return platform
-
     def _initialize_create(self):
         """
         Initialize the simulation and create a storage file, closing it after completion.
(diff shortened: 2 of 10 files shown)


Lnaden:
Looking over it now

jchodera:
The `CPU` platform doesn't support `mixed`:
```
ERROR: Test that YamlBuilder select the fastest platform available when unspecified.
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.11.1-py2.7-linux-x86_64.egg/yank/tests/test_yaml.py", line 1497, in test_select_fastest_platform
    yaml_builder._configure_platform(platform_precision='single')
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.11.1-py2.7-linux-x86_64.egg/yank/yamlbuild.py", line 2132, in _configure_platform
    "only 'mixed' is supported.".format(platform_precision))
RuntimeError: CPU platform does not support precision model 'single';only 'mixed' is supported.
```
https://travis-ci.org/choderalab/yank/jobs/166822788

jchodera:
cc: https://github.com/pandegroup/openmm/issues/1626

andrrizzi:
@jchodera That's a bug in the tests, the error doesn't come from OpenMM. I'll fix it after lunch, but you can close the OpenMM issue.

Yank/commands/script.py
```diff
@@ -25,14 +25,17 @@
 YANK script
 
 Usage:
-  yank script (-y=FILEPATH | --yaml=FILEPATH)
+  yank script (-y=FILEPATH | --yaml=FILEPATH) [--platform=PLATFORM]
```
Lnaden: Why is this its own option instead of just part of the script? Having a redundant way to do the same action when not meant as a backup will confuse people.

Yank/yank.py
```diff
@@ -131,7 +131,7 @@ class Yank(object):
         'mc_displacement_sigma': 10.0 * unit.angstroms
     }
 
-    def __init__(self, store_directory, mpicomm=None, **kwargs):
+    def __init__(self, store_directory, mpicomm=None, platform=None, **kwargs):
```
Lnaden: This should really be thrown in with the `**kwargs`` and not carried around as its own call

Yank/yamlbuild.py
```diff
@@ -2171,8 +2286,11 @@ def _run_experiment(self, experiment, experiment_dir):
         utils.config_root_logger(exp_opts['verbose'], os.path.join(results_dir, exp_name + '.log'),
                                  self._mpicomm)
 
+        # Configure platform and precision
+        self._configure_platform(exp_opts['precision'])
+
         # Initialize simulation
-        yank = Yank(results_dir, mpicomm=self._mpicomm, **yank_opts)
+        yank = Yank(results_dir, mpicomm=self._mpicomm, platform=self._platform, **yank_opts)
```
Lnaden: `platform` should be part of `yank_opts`

Yank/repex.py
```diff
@@ -570,7 +570,7 @@ class ReplicaExchange(object):
     # Options to store.
     options_to_store = ['collision_rate', 'constraint_tolerance', 'timestep', 'nsteps_per_iteration', 'number_of_iterations', 'equilibration_timestep', 'number_of_equilibration_iterations', 'title', 'minimize', 'replica_mixing_scheme', 'online_analysis', 'show_mixing_statistics']
 
-    def __init__(self, store_filename, mpicomm=None, mm=None, **kwargs):
+    def __init__(self, store_filename, mpicomm=None, platform=None, mm=None, **kwargs):
```
Lnaden: Roll `platform` into `**kwargs`

Yank/yamlbuild.py
```diff
@@ -1177,14 +1178,17 @@ class YamlBuilder:
     def yank_options(self):
         return self._isolate_yank_options(self.options)
 
-    def __init__(self, yaml_source=None):
+    def __init__(self, yaml_source=None, platform_name=None):
```
Lnaden: Can we not make platform part of the options in the YAML script?

jchodera: I think we decided that the platform is something that will very much depend on _where_ you are running the simulation, so wanted to make that part of the command-line API instead.

Yank/commands/script.py
andrrizzi: I thought we decided in #441 to specify precision in the YAML file and the platform as part of the command. I can definitely put both of them in the YAML file (in fact it's going to make the implementation easier) if we changed our minds on this.

jchodera: precision should be specified in the YAML file since we may want to regress over precision choices.

andrrizzi: Oh, I see your confusion. This is not redundant. There is no way to specify the platform name in they YAML script.

Yank/yank.py
andrrizzi: The main problem is that stuff in `kwargs` are automatically added by `YamlBuilder` as available options, so if I do it the platform name will be specified both in the command line and in the YAML file. Unless I start adding exceptions to the way we treat options, which is something I'd like to avoid.

Beside that, in general my opinion is that we make too much use of `kwargs`, and I believe that we should move away from it when possible. I think it will make the code and its documentation much clearer.

Yank/yamlbuild.py
andrrizzi: That's exactly why I added it as part of the constructor and not part of the parsed options.

Yank/yamlbuild.py
andrrizzi: See above.

Yank/yamlbuild.py
Lnaden: That may be, but it just feels like a very bad habit to start in on. What do we do when we run into another location dependent option? I don't want to get just keep adding keywords if we can help it. 

There may be a more fundamental paradigm shift here we may want to talk about. Each of the YANK components are getting complex enough that trying to run them through a pure api or command line may soon be come untenable. e.g. the `yank prepare` CLI is nearly unusable already. I think we may want to consider burying some internals from users going forward. But this is a larger discussion to have in a different place

Yank/repex.py
andrrizzi: Same thing. If I put `platform` in `kwargs` in `ReplicaExchange`, it's going to be automatically included in the YAML syntax.

Yank/yamlbuild.py
andrrizzi: As far as concern `YamlBuilder`, I don't think there's a way around it. If we don't want the user to specify an option to it through the YAML script, we have to pass it as a keyword somewhere else in the API. I don't think it's going to be many anyway (still talking about `YamlBuilder` only) since most can be specified in the YAML script.

I agree with you and I'm a little uneasy sometimes with how we handle the myriad of options in the API. We can talk about it in the next huddle maybe?

andrrizzi: The only solution to this, is to *not* have command line arguments at all, and have everything in the YAML script only, which I'm still open too. I don't have a strong opinion about it.

Lnaden: Deal! I'll approve these changes otherwise.

Lnaden: Hmm, I actually dont see where it tests the fastest platform on the fly, there is the formal nosetest for it, but I don't see one outside that

jchodera: There's not a great solution to this, but using `Platform.getSpeed()` sounds like the best approach for now. We could later add checks to see if double precision is required/supported, but it may be best to work to improve the OpenMM `Platform` API instead so we don't have to do this within YANK.

andrrizzi: I see. I'm on it!

andrrizzi:
Done! I've also updated the documentation about the new two YAML arguments on the model set up by @Lnaden. If there are no objections (and tests pass), I'd like to merge this by tonight, so that I can run the hydration free energy validation overnight.

andrrizzi:
@Lnaden for now I left the `platform` parameter in `Yank` and `ReplicaExchange` constructor. We can talk about it during the next huddle and correct it in a different PR, if that's ok.

Lnaden:
Looks good to me, we have to fix that 1 test though, can be done in a different pr

andrrizzi:
Great! I think I'll have time tonight to fix the py3 and squish it into this PR. I'll merge as soon as tests pass. Thanks!

andrrizzi:
Tests pass! Merging.

PR #529, opened 2016-10-14
Analyze compute the free energy difference between wrong states

### LINKED ISSUE #528: Analyze compute the energy difference between the wrong states
The function `analyze.analyze()` extracts from the ncfile the number of states without taking into account the presence of the fully interacting state, and ends up computing the free energy difference between the fully interacting state and the next to last state for each phase [here](https://github.com/choderalab/yank/blob/master/Yank/analyze.py#L501-L508)
```python
# Accumulate free energy differences
entry = dict()
entry['DeltaF'] = Deltaf_ij[0,nstates-1]
entry['dDeltaF'] = dDeltaf_ij[0,nstates-1]
entry['DeltaH'] = DeltaH_i[0,nstates-1]
entry['dDeltaH'] = dDeltaH_i[0,nstates-1]
entry['DeltaF_restraints'] = DeltaF_restraints
data[phase] = entry
```
In general, since we treat the fully interacting state as special, we need to be careful during the analysis.

I'll soon open a PR with the fix.

andrrizzi:
Fix #528 . Should be ready for review. I would refactor `analyze.py` and add the tests in a separate PR, if you agree.

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -500,10 +500,10 @@ def analyze(source_directory):
 
             # Accumulate free energy differences
             entry = dict()
-            entry['DeltaF'] = Deltaf_ij[0,nstates-1]
-            entry['dDeltaF'] = dDeltaf_ij[0,nstates-1]
-            entry['DeltaH'] = DeltaH_i[0,nstates-1]
-            entry['dDeltaH'] = dDeltaH_i[0,nstates-1]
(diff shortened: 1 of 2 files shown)


Lnaden:
Can you change the ISRELEASED flag?

Lnaden:
If you change that, you can merge with aproval

andrrizzi:
Right! Thanks for catching it!

PR #536, opened 2016-10-21
Removing Examples from main YANK repo

Lnaden:
- Only specific files needed from examples kept and stored in `data` directory.
- Many tests re-written to point to the specific files
- Deactivated all `yank prepare` tests in preparation for removing this feature
 * Also removed due to explicit pointers to `examples` dir
- All explicit pointers to `examples` removed
- Examples are no longer coppied to `share/yank`


Lnaden:
Tests are passing, this is ready for code "review." May be a bit tricky to sift through since over 1M lines were deleted, but most of that was the old examples dir disappearing.

andrrizzi:
Looks good to me!

jchodera:
Thanks, guys!

PR #542, opened 2016-11-03
Attempting fix at ambermini 16 issues

### LINKED ISSUE #541: Update tests and dependencies for AmberMini 16
Some tests are failing as I think they are looking for old FF files which got moved in AmberMini v16. 

We will need to update tests, and version dependency of ambermini in YANK.

I'll make a PR soon for this.

Lnaden:
Patches in AmberMini v16 requirements and points moved FF files

Fixes #541

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,8 @@ script:
   - conda config --add channels https://conda.anaconda.org/omnia/label/dev
   # Install OpenEye toolkit
   - conda install --yes --quiet pip
-  - pip install $OPENEYE_CHANNEL openeye-toolkits && python -c "import openeye; print(openeye.__version__)"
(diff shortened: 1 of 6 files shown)


Lnaden:
These tests are now passing, approval needed

jchodera:
Looks good to me.

In future updates, I wonder if we should pick a mix of forcefields for the tests, or perhaps add a test specifically to try a few forcefields for a small system.

Lnaden:
I don't think thats something that should go into functionality tests for YANK. We mainly want the tests to make sure everything works. I think comparing FFs in YANK should be handled somewhere else, doing it in tests may add a long time to the tests themselves.

jchodera:
OK, works for me.

PR #550, opened 2016-11-09
Fix restraints for periodic systems and expanded state. YAML syntax future-proof.

### LINKED ISSUE #245: Extend restraints to explicit solvent case
Currently, restraints only are implemented for the implicit solvent case. They ought to be extended to handle the explicit solvent case. Are there any plans to do this? 

As I understand it from @nividic in my group, the challenge is that the custom bonded force currently used for restraints will only work in nonperiodic systems, so the implementation for explicit solvent with cutoffs will need to use a different mechanism (custom nonbonded force, perhaps). 

In my view, clearly this needs to be done, because as far as I know:
- No one has looked at the relative efficiency of running with restraints vs running without restraints
- The (old) data I have from ~2006 suggests that restraints will make some aspects substantially more efficient
- Conventionally we've argued (and I still think) that using restraints avoids problems with ligand stickiness at odd points in the protein at intermediate lambda values which slow convergence

Have you guys looked at these issues in implicit solvent and concluded that restraints are not needed in explicit solvent for some reason? I tried searching on here for info along those lines but didn't find anything. 

If this hasn't already been checked, it seems like the two next logical steps would be:
a) In implicit solvent, compare efficiency of running with restraints versus running without restraints
b) Implement restraints for explicit solvent (@nividic says he thinks he knows how it could be done, so we could probably get involved with this if necessary). 

I'll also raise a separate issue for another aspect I'd like to discuss relating to restraints.

### LINKED ISSUE #523: Do not pass fully interacting state to repex for NoCutoff scheme
I didn't test this, but looking at the code I think we are still computing the energies for `reference_system`, `LJ_only_system` and `LJ_only_expanded_system` also for `NoCutoff` nonbonded method. We should probably pass only `None` to `simulation.create()` for these parameters if that's the case.

### LINKED ISSUE #538: Create versioning system for YAML syntax
Currently there's no mechanism to check YAML syntax version, but I think we should add it before 1.0 release.

andrrizzi:
Fix restraints:
- Fix #245: Add restraint also if the system is periodic.
- Fix #523: Do not reweight to the expanded state systems in implicit solvent/vacuum.
- Fix a bug where the restraint was added also to the expanded state.

Future-proof YAML syntax:
- `restraint` specification has been moved from the `options` to the `experiment` section, and has a nested `type` (example: `restraint: {type: Harmonic}`)
- dropped support for specifying epik selection with `epik: 0` in favor of `epik: {select: 0}`
- Fix #538: add YAML `version` for backward compatibility

### DIFF
--- a/Yank/commands/prepare.py
+++ b/Yank/commands/prepare.py
@@ -396,7 +396,7 @@ def dispatch_binding(args):
     yank = Yank(store_dir, **options)
     if args['--restraints']:
         restraint_type = args['--restraints']
-    yank.create(thermodynamic_state, *alchemical_phases, restraint_type)
+    yank.create(thermodynamic_state, *alchemical_phases, restraint_type=restraint_type)
 
     # Dump analysis object
     analysis = [[alchemical_phases[0].name, 1], [alchemical_phases[1].name, -1]]
--- a/Yank/commands/prepare.py
+++ b/Yank/commands/prepare.py
@@ -71,7 +71,7 @@ Simulation options:
   --platform=PLATFORM           OpenMM Platform to use (Reference, CPU, OpenCL, CUDA)
   --precision=PRECISION         OpenMM Platform precision model to use (for CUDA or OpenCL only, one of {mixed, double, single})
   --pressure=PRESSURE           Pressure for simulation (in atm, or simtk.unit readable string) [default: 1*atmospheres]
-  --restraints=TYPE             Restraint type to add between protein and ligand in implicit solvent (harmonic, flat-bottom) [default: flat-bottom]
+  --restraints=TYPE             Restraint type to add between protein and ligand in implicit solvent (Harmonic, FlatBottom)
   --temperature=TEMPERATURE     Temperature for simulation (in K, or simtk.unit readable string) [default: 298*kelvin]
 
 
@@ -357,8 +357,6 @@ def dispatch_binding(args):
         options['number_of_equilibration_iterations'] = int(args['--equilibrate'])
     if args['--online-analysis']:
         options['online_analysis'] = True
-    if args['--restraints']:
-        options['restraint_type'] = args['--restraints']
     if args['--randomize-ligand']:
         options['randomize_ligand'] = True
     if args['--minimize']:
@@ -396,7 +394,9 @@ def dispatch_binding(args):
 
     # Create new simulation.
     yank = Yank(store_dir, **options)
-    yank.create(thermodynamic_state, *alchemical_phases)
+    if args['--restraints']:
+        restraint_type = args['--restraints']
+    yank.create(thermodynamic_state, *alchemical_phases, restraint_type)
 
     # Dump analysis object
     analysis = [[alchemical_phases[0].name, 1], [alchemical_phases[1].name, -1]]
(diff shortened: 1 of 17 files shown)


andrrizzi:
Still work in progress. I want to add also the `version` key.

andrrizzi:
This should be ready for review! In particular:

@jchodera could you double check if I'm [handling correctly the standard state correction](https://github.com/choderalab/yank/blob/future-proof-yaml/Yank/yank.py#L447-L470)?

@Lnaden I did my best to update the docs on the restraints with the new syntax, but it would be great if you could double check. Also, I'm not so sure where to document the new `version` keyword for YAML.
```yaml
version: 1.0  # Optional, if not specified, YamlBuilder.HIGHEST_VERSION is used
options:
  ...
experiments:
  ...
```
I think we can start the T4 Lysozyme validation after merging this!

jchodera:
> @jchodera could you double check if I'm handling correctly the standard state correction?

Looks correct to me!

Yank/commands/prepare.py
```diff
@@ -71,7 +71,7 @@
   --platform=PLATFORM           OpenMM Platform to use (Reference, CPU, OpenCL, CUDA)
   --precision=PRECISION         OpenMM Platform precision model to use (for CUDA or OpenCL only, one of {mixed, double, single})
   --pressure=PRESSURE           Pressure for simulation (in atm, or simtk.unit readable string) [default: 1*atmospheres]
-  --restraints=TYPE             Restraint type to add between protein and ligand in implicit solvent (harmonic, flat-bottom) [default: flat-bottom]
+  --restraints=TYPE             Restraint type to add between protein and ligand in implicit solvent (Harmonic, FlatBottom)
```
jchodera: Didn't we deprecate the `prepare` command? Or is that happening in a different PR?

Yank/pipeline.py
```diff
@@ -281,7 +281,7 @@ def prepare_phase(positions_file_path, topology_file_path, ligand_dsl, system_op
             err_msg = 'Found periodic box in inpcrd file but nonbondedMethod is NoCutoff'
     else:
         if system_options['nonbondedMethod'] != openmm.app.NoCutoff:
```
jchodera: This should check for all of the periodic nonbonded methods, right?
```python
periodic_nonbonded_methods = [openmm.app.CutoffPeriodic, openmm.app.PME, openmm.app.Ewald]
if system_options['nonbondedMethod'] in periodic_nonbonded_methods:
```
Otherwise, we may miss `CutoffNonPeriodic`, which we don't want to recommend but should still be allowed for benchmarking purposes.

Yank/tests/test_cli.py
```diff
@@ -109,6 +109,8 @@ def test_script_yaml():
         experiments:
             system: system
             protocol: absolute-binding
+            restraint:
+                type: FlatBottom
```
jchodera: Is it worth testing
```YAML
            restraint:
                type: !Combinatorial [FlatBottom, Harmonic, Boresch]
```
?

Yank/commands/prepare.py
andrrizzi: Not yet officially, so I figured I kept this info up-to-date.

Yank/pipeline.py
andrrizzi: Right! I'll fix it.

Yank/tests/test_cli.py
andrrizzi: Ah! The Boresch restraint is already included in the test you wrote for restraints, and the syntax for combinatorial restraints is tested in `test_yaml.py`, so we should be good.

andrrizzi:
Tests pass! Merging.

PR #555, opened 2016-11-11
Resume phases in deterministic order.

### LINKED ISSUE #554: Resume phases in deterministic order for MPI parallel processes
`Yank.resume()` looks in the store directory for `.nc` files of the two phases, but we don't define an order in which they are executed. With Python 2 dictionary implementation, this order was random but deterministic, so parallel MPI processes were always running the same phase. With Python 3 this got broken. I'll open a PR with the fix soon.

andrrizzi:
Fix #554 . Ready for review.

### DIFF
--- a/Yank/yank.py
+++ b/Yank/yank.py
@@ -218,7 +218,7 @@ class Yank(object):
         # If no phases specified, construct a list of phases from the filename prefixes in the store directory.
         if phases is None:
             phases = utils.find_phases_in_store_directory(self._store_directory)
-        self._phases = phases.keys()
+        self._phases = sorted(phases.keys())  # sorting ensures all MPI processes run the same phase
 
         # Construct store filenames.
         self._store_filenames = { phase : os.path.join(self._store_directory, phase + '.nc') for phase in self._phases }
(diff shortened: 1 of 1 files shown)


Lnaden:
Looks good to me if all the tests pass

Yank/yank.py
```diff
@@ -218,7 +218,7 @@ def resume(self, phases=None):
         # If no phases specified, construct a list of phases from the filename prefixes in the store directory.
         if phases is None:
             phases = utils.find_phases_in_store_directory(self._store_directory)
-        self._phases = phases.keys()
+        self._phases = sorted(phases.keys())  # sorting ensures all MPI processes run the same phase
```
jchodera: Alternatively, we could use `_mpicomm.bcast()` to sync these between processes, but this sounds like it should work fine.

andrrizzi: Good idea! I guess we'll have to switch to this solution if we'll ever want to force a particular order. For now, I'd say this code is simpler. Also, it may be refactored during the API overhaul with the introduction of single-phase `AlchemicalPhase` objects.

I'll merge as soon as tests pass then!

PR #556, opened 2016-11-11
Laundry List of Doc and Typo fixes

Lnaden:
Fixes a number of documentation issues, will need approved.


Lnaden:
Also added ability for `yank selftest` to try and determine the Compute Mode of detected CUDA GPUs. Read-only operation.

README.md
```diff
@@ -25,6 +25,37 @@ Examples are available in the [`yank-examples` repository](http://github.com/cho
...
 ```
 
+Requirements
+------------
+
+**Operating System**
+* Linux - 64bit (Ubuntu 12.04+, RHEL 6+, etc) 
+* OS X 10.7 or later (10.10.3 or later for OpenCL)
+
+**Python Requirements (installed automatically through conda**
```
andrrizzi: Close parenthesis (funny how this feels terribly important after you learn programming)

jchodera: Actually, let's ditch the Requirements. This is just going to become out of date immediately, and we **don't** want people installing these manually. It also clutters our main landing page.

jchodera:
Thanks for the updates!

I'm still not sure we want to list the prerequisites, since we already have three places where we have two places where we have to keep these in sync and up to date inside this repo (`setup.py`, `devtools/conda-recipe/meta.yaml` and the `omnia` recipe).

It looks like [`mdtraj` has ditched specification of requirements in `setup.py`](https://github.com/mdtraj/mdtraj/blob/master/setup.py#L282-L288), which may help reduce the maintenance burden. But maybe we could just link to the [requirements in `devtools/conda-recipe/meta-yaml`](https://github.com/choderalab/yank/blob/master/devtools/conda-recipe/meta.yaml#L12-L53) instead of manually reprinting all the requirements in the `README.md`, just to keep things from getting out of sync?

Lnaden:
I did pull them from the `README.md`, the whole section is gone. Do you think we should pull them from the online docs and the setup.py file?

jchodera:
Discussed this in person: I thik we should just link to the `devtools/conda-recipe/meta.yaml` as the authoritative list of dependencies.

Lnaden:
I figured out how to make the "requirements" section on the docs point directly at the `meta.yaml` file. Now there is truly a singular source of requirements across all doc pages.

There is still a separate list in `setup.py`, but we can come back to that.

PR #564, opened 2016-11-15
Switch to using periodic boundary restraints

Lnaden:
Restraints should now use periodic images for computing distances. 
Did some PEP8 cleanup while I was in there

Now requires OpenMM 7.1.0 minimum


jchodera:
Awesome, looks great---thanks!

PR #565, opened 2016-11-15
Fix YAML syntax for null restraint

andrrizzi:
This allows to specify in combinatorial experiments
```yaml
experiments:
  ...
  restraint:
    type: !Combinatorial [FlatBottom, Harmonic, null]
```
Ready for review.


andrrizzi:
Added a fix for #567 to avoid setting periodicity when restraining a nonperiodic system.

PR #566, opened 2016-11-16
Fix extract trajectory and adds image molecules

andrrizzi:
Ready for review.

- Fix a bug where `start`, `end` and `skip` frames where ignored when extracting a replica trajectory and not a state.
- Fix #564: automatically detects if the system is periodic, sets box vectors, and adds an `--imagemol` parameter to `yank analyze extract-trajectory` to apply periodic boundary conditions to molecules positions.

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -561,7 +561,7 @@ def analyze(source_directory):
 
 def extract_trajectory(output_path, nc_path, state_index=None, replica_index=None,
                        start_frame=0, end_frame=-1, skip_frame=1, keep_solvent=True,
-                       discard_equilibration=False):
+                       discard_equilibration=False, image_molecules=False):
     """Extract phase trajectory from the NetCDF4 file.
 
     Parameters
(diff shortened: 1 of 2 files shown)


Yank/analyze.py
```diff
@@ -604,9 +604,20 @@ def extract_trajectory(output_path, nc_path, state_index=None, replica_index=Non
     try:
         nc_file = netcdf.Dataset(nc_path, 'r')
 
+        # Extract topology and system serialization
```
Lnaden: Would it make more sense to just store a bool in the NC file to get the ``isPeriodic`` flag? Because this seems like it will pull a number of resources to do. It might be okay for now since we should avoid trying to store EVERYTHING in the NC file when we can just as well check the properties from reloading the system.

andrrizzi: Yeah, it felt weird to deserialize the system only to know if it's periodic, but it only takes few seconds (the extraction of the trajectory takes in the order of minutes), so I didn't think that changing the `.nc` format would be worth.

Lnaden: Fair enough, looks good then

PR #569, opened 2016-11-17
Fix standard state calculation approximation for PBC bonds

Lnaden:
Simulation will still generate use the PBC bonds for restraint, however the standard state correction which is based on estimating an analytical solution now does not (to correctly compute integral out to r=inf while also using OpenMM Contexts to get energy)

Wrote a test to make sure that what OpenMM computes is close to analytical solution.

Some small PEP8's as I saw them.

### DIFF
--- a/Yank/tests/test_restraints.py
+++ b/Yank/tests/test_restraints.py
@@ -16,7 +16,7 @@ Test restraints module.
 import tempfile
 import shutil
 import math
-import unittest
+import numpy as np
 
 from nose.plugins.attrib import attr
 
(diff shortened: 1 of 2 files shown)


Yank/tests/test_restraints.py
```diff
@@ -1,33 +1,38 @@
...
 
 import tempfile
 import shutil
+import math
+import numpy as np
 
 from nose.plugins.attrib import attr
 
 import yank.restraints
+from openmmtools.testsystems import ThermodynamicState
```
andrrizzi: Can we use the ThermodynamicState in `repex` here?

jchodera:
Can you also log `sigma` in `Harmonic`?

PR #570, opened 2016-11-19
Use expanded cutoff for fully interacting and noninteracting states

Lnaden:
Removes old way of computed expanded cutoff in favor of computing free energy relative to expanded cutoffs at both the fully interacting and non-interacting systems.

Phew!

### DIFF
--- a/Yank/sampling.py
+++ b/Yank/sampling.py
@@ -90,14 +90,12 @@ class ModifiedHamiltonianExchange(ReplicaExchange):
 
         """
         super(ModifiedHamiltonianExchange, self).__init__(store_filename, **kwargs)
-        self._reference_context = None
-        self.reference_state = None
-        self.reference_LJ_state = None
-        self._reference_LJ_context = None
-        self.reference_LJ_expanded_state = None
-        self._reference_LJ_expanded_context = None
+        self._fully_interacting_expanded_context = None
+        self.fully_interacting_expanded_state = None
+        self._noninteracting_expanded_context = None
+        self.noninteracting_expanded_state = None
 
-    def create(self, base_state, alchemical_states, positions, displacement_sigma=None, mc_atoms=None, options=None, metadata=None, reference_state=None, reference_LJ_state=None, reference_LJ_expanded_state=None):
+    def create(self, base_state, alchemical_states, positions, displacement_sigma=None, mc_atoms=None, options=None, metadata=None, fully_interacting_expanded__state=None, noninteracting_expanded_state=None):
         """
         Initialize a modified Hamiltonian exchange simulation object.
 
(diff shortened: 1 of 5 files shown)


Yank/sampling.py
```diff
@@ -90,14 +90,12 @@ def __init__(self, store_filename, **kwargs):
...
-        self.reference_LJ_expanded_state = None
-        self._reference_LJ_expanded_context = None
-
-    def create(self, base_state, alchemical_states, positions, displacement_sigma=None, mc_atoms=None, options=None, metadata=None, reference_state=None, reference_LJ_state=None, reference_LJ_expanded_state=None):
+        self._fully_interacting_expanded_context = None
+        self.fully_interacting_expanded_state = None
+        self._noninteracting_expanded_context = None
+        self.noninteracting_expanded_state = None
+
+    def create(self, base_state, alchemical_states, positions, displacement_sigma=None, mc_atoms=None, options=None, metadata=None, fully_interacting_expanded__state=None, noninteracting_expanded_state=None):
```
andrrizzi: Here you have the typo in `fully_interacting_expanded__state` (two underscores) that is causing the tests to fail.

Lnaden: Damn. If you want to fix that? This is going to take longer than i hoped

andrrizzi: Sure! No problem.

Yank/yank.py
```diff
@@ -480,6 +460,33 @@ def _create_phase(self, thermodynamic_state, alchemical_phase, restraint_type):
...
+            # We dont know which force has which variable, so try all
+            for force in noninteracting_expanded_system.getForces():
+                try:
+                    number_global = force.getNumGlobalParameters()
+                    for parameter_index in range(number_global):
+                        global_name = force.getGlobalParameterName(parameter_index)
+                        if global_name in alchemical_parameters:
+                            force.setGlobalParameterDefaultValue(parameter_index, 0)
+                except:
+                    pass
```
andrrizzi: Ah! Sorry, I didn't understand that this was the goal here. We can easily do this by something like
```
noninteracting_expanded_system = copy.deepcopy(alchemical_system)
alchemical_state = alchemical_states[-1]
AbsoluteAlchemicalFactory.perturbSystem(noninteracting_expanded_system, alchemical_state)
```
which does the same thing basically without the need to hardcode setting everything to 0.

Lnaden: Well.... kind of, it would set the restraint to 1 as well. Which... do we want that or not? I'm thinking we might actually want that

andrrizzi:
Everything else looks great! Feel free to merge when tests pass.

Yank/sampling.py
```diff
@@ -261,32 +256,29 @@ def _store_thermodynamic_states(self, ncfile):
...
+            ncvar_temperatures[0] = self.fully_interacting_expanded_state.temperature / temperature_unit
             # Pressures
-            if self.reference_state.pressure is not None:
+            if self..fully_interacting_expanded_state.pressure is not None:
                 ncvar_pressures = ncgrp_stateinfo.createVariable('pressures', 'f', ('scalar',))
                 setattr(ncvar_pressures, 'units', 'atm')
                 setattr(ncvar_pressures, 'long_name', "pressures[state] is the external pressure of thermodynamic state 'state'")
                 for state_index in range(self.nstates):
-                    ncvar_pressures[0] = self.reference_state.pressure / pressure_unit
+                    ncvar_pressures[0] = self..fully_interacting_expanded_state.pressure / pressure_unit
```
andrrizzi: And copy paste typo here.

Issue #578, opened 2016-11-25
Add support for two solvents when starting from system files


andrrizzi:
Currently it's only possible to specify a single `solvent` in the `systems` section when starting from AMBER or GROMACS files (we don't need to specify the `solvent` with OpenMM XML files since the solvent options are already in it). We should add support for something like
```yaml
systems:
  my_system:
    phase1_path: [solvent.inpcrd, solvent.prmtop]
    phase2_path: [vacuum.inpcrd, vacuum.prmtop]
    ligand_dsl: resname MOL
    solvent1: PME
    solvent2: vacuum
```
I'll open a PR for this soon.

jchodera:
Good catch! These solvent parameters currently get fed to `AmberPrmtopFile.createSystem()`, so it makes sense to do the same thing for both of these cases, since they either call that same function or `GromacsTopFile.createSystem()`.

PR #579, opened 2016-11-25
Support for XML OpenMM system

andrrizzi:
Ready for review. cc'ing @davidlmobley open-forcefield-group/smarty#184

Adds support for XML OpenMM system, which can be specified with the syntax
```yaml
systems:
  my_system:
    phase1_path: [solvent.pdb, solvent.xml]
    phase2_path: [vacuum.pdb, vacuum.xml]
    ligand_dsl: resname MOL
```

Note that I found a limitation of the current syntax if one wants to run hydration free energies starting from AMBER or GROMACS files (see #578). I'll add support for that syntax in a separate PR.


Yank/pipeline.py
```diff
@@ -291,26 +263,92 @@ def prepare_phase(positions_file_path, topology_file_path, ligand_dsl, system_op
...
+        # Read Amber prmtop and inpcrd files
+        if verbose:
+            logger.info("xml: %s" % parameters_file_path)
+            logger.info("pdb: %s" % positions_file_path)
+        with open(parameters_file_path, 'r') as f:
+            serialized_system = f.read()
+        system = openmm.XmlSerializer.deserialize(serialized_system)
+        positions_file = openmm.app.PDBFile(positions_file_path)
+        parameters_file = positions_file  # needed for topology
+    else:
```
jchodera: Is this nested `if...else` structure dangerous? What if we extract the extension at the beginning (`.xml`, `.prmtop`, or `.top`), handle each extension in a branch of `if...elif...elif..` and have the `else` case raise an exception saying that the provided extension isn't recognized?

andrrizzi: The two branches are fundamentally different because with `.prmtop` and `.top` file we need to create the system (and thus build and pass system options, set box vectors etc), while with `.xml` files we just need to deserialize it.

The exception will be raised anyway because in `OpenMM` won't be able to parse the file if it has a different format, but I'll add a better error message.

jchodera: Thanks!

andrrizzi:
Done! I'll merge when tests pass.

PR #580, opened 2016-11-25
Allow analyze to work on NetCDF files that do not contain 'metadata'

jchodera:
This minor change allows `yank analyze` to work on NetCDF files that do not contain `metadata`. The standard state correction is set to zero in this case.

This is useful for some analysis of simulation data generated by `repex` in tests of cutoff-dependence I am running right now.

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -471,9 +471,11 @@ def analyze(source_directory):
             nstates = ncfile.variables['positions'].shape[1]
             logger.info("Read %(niterations)d iterations, %(nstates)d states" % vars())
 
-            # Read phase direction and standard state correction free energy.
-            # Yank sets correction to 0 if there are no restraints
-            DeltaF_restraints = ncfile.groups['metadata'].variables['standard_state_correction'][0]
+            DeltaF_restraints = 0.0
+            if 'metadata' in ncfile.groups:
+                # Read phase direction and standard state correction free energy.
+                # Yank sets correction to 0 if there are no restraints
+                DeltaF_restraints = ncfile.groups['metadata'].variables['standard_state_correction'][0]
 
             # Choose number of samples to discard to equilibration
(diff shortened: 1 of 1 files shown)


andrrizzi:
Hmm. That error happened to me before, and with Python 3.5 too. It went away when I restarted the build, but maybe we should take a look?

jchodera:
Yeah, I worry something is wrong there:
```
Testing replica exchange facility with harmonic oscillators: Delta_u_ij
[[ 0.          0.03521011  0.22013811]
 [-0.03521011  0.          0.184928  ]
 [-0.22013811 -0.184928    0.        ]]
Delta_u_ij_analytical
[[  0.00000000e+00   1.11022302e-15   2.22044605e-16]
 [ -1.11022302e-15   0.00000000e+00  -8.88178420e-16]
 [ -2.22044605e-16   8.88178420e-16   0.00000000e+00]]
error
[[ 0.          0.03521011  0.22013811]
 [-0.03521011  0.          0.184928  ]
 [-0.22013811 -0.184928    0.        ]]
nsigma
[[ 0.          2.11798358  5.40018606]
 [-2.11798358  0.          6.88002396]
 [-5.40018606 -6.88002396  0.        ]]
ERROR
```

jchodera:
Oh, it looks like the free energy tests pass, but the expected potential energy tests fail.

I think those `Delta_u_ij_analytical` numbers actually look wrong to me. Those should be expectations of potential energies in kT, but I think there's a unit conversion error.

jchodera:
Should I debug that, @andrrizzi, or did you want to look into it?

andrrizzi:
Feel free to debug that if you already have a track!

jchodera:
I'll fix in a separate PR.

jchodera:
Actually, those delta_u differences are correct---they should be zero, but due to numerical error, they are just close to zero. So something weird might be going on with `repex` or the analysis itself.

PR #584, opened 2016-11-25
Add support for two solvents in YAML when starting from system files

### LINKED ISSUE #578: Add support for two solvents when starting from system files

andrrizzi:
Fix #578 .

This was way easier to implement than I thought. All the logic was already there, so I just had to enable the syntax. Should be ready if tests pass.

### DIFF
--- a/Yank/tests/test_yaml.py
+++ b/Yank/tests/test_yaml.py
@@ -413,6 +413,7 @@ def test_validation_correct_systems():
         lig: {{name: lig, leap: {{parameters: leaprc.gaff}}}}
     solvents:
         solv: {{nonbonded_method: NoCutoff}}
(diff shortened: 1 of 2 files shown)


jchodera:
Cool! Looks good, thanks!

PR #587, opened 2016-11-27
Fix test default lambda value after new alchemy release

andrrizzi:
Just fixing a test that now fails with changes included in alchemy 1.2.3 release. Ready to be reviewed.


jchodera:
Looks good. Thanks for catching this!

PR #588, opened 2016-11-28
Restraint Generalization Tests

Lnaden:
These tests currently all fail due to some factor I can't track down. I don't think its the ln(8*pi^2), but it might have something to do with the periodic copies all interacting. Any thoughts would be helpful as we will want to solve this before merging.

In this PR:
* Added tests for the generalized restraints; built around common test structure
* Extended YAML to allow for `anisotropic_dispersion_correction` option in `options`
* Extended valid solvents to include periodic without filling in solvent through LEaP. Warns users as this is a rare, but occasionally scenario.
* Updated YAML version to 1.1
* Some PEP8 as I saw it.


Yank/tests/test_restraints.py
```diff
@@ -71,6 +76,112 @@ def __init__(self, **kwargs):
...
+  temperature: 300*kelvin
+  pressure: null
+  anisotropic_dispersion_correction: no
+  platform: OpenCL
+
+molecules:
+  benzene-rec:
+    filepath: data/benzene-toluene-standard-state/standard_state_benzene.pdb
+  toluene-lig:
+    filepath: data/benzene-toluene-standard-state/toluene.pdb
```
andrrizzi: I think you can safely delete the molecules section.

Yank/tests/test_restraints.py
```diff
@@ -71,6 +76,112 @@ def __init__(self, **kwargs):
...
+  absolute-binding:
+    complex:
+      alchemical_path:
+        lambda_restraints:     [0.0, 0.25, 0.5, 0.75, 1.0]
+        lambda_electrostatics: [1.0, 1.00, 1.0, 1.00, 1.0]
+        lambda_sterics:        [1.0, 1.00, 1.0, 1.00, 1.0]
+    solvent:
+      alchemical_path:
+        lambda_electrostatics: [1.0, 1.0]
+        lambda_sterics:        [1.0, 1.0]
```
andrrizzi: Do `lambda_sterics/electrostatics` need to be 1.0 or 0.0 here?

jchodera:
lambda_sterics and electrostatics should all be zero, right?

Lnaden:
It theoretically should not matter. they do not change state to state so they should contribute 0 to DeltaF. We have them set to 1 while we turn on the restraints in all test cases so we dont have a decoupled ligand, then try to restrain it.

@jchodera one thing I did notice is that you actually do account for the 8*pi^2 term in the Boresh restraints. Is that a separate term or is it the correction factor we have been discussing?

andrrizzi:
> It theoretically should not matter.

Hmm. If the sterics repulsions are on, shouldn't they limit the volume explored during the simulation?

Lnaden:
Hmm, yes, that is a good point

jchodera:
I agree with @andrizzi here. Leci would be correct if there was no coupling
between the restraint and interactions, but there is. The only way to
measure the free energy of the restraint is to set all the other lambda
values to 0.

Levi, I think you forgot to check in some test PDB files. It wouldn't run
on my checkout because of missing test PDBs.

andrrizzi:
> It wouldn't run on my checkout because of missing test PDBs.

I think that's just a scheme error (it checks if the files exist during parsing). If you delete the `molecules` section, it should work.

jchodera:
Failing with:
```
ERROR: Generalized restraint simulation run to test free energy = standard state correction.
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
TypeError: general_restraint_test() takes exactly 1 argument (0 given)
```

Lnaden:
So the Harmonic restraints pass but that's it, 

* The Flat Bottom restraints fail since the ligand never experiences the restraint - border of flat bottom box never reached in PBC conditions, so simulated free energy is 0
* Boresh restraint is just wrong for reasons I haven't debugged yet, and its by more than a factor of 8pi^2

andrrizzi:
For `FlatBottom`, if you are using the API maybe you can change the `_bond_parameters` to make it tighter?

Lnaden:
We're not using the API though, thats the point

jchodera:
> The Flat Bottom restraints fail since the ligand never experiences the restraint - border of flat bottom box never reached in PBC conditions, so simulated free energy is 0

I guess we have to refactor the API in order to allow us to pass arguments to the restraints class through the YAML. I'm pretty confident that works correctly since it uses the same scheme as `Harmonic` for computing the restraining free energy.

> Boresh restraint is just wrong for reasons I haven't debugged yet, and its by more than a factor of 8pi^2

What are the actual numbers here?

jchodera:
I've checked in an improvement to the Boresch automated restraint selection algorithm where multiple attempts are made to select `r_aA0` far enough away from 0 and `theta_A0`, `theta_B0` far enough from 0 or `pi` for the standard state corrections to be accurate.

This may not be the only problem, but should at least increase the likelihood that the analytical standard state correction (which assumes the restraint coordinates can be decoupled into Gaussian integrals) will be accurate.

If we find this is not sufficiently accurate, an alternative would be to use this scheme to estimate the free energy of adding the restraints using sampling but noninteracting protein:ligand pair without solvent.

Lnaden:
Got the new syntax in for checking heavy atoms and range. I'll need to modify on of the other tests.

More problematic though, we still get unreliable calculated vs 'analytical' free energies.

jchodera:
Any chance you can compile some data? Not sure if it's easy to automate
running your test a few times to compile analytical, estimated with
uncertainty, and deviation.

jchodera:
@davidlmobley: We're validating the actual computed free energy of releasing Boresch restraints into a box of standard state volume against the approximate analytical solution for the standard state correction from Eq. 32 of [1], but we're finding that much of the time, this approximation is quite wrong. Looking at the paper, these assume that the integrals are all Gaussian and decoupled from each other. We've restricted the range of `r_aA0`, `theta_A0`, and `theta_B0` so that the Gaussian distribution doesn't leak into invalid ranges, but we're wondering if you've seen issues with this disagreement as well. Have you ever compared the analytical and actual free energies to make sure they agree?

> Boresch S, Tettinger F, Leitgeb M, Karplus M. J Phys Chem B. 107:9535, 2003.
> http://dx.doi.org/10.1021/jp0217839

davidlmobley:
@jchodera - I have not done that check, though I did usually do some checking that the "small deviation" approximation held, IIRC. I believe Huafeng commented at one point that he had observed the issue you are noting in some cases and did something else (details elude me - it's been several years) to avoid moving into regions where the approximation was invalid.

jchodera:
My guess is that he just computes the restraint energy alchemically. You could either do this *during* the simulation in alchemical intermediates added to turn off the restraint after decoupling the ligand, or you could do this in a separate calculation (like we're doing to validate the standard state correction).

Given the magnitude of the errors we're seeing, it may be necessary.

davidlmobley:
Can you be more explicit about how you would compute the restraint energy in a separate simulation alchemically? Don't you end up having to thoroughly sample the whole box, then?

jchodera:
For our test, @Lnaden sets up a periodic box of the standard state volume containing only noninteracting receptor and ligand molecules connected by the Boresch restraints. By turning the restraints off over a series of alchemical intermediates, we can compute the free energy of releasing the ligand into the standard state volume. With appropriately spaced alchemical intermediates, this can be rapidly computed, but it's a bit of a pain.

Huafeng notes that the Boresch restraint integrals can be computed exactly without making the approximations made in the paper. We'll investigate that.

davidlmobley:
Yes, it seems like computing them exactly should be possible -- even if they can't be done analytically, exact solution should be possible, and would be far preferable to any simulation-based scheme.

jchodera:
At the very least, it looks like we can easily turn this into one-dimensional numerical integrals for [`r_aA`, `theta_A`, `theta_B`] while the remaining [`phi_A`, `phi_B`, `phi_C`] have simple closed-form integrals.

jchodera:
@davidlmobley : Here is a comparison of the true Boresch restraint correction (computed using a mixture of one-dimensional numerical quadrature and analytical integration) with the analytical approximation appearing as Eq. 32 of the Boresch paper (free energies in kT):
boresch_compare
(Credit to @Lnaden for this plot and calculation.)

As you can see, it's awful, and the errors are much larger than our statistical error. 

The discrepancies come from the region one would expect:`r0` near 0, `theta0` near 0 or `pi`, and (surprisingly) `phi_0` near `-pi` or `+pi` (presumably due to OpenMM's instability in computing these near `+/-pi`). If you reduce the domain to `r0 > 1A` and narrow the `theta` and `phi` domains to be >= 20% away from the boundaries, then the agreement improves significantly:
boresch_compare_good

Our recommendations are therefore:
* restrict r, theta, and phi substantially to ensure the analytical approximation holds, or
* use the numerical exact evaluation of the correction

This may impact calculations we have done in the past and reported in the literature, or calculations you are doing right now!

davidlmobley:
@jchodera - thanks, this is interesting to know. We've always carefully avoided the endpoints you mention when selecting restraints (and we haven't actually been doing many of these calculations lately) so I don't believe there are any significant effects to worry about, but as we progress towards automated pipelines for selecting these restraints this looks critical.

jchodera:
The simple automated scheme we have right now avoids these ranges in picking atoms to restraint, but also uses the numerical standard state correction to be accurate regardless.

jchodera:
@Lnaden : Do we have a problem with the system you're testing here?
```
ERROR: Test the restraints in a protein:ligand system.
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.14.0-py2.7-linux-x86_64.egg/yank/tests/test_restraints.py", line 302, in test_protein_ligand_restraints
    yaml_builder.build_experiments()
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.14.0-py2.7-linux-x86_64.egg/yank/yamlbuild.py", line 1300, in build_experiments
    self._run_experiment(combination, output_dir)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.14.0-py2.7-linux-x86_64.egg/yank/yamlbuild.py", line 2480, in _run_experiment
    yank.create(thermodynamic_state, *phases, restraint_type=restraint_type)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.14.0-py2.7-linux-x86_64.egg/yank/yank.py", line 286, in create
    self._create_phase(thermodynamic_state, phase, restraint_type)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.14.0-py2.7-linux-x86_64.egg/yank/yank.py", line 440, in _create_phase
    atom_indices['receptor'], atom_indices['ligand'])
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.14.0-py2.7-linux-x86_64.egg/yank/restraints.py", line 125, in create_restraints
    return cls(topology, state, system, positions, receptor_atoms, ligand_atoms)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.14.0-py2.7-linux-x86_64.egg/yank/restraints.py", line 791, in __init__
    self._automatic_parameter_selection(positions, receptor_atoms, ligand_atoms)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.14.0-py2.7-linux-x86_64.egg/yank/restraints.py", line 817, in _automatic_parameter_selection
    self._restraint_atoms = self._select_restraint_atoms(positions, receptor_atoms, ligand_atoms)
  File "/home/travis/miniconda/envs/test/lib/python2.7/site-packages/yank-0.14.0-py2.7-linux-x86_64.egg/yank/restraints.py", line 915, in _select_restraint_atoms
    raise ValueError(error_string.format(min_distance, max_distance))
ValueError: There are no heavy ligand atoms within the range of [0.1,0.4] nm heavy receptor atoms!
Please Check your input files or try another restraint class
```

Yank/yamlbuild.py
```diff
@@ -36,17 +34,18 @@
...
 from .sampling import ModifiedHamiltonianExchange
 
+logger = logging.getLogger(__name__)
 
 # =============================================================================
 # CONSTANTS
 # =============================================================================
 
-HIGHEST_VERSION = '1.0'  # highest version of YAML syntax
+HIGHEST_VERSION = '1.1'  # highest version of YAML syntax
```
andrrizzi: I think we can avoid bumping this since we haven't released version `1.0` yet.

jchodera:
I can tackle the cleanup here if we know what remains to be done.

Lnaden:
* There is one test failing, but I think it might also be able to be disabled with the new tests in
* Revert YAML version to 1.0

I'll be able to do those early this evening, but if you want to take care of them sooner, go ahead.

jchodera:
Would be great if you could tackle this and take a quick look at #592 (which adds docs for what we are doing).

Lnaden:
Woo! These are all passing now. I think this is ready with approval

PR #594, opened 2016-12-04
Make clearance optional only for amber/gromacs/openmm systems

andrrizzi:
Ready for review if tests pass.

I've also updated the docs with the new `clearance` behavior, I mentioned that it is possible to use `solvent1` and `solvent2` with Amber/GROMACS/OpenMM systems, and I've added a couple of tests for the new `anisotropic_dispersion_correction` option.


jchodera:
Awesome! Looks good. Thanks!

andrrizzi:
Great! Tests pass. Merging.

PR #596, opened 2016-12-05
Cut 0.14.1: "Early Access of 1.0" Release

Lnaden:


andrrizzi:
From the YAML side in 0.14.1:
* support starting free energy calculations from OpenMM `xml` systems and `pdb` files.
* support for separate solvent configurations for the two phases when defining systems from amber/gromacs/openmm files.
* `clearance` is a mandatory parameter of an explicit solvent, only when it is associated to a system that has to go through the pipeline.

docs/whatsnew.rst
```diff
@@ -8,7 +8,10 @@ The full release history can be viewed `at the github yank releases page <https:
 
 0.14.1 Early Access of 1.0 Release
```
Lnaden: Oh, yes, that would be important. Adding

PR #598, opened 2016-12-07
Cleaning up website compile to make it more helpful

Lnaden:
- Removed the links to the old statically compiled version since `sphinx_rtd_theme` is now native to sphinx. 
- Some general page cleanup
- Changed names of `index.rst` files so the URLs we get are less ambiguous.
    - `getyank.org/index.html` -> `getyank.org/home.html`
    - `getyank.org/examples/index.html` - >`getyank.org/examples/examples.html`
    - `getyank.org/yaml/index.html` - >`getyank.org/yamlpages/yaml.html`

Hard to fully see what will happen when uploaded since I think it does do some auto-versioning magic, may need another pass after we see what happens.


Lnaden:
Fighting Sphinx makes my head hurt

PR #604, opened 2016-12-09
New states classes

andrrizzi:
This is the first big chunk of stuff for the new API. I've added a new module `states.py` implementing `ThermodynamicState`, `SamplerState` and `CompoundThermodynamicState`. The last one allows to extend `ThermodynamicState` by composition rather than inheritance and it will be used to mix `ThermodynamicState` with `AlchemicalState`. I've added a lot of documentation and tests so hopefully it should be clear and it shouldn't introduce new bugs.

Nothing is attached to the main code yet. Before doing it, I need to adapt `alchemy.AlchemicalState`, and if we want to directly move to a scheme that uses `MCMoves`, I'll also add a `ContextCache`. Next thing is to write the first implementation of the lower level `Storage` class.

### DIFF
--- a/Yank/cache.py
+++ b/Yank/cache.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python
+
+# =============================================================================
+# MODULE DOCSTRING
+# =============================================================================
+
+"""
+Provide cache classes to handle creation of OpenMM Context objects.
+
+"""
+
+
+# =============================================================================
+# GLOBAL IMPORTS
+# =============================================================================
+
+import collections
+
+
+# =============================================================================
+# GENERAL LRU CACHE
+# =============================================================================
+
+class LRUCache(object):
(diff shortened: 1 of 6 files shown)


jchodera:
Awesome!

Can we chat about this briefly late today?

andrrizzi:
Sure! I forgot to add that I still need to make few more changes. I'd like to add automatic treatment of the thermostat, if I find a way to squeeze it in.

andrrizzi:
Done! Ready for review. Nothing is attached to the code yet. It just adds a new `states.py` module with tests. The stuff left in QuantifiedCode are quite useless, and one of them suggest me to introduce a bug!

In a separate PR I'll create a new `cache.py` for the global context cache.

andrrizzi:
Closing this since this has been moved to `openmmtools`.

PR #607, opened 2016-12-15
Laid out API for Storage Interface and its internal class.

Lnaden:
Here is the API have laid out for the super generic storage interface. Nothing is coded yet but I have psuedo-code at the bottom of the `storage.py` file.

Please look over what I have documented and let me know your thoughts. Feel free to ask me any questions as you might have them.

Other pep8, spelling, unused import. No other code changes. None of this is hooked into YANK proper yet and eventually we may want to split this storage module into its own package for internal group use.

### DIFF
--- a/Yank/storage.py
+++ b/Yank/storage.py
@@ -1,0 +1,451 @@
+#!/usr/bin/env/python
+
+# =============================================================================
+# MODULE DOCSTRING
+# =============================================================================
+
+"""
+Classes that store arbitrary NetCDF (or other) options and describe how to handle them
+
+"""
+
+# =============================================================================
+# GLOBAL IMPORTS
+# =============================================================================
+
+import abc
+import copy
+import netCDF4 as nc
+
+import numpy as np
+from simtk import openmm, unit
+
+from . import utils
+
+ABC = abc.ABCMeta('ABC', (object,), {})  # compatible with Python 2 *and* 3
+
+# =============================================================================
+# MODULE FUNCTIONS
+# =============================================================================
+
+# =============================================================================
+# CUSTOM EXCEPTIONS
+# =============================================================================
+
+# =============================================================================
+# GENERIC SAVABLE DATA
+# =============================================================================
+
+
+class StorageInterfaceDirVar(object):
+    """
+    Storage Interface Directory/Variable (SIDV) class is a versatile, dynamic class which gives structure to the
+    variables stored on disk by representing them as methods of the StorageInterface and other instances of itself.
+    New variables and folders are created by simply trying to use them as properties of the class itself.
+    The data stored and read is not kept in memory, the SIDV passes the data onto the writer or tells the fetcher to
+    read from disk on demand.
+
+    The API only shows the protected and internal methods of the SIDV class which cannot be used for variable and
+    directory names.
+
+    This class is never meant to be used as its own object and should never be invoked by itself. The class is currently
+    housed outside of the main StorageInterface class to show the API to users.
+
+    TODO: Move this class as an internal class of StorageInterface (NOT a subclass)
+
+    == Examples ==
+
+    Create a basic storage system and write new DATA to disk
+    >>> my_storage = StorageInterface('my_store.nc')
+    >>> my_storage.my_variable.write(DATA)
+
+    Create a folder called "vardir" with two variables in it, "var1" and "var2" holding DATA1 and DATA2 respectively
+    >>> my_storage = StorageInterface('my_store.nc')
+    >>> my_storage.vardir.var1.write(DATA1)
+    >>> my_storage.vardir.var2.write(DATA2)
+
+    Fetch data from the same folders and variables as above
+    >>> my_storage = StorageInterface('my_store.nc')
+    >>> var1 = my_storage.vardir.var1.fetch()
+    >>> var2 = my_storage.vardir.var2.fetch()
+
+    Run some_function() to generate data over an iteration, store each value in a dynamically sized var called "looper"
+    >>> my_storage = StorageInterface('my_store.nc')
+    >>> for i in range(10):
+    >>>     mydata = some_function()
+    >>>     my_storage.looper.append(mydata)
+
+    """
+
+    def __init__(self, file_name, storage_system='netcdf'):
+        """
+        Parameters
+        ----------
+        file_name : string
+            Specify what file name to read and write to on the disk.
+        storage_system : string
+            Specify what type of storage system module to load for interfacing with the file_name
+        """
+
+    """
+    USER FUNCTIONS
+    Methods the user calls on SIDV variables to actually store and read data from disk. These cannot be names of child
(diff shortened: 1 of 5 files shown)


Yank/storage.py
```diff
@@ -0,0 +1,451 @@
...
+# =============================================================================
+# CUSTOM EXCEPTIONS
+# =============================================================================
+
+# =============================================================================
+# GENERIC SAVABLE DATA
+# =============================================================================
+
+
+class StorageInterfaceDirVar(object):
```
andrrizzi: I don't know how the internals will be, but is there a reason why we can't have two separate `StorageInterfaceDir` and `StorageInterfaceVar(StorageInterfaceDir)` classes? This way you could avoid dynamically locking methods.

Yank/storage.py
```diff
@@ -0,0 +1,451 @@
...
+
+        Parameters
+        ----------
+        data
+            This is what will be written to disk, the data will be processed by the STORAGESYSTEM as to how to actually
+            handle the data once the units have been stripped out and assigned to this instance of SIDV.
+
+        """
+
+    def fetch(self):
```
andrrizzi: Sorry if I missed it. Does it provide a way to read a particular iteration or the whole array?

Yank/storage.py
```diff
@@ -0,0 +1,451 @@
...
+# =============================================================================
+
+# =============================================================================
+# GENERIC SAVABLE DATA
+# =============================================================================
+
+
+class StorageInterfaceDirVar(object):
+    """
+    Storage Interface Directory/Variable (SIDV) class is a versatile, dynamic class which gives structure to the
```
andrrizzi: PEP8 suggest 72 characters line width for docstrings, which is nuts, but limiting it to 80-90 characters makes them more readable when viewing diffs.

Lnaden:
A few things:

> implementing only `__get__/__set__/__getitem__` 

I originally wanted to do that, but it makes no sense and is a bit reckless. Python does not have a `__get__` and `__set__` methods which to overload the equivalent of `x.var` and `x.var = ...` since then you could never retrieve the object stored at `x.var` nor could you ever change the object set at `x.var`. I can get around the `__get__` limitation by making the `parent class.property` refer to a hidden function of `fetch` down at the lower class, but that would only implement a half solution and be inconsistent with the `set` actions. So I decided to make it consistent at the cost of true magic.

> The utils module now comes with a with_metaclass function shipped with the new-states PR.

I'll look into that

> PEP8 suggest 72 characters line width

They are 120 characters right now (with indentation). Are you saying make them shorter?

> Sorry if I missed it. Does it provide a way to read a particular iteration or the whole array?

That is one of the things I have not figured out. But there is nothing stopping you from doing `.fetch()[some:slice]` if that is good enough. I can probably also load in a `.fetchlast()` argument. What I would like to not do is `fetch([some:slice])` since that makes the function's returns somewhat ambiguous and assumes you already know the length which this system does not care about since nothing is kept in memory longer than it has to be. `fetch` is probably only really used in analysis and resume which is why a `fetchlast` may be worth putting in. I would also be okay with `fetchslice`, but that is kind of hard to pass in `[x:y]` without casting it to a slice object since Python does not know that syntax should be slice natively.

> but maybe the property here can be just _variable instead of all capital letters

Thats fine too. I was just thinking from a standard API display that the `_method` and `__method`s are usually hidden from the user facing API docs, so by making them like this and pointing them to the internal `_property` objects, it shows the user what cannot be used.

> Any reason why we can't have the type of storage system as separate 'NetCDFStorage` class?

We absolutely will! I just put it in as a string arg and figure load it on the back end, but passing it as a full class would also work. This is only the user facing API I have written out here. There is an entire layer between the `StorageInterface` and the Disk I haven't written which actually handles all the casting of Python Types <=> Disk Storage Type which is obscured from the user. I kind of started that in the `storage_handlers` module (will change once the SI API is in) and that is where we in the lab can change and extend the behavior of certain object types. I will still need to work that out but once we know how the `StorageInterface` will behave I can flesh that out some more.

Lnaden:
> why we can't have two separate StorageInterfaceDir and StorageInterfaceVar(StorageInterfaceDir)

Because when the object is instantiated it is not bound to any one type, so we would not know which to call, and when we trap the dynamic api through the `__getattr__` command all that is passed in is the string of the first method's call, we know nothing of the following methods.

e.g. `x.soup.chicken` the `__getattr__` of `x` knows only of `soup` and nothing of `chicken` so we don't know what will come next if anything. Similarly, we don't know if we will then have a `x.soup.chicken.noodle` later so we cant just assume that `soup` -> directory and `chicken` -> variable. The way I plan it now each class will cascade down instantiating itself and any children as variables, then cascading back up and converting to directories as needed before returning control to the user. This will only happen once per `SIDV` instance so it should not be too much overhead on initial setup.

andrrizzi:
> Python does not have a `__get__` and `__set__` methods

Sure it does! https://docs.python.org/3/howto/descriptor.html . Unless I misunderstood.
> and assumes you already know the length

That's true but every list type in Python assumes that the user knows it, and the expected behavior in Python is that `my_list[:200000]` returns as many objects as it can.

andrrizzi:
> Because when the object is instantiated it is not bound to any one type

Ah! I see, right!

Lnaden:
> Sure it does!

Hmm... well... I missed that entirely apparently. Those could replace some of the functionality. I would need to test with those since I have never really seen them used in code. I'm not sure how we would do the append then, unless of course we literally do just a `.append()` method and let the `__set__` method be the hard write, although I'm not sure how to handle overwrite protection in that case. 

> ... the expected behavior in Python is that ...

Fair point, but that is why I suggested the `.fetch()[some:slice]` but the trade of is it does have to load the whole thing to memory, although its usually in memory in most cases that it is sliced.

andrrizzi:
> Those could replace some of the functionality.

To be clear I was not suggesting to replace the `write` and `fetch` functions with descriptors, only to implement descriptors functions that behave as `write` and `fetch` with default arguments (they can make use of them internally). But it is just a stylistic thing.
> unless of course we literally do just a `.append()` method

Yeah, that's probably unavoidable, and probably also the clearest way to make the user get what's happening.
> but the trade of is it does have to load the whole thing to memory

I see, so it would be to make clear to the user that it loads the whole thing in the memory? Btw in the future adding a time-to-live cache could be as simple as adding a decorator to the fetch method https://docs.python.org/3/library/functools.html#functools.lru_cache (pretty cool).

andrrizzi:
> `x.soup.chicken` the `__getattr__` of `x` knows only of `soup` and nothing of `chicken` so we don't know what will come next if anything.

Sorry, I'm confused, did you mean `__setattr__`? Or you plan to assign type only when you get stuff?

andrrizzi:
Ah, ok, I just got it!

Lnaden:
Closing this for now as I push multiple backups to github

PR #609, opened 2016-12-17
Solvent DSL keyword.

### LINKED ISSUE #606: Need way to designate what residue(s) are solvent aside from specified set
@andrrizzi [points out](https://github.com/open-forcefield-group/smarty/issues/184#issuecomment-266244128) that currently, solvents are recognized from [this list](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/residue_names.py#L23-L42) and everything else is treated as receptor. Currently this means that I can't look at solvation of, say, a small molecule in cyclohexane unless I give cyclohexane one of the recognized residue names (because by default, cyclohexane is recognized as "receptor"). However, this won't work as a general solution, because (a) it's cumbersome to have to artificially change the residue name for each solvent, but more importantly (b) I might be interested in simulating a small molecule in various combinations of solvents or mixtures (binary, ternary, or higher order) where this could become a real problem (e.g. I might have to use different residue names for different solvents depending on what they're mixed with to avoid name clashes, since no two residues can have the same name). 

I can bring this up on mdtraj as well.

andrrizzi:
Cc'ing @davidlmobley .

Fix #606 . Ready for review.

### DIFF
--- a/Yank/pipeline.py
+++ b/Yank/pipeline.py
@@ -84,7 +84,7 @@ _SOLVENT_RESNAMES = frozenset(['118', '119', '1AL', '1CU', '2FK', '2HP', '2OF', 
         'YH', 'YT3', 'ZN', 'ZN2', 'ZN3', 'ZNA', 'ZNO', 'ZO3'])
 
 
-def find_components(system, topology, ligand_dsl, solvent_resnames=_SOLVENT_RESNAMES):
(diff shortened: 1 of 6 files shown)


davidlmobley:
Thanks. Hopefully I'll try this this week, @andrrizzi !

PR #617, opened 2016-12-27
New storage system with auto-magical variable handling

Lnaden:
This is the new storage system module which handles hard drive IO for arbitrary data sets with minimal user input. Currently handles NetCDF storage, but can be extended to add new storage modules with from the abstract class.

Handles the following types:
* String
* Int
* Float
* List (any size)
* Tuple (any size)
* Quantity (of any of the previous types)
* Dict (entries can be any of the previous types with string keywords)
* numpy.ndarray (and Quantity(ndarray))

The `StorageInterface` module allows auto-magical data IO such that you define your directories and variables on-the-fly with commands such as 

`StorgeInterfaceInstance.my_directory.my_variable.write(data)` 

where a folder called `my_directory` and `my_variable` are created for you with `data` being automatically encoded to the storage medium and saved. Accessing it is done with 

`StorgeInterfaceInstance.my_directory.my_variable.read()`

and the data at that same location is fetched and automatically decoded for you. No template nor pre-determining structure required.

----

There are a few things I would like reviewed before I merge this:
- Does the public API make sense? (functions not proceeded by `_`)
- Would people outside of YANK find this helpful?
  - Tagging @bas-rustenburg @gregoryross @pgrinaway 
  - Very light refactoring to split off into own module
- Should I add more examples?
- Any other comments/suggestions. 

Has not been integrated into current code. @andrrizzi I could not get the `__get__` and `__set__` attributes working, although there was not much time I spent on it.

Realization of #607

### DIFF
--- a/Yank/storage.py
+++ b/Yank/storage.py
@@ -21,6 +21,7 @@ import numpy as np
 from simtk import openmm, unit
 
 from . import utils
+from .storage_handlers import *
 
 ABC = abc.ABCMeta('ABC', (object,), {})  # compatible with Python 2 *and* 3
 
--- a/Yank/storage.py
+++ b/Yank/storage.py
@@ -13,25 +13,13 @@ Classes that store arbitrary NetCDF (or other) options and describe how to handl
 # GLOBAL IMPORTS
 # =============================================================================
 
+import os
 import abc
-import copy
-import netCDF4 as nc
+import warnings
 
-import numpy as np
-from simtk import openmm, unit
-
-from . import utils
-from .storage_handlers import *
+from .storage_handlers import NetCDFIODriver
 
 ABC = abc.ABCMeta('ABC', (object,), {})  # compatible with Python 2 *and* 3
-
-# =============================================================================
-# MODULE FUNCTIONS
-# =============================================================================
-
-# =============================================================================
-# CUSTOM EXCEPTIONS
-# =============================================================================
 
 # =============================================================================
 # GENERIC SAVABLE DATA
@@ -78,15 +66,30 @@ class StorageInterfaceDirVar(object):
 
     """
 
-    def __init__(self, file_name, storage_system='netcdf'):
+    def __init__(self, name, storage_driver, storage_interface, predecessor=None):
         """
         Parameters
         ----------
-        file_name : string
-            Specify what file name to read and write to on the disk.
-        storage_system : string
-            Specify what type of storage system module to load for interfacing with the file_name
+        name : string
+            Specify the name of the storage variable on the disk. Full path is determined from the predecessor chain
+        storage_driver : instanced storage driver of top level StorageInterface object
+            This is the driver where commands are issued to fetch/create the variables as needed
+        storage_interface : StorageInterface instance
+            Acting interface which is handling top level IO operations on the file itself
+        predecessor : StorageInterfaceDirVar
+            Directory-like SIDV above this instance
+
         """
+        self._name = name
+        self._storage_driver = storage_driver
+        self._storage_interface = storage_interface
+        self._predecessor = predecessor
+        # Define the object this instance is attached to
+        self._variable = None
+        # initially undetermined type
+        self._directory = None
+        self._variable = None
+        self._metadata_buffer = {}
 
     """
     USER FUNCTIONS
@@ -116,6 +119,25 @@ class StorageInterfaceDirVar(object):
             If False, overwriting is allowed but the variable will still not be able to be appended to.
 
         """
+        self._bind_write_append()
+        writeable = False
+        if not self._variable:
+            path = self.path
+            # Try to get already on disk variable
+            try:
+                self._variable = self._storage_driver.get_variable_handler(path)
+            except KeyError:  # Trap the "not present" case, AttributeErrors are larger problems
+                self._variable = self._storage_driver.create_storage_variable(path, type(data))
+                writeable  = True
+            self._directory = False
+            self._dump_metadata_buffer()
+            if self.predecessor:
+                self.predecessor._write_append_directory()
+        if not writeable and protected_write:
+            # Lock ability to protect the variable
+            raise IOError("Cannot write to protected object on disk! Set 'protected_write = False` to overwrite "
+                          "this protection.")
+        self._variable.write(data)
 
     def append(self, data):
         """
@@ -132,8 +154,21 @@ class StorageInterfaceDirVar(object):
             handle the data once the units have been stripped out and assigned to this instance of SIDV.
 
         """
+        self._bind_write_append()
+        if not self._variable:
+            path = self.path
+            # Try to get already on disk variable
+            try:
+                self._variable = self._storage_driver.get_variable_handler(path)
+            except KeyError:  # Trap the "not present" case, AttributeErrors are larger problems
+                self._variable = self._storage_driver.create_storage_variable(path, type(data))
+            self._directory = False
+            self._dump_metadata_buffer()
+            if self.predecessor:
+                self.predecessor._write_append_directory()
+        self._variable.append(data)
 
-    def fetch(self):
+    def read(self):
         """
         Read the variable and its data from disk.
 
@@ -146,6 +181,18 @@ class StorageInterfaceDirVar(object):
             Data stored on VARIABLE read from disk and processed through the STORAGESYSTEM back into a Python type
             and possibly through the UNIT logic to recast into Quantity before being handed to the user.
         """
+        self._bind_read()
+        if not self._variable:
+            path = self.path
+            # Try to get variable on disk
+            try:
+                self._variable = self._storage_driver.get_variable_handler(path)
+            except KeyError as e:
+                raise e
+            self._directory = False
+            if self.predecessor:
+                self.predecessor._read_directory()
+        return self._variable.read()
 
     def add_metadata(self, name, data):
         """
@@ -166,16 +213,20 @@ class StorageInterfaceDirVar(object):
         data: What data you wish to attach to the `name`d metadata pointer.
 
         """
+        if self._variable:
+            self._variable.add_metadata(name, data)
+        elif self._directory:
+            self._variable.add_metadata(name, data)
+        else:
+            self._metadata_buffer[name] = data
 
     """
     PROTECTED FLAGS
     Used for functionality checks that will likely never be called by the user, but exposed to show
     what names can NOT be used in the dynamic directory/variable namespace a user will generate.
-
-    All capital names to distinguish them as protected types and avoid clashes with built-in functions.
     """
     @property
-    def VARIABLE(self):
+    def variable(self):
         """
         Checks if the object can be used in the .write, .append, .fetch functions can be used. Once this is set, this
         instance cannot be converted to a directory type.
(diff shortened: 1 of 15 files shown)


Yank/storage.py
```diff
@@ -0,0 +1,506 @@
...
+            # Try to get already on disk variable
+            try:
+                self._variable = self._storage_driver.get_variable_handler(path)
+            except KeyError:  # Trap the "not present" case, AttributeErrors are larger problems
+                self._variable = self._storage_driver.create_storage_variable(path, type(data))
+                writeable  = True
+            self._directory = False
+            self._dump_metadata_buffer()
+            if self.predecessor:
+                self.predecessor._write_append_directory()
```
andrrizzi: Doesn't `_bind_write_append()` already fix the predecessor? Same in `append()` and `read()`.

Yank/storage.py
```diff
@@ -0,0 +1,506 @@
...
+
+        """
+        path = []
+        # Cascade the path
+        if self.predecessor is not None:
+            path = self.predecessor.path.split('/')  # Break path-like string to list to process
+        # Add self to the end
+        path.extend([self.name])  # Wrap in list or it iterates over the name chars
+        # Reduce to a path-like string
+        return '/'.join(path)
```
andrrizzi: Hmm. Why splitting the predecessor path and then joining it back?

Yank/storage.py
```diff
@@ -0,0 +1,506 @@
...
+        name : string
+
+        """
+        return self._name
+
+    @property
+    def bound_target(self):
+        """
+        Fetch the handler for the bound target, either the directory or the variable so that the user can directly
+        manipulate the handler has they need.
```
andrrizzi: typo "has".

Lnaden:
> Docs: there are many classes here and it would be good to add a general description of the various components in the module...
> File structure: Maybe it's cleaner to have a `storage/`...

I have converted to the storage module to a submodule and have added full paragraph descriptions of each class and their interplay to the `__init__.py` file.

> NCTypeHandlers Many of the functions here are __almost__ identical

Emphasis mine, They are "almost" identical, but they all have slightly different syntax for some of the netcdf properties, the chunk sizing, the possible multi-dimensional nature of the data, etc. And then compound that with the fact that some types are actually compound types, which is why the codec functions are all split off into their own mini-functions, which are also very similar. Because of the subtle irregular differences, I could not think of a way to encapsulate all of this into a single parent function/class.

> Should these examples go in the StorageInterface docs?

Fixed

> I think it would be very useful to have a change_directory

Im not sure what the advantage is here, also, there is no way to be "in" the directory so to speak since the storage medium may not even have directories.

> Doesn't this always raise an IOError if self._variable already exist (and `writable` is False)? Also I think the protected_write check should be done before creating the variable. If the user catch and handle the exception the _storage_driver would still have created the variable.

So `protected_write` only prevents overwriting on the disk if data is already there. We still want you to be able to `write` on new data, so the only time the IOError is raised is if the data is already on disk, and `protected_write` is true, which is what we want. New data you write will write regardless, and exiting data will overwrite if `protect_write` is `False`. If the data is already on disk and bound, its stored in the `._variable` property so we already know its on disk, therefore we only have to worry about the creation. I changed some of the variables around to avoid negation and remove logic confusion: if the variable exists and is protected, do not overwrite.

> I think os.path.split() does this.

`os.path.split` only breaks it into a `(tail, head)` and not into a full separated list. So I had to split it myself.

> There's a portable file separator that can be used instead of '/'. I think it's os.path

Again, because os.path.split doesn't really do what we need, its a bit 

> Hmm. Why splitting the predecessor path and then joining it back?

Because we don't know when called if there are predecessors or if we are just a singular object, if it is a single object, then there are no "/" and the path is a single string. Splitting the path provides a regulated way to remove leading and trailing `/`. 

I'll work on refactoring the NCTypeHandlers

andrrizzi:
Got it, thanks!

> Im not sure what the advantage is here, also, there is no way to be "in" the directory so to speak since the storage medium may not even have directories.

If different classes/modules want to store stuff they'll all have to create a unique directory so we risk to have a lot of lines like
```python
storage.module_dir.directory1.variable = 3
storage.module_dir.directory2.variable = 3
```
I think being able to cd will make it shorter and more readable. If the particular storage implementation doesn't support directories, then changing directory can simply fail, but all our repos will need directories.
> `os.path.split` only breaks it into a (tail, head) and not into a full separated list

Here I wasn't referring only to `decompose_path()` but to the whole code block. Is that somewhat equivalent to
```python
head_path, target_name = os.path.split(path)
```
?

Lnaden:
> I think being able to cd will make it shorter and more readable.

There should be nothing stopping you from doing something like:
```python
dir1 = storage.module_dir.directory1
dir2 = storage.module_dir.directory2
dir1.variable = 3
dir2.variable = 3
dir1.variable2 = 4
...
```

> Here I wasn't referring only to decompose_path()...

I suppose in this case you are correct, but I am trying to be consistent with the internal path representation while avoiding all leading `/` and reducing external dependency as much as I can. Since that would be the only place we use it, I just decided to write my own line of code to do it instead of requiring an extra import.

andrrizzi:
> There should be nothing stopping you from doing something like:

Cool!

Lnaden:
Because the `StorageInterfaceDirVar` (SIDV) objects are created at the `__getattribute__` method, the act of trying to reference the object calls it into existence. You could create the SIDVs for the variables in advance as well such as `energies = storage.energies`, however they will not be allocated to disk until you first try to to `write/read/append`, so that first call may be a bit slow, but then its in memory and the SIDV is bound to a location on disk. This helps reduce the amount of IO to disk and avoids creating a bunch of stuff that isn't used.

Lnaden:
Found an interesting problem in the way I have the `append` system implemented. Right now the append assumes the `iterable` dimension will be shared across all variables. Without passing in additional information to the append argument, we cannot have multiple `iterable` dimensions. However, I am worried about also about assuming every append data should have its own dimension, that will just kind of bloat the file.

How do we want to handle the iterable dimension for data which gets appended to?

I have to fix a bug with this anyways, so I want to fix it right the first time.

Lnaden:
There are a few solutions I see:
1. Have every iterable be its own dimension
    - Pros: every iterable is independent of the other, can be coded independent of other possible platforms besides NetCDF, `StorageInterface` layer does not need to worry about it
    - Cons: Larger memory footprint, no way to mix variables
2. Have all iterables be on the SAME dimension
    - Pros: less memory footprint, can be coded independent of other possible platforms besides NetCDF, `StorageInterface` layer does not need to worry about it
    - Cons: All iterables will have the same number of 'iterations', no way to mix variables
3. Be able to specify what the dimension they should be under 
    - Pros: Every iterable has fine grained control on which dimension they are a part of
    - Cons: Medium memory footprint, storage platform specific coding that will be hard to generalize, requires special call to set the dimension adding an initilization that we have been trying to avoid with the storage module

andrrizzi:
I'm not sure I understand. Isn't the point of having an `append` to not have a fixed dimension?

Lnaden:
Yes, the actual dimensions themselves are infinite in length, but its a matter of pointing to which infinite length dimension. Right now they are all tied to the same dimension and the index at which data is being appended to is the same.

The bug is that saying
```python
var1.append(data)
var2.append(data2)
```
actually advances the index of the iterable dimension by 2, and then `len(iterable_dim) = 2`, even if I fix the bug such that each variable is counting over its own index, then its up to the user to figure out what the last max index was along that dimension, since the non-written to indices are left with filler data.

andrrizzi:
I see. I personally I vote for everything has its own dimension. We shouldn't have too many variables so the memory footprint should not be that bad, and it's consistent with the dynamic philosophy of the `StorageInterface` class.

Lnaden:
@andrrizzi could you take a look at [this bit of code](https://github.com/choderalab/yank/pull/617/files#diff-9710953b4697a71f5014c76ef226f053R882) for the unified scalar handler. Is there a better way you can think of to do this? 

I could not think of a way to partially initialize a class besides this, especially considering I need to act on the staticmethods, of which each type handler has and abstract static method (not sure how else to handle that).

andrrizzi:
I see. I think it would be simpler to let the subclasses `NCFloat(NCScalar)` implement their own `_dtype` and `_dtype_string` properties (i.e. `NCScalar` would be abstract).

Lnaden:
This should be ready for looking over again

andrrizzi:
I'm looking at this now.

Yank/storage/iodrivers.py
```diff
@@ -0,0 +1,1753 @@
...
+            self._attempt_storage_read()
+            # Handle variable size objects
+            # This line will not happen unless target is real, so None should not be returned by _check_write_append
+            self._check_write_append()
+            if self._output_mode is 'a':
+                self._save_shape = self._bound_target.shape[1:]
+            else:
+                self._save_shape = self._bound_target.shape
+        except KeyError as e:
+            raise e
```
andrrizzi: Isn't this equivalent to not have a `try..except` clause? Same in `NCScalar` and `NCIterable`.

Yank/storage/storageinterface.py
```diff
@@ -0,0 +1,514 @@
...
+    >>> var2 = my_storage.vardir.var2.fetch()
+
+    Run some_function() to generate data over an iteration, store each value in a dynamically sized var called "looper"
+    >>> mydata = [-1, 24, 5]
+    >>> my_storage = StorageInterface('my_store.nc')
+    >>> for i in range(10):
+    >>>     mydata = some_function()
+    >>>     my_storage.looper.append(mydata)
+    """
+    def __init__(self, file_name, storage_system=NetCDFIODriver):
```
andrrizzi: We just talked about this, but I'm saving this for the records. We should pass an object instead of a class name to make the user be able to pass options in the driver's constructor (including his own codecs).

Yank/storage/iodrivers.py
```diff
@@ -0,0 +1,1753 @@
...
+    def check_scalar_dimension(ncfile):
+        """
+        Check that the `scalar` dimension exists on file and create it if not
+
+        """
+        if 'scalar' not in ncfile.dimensions:
+            ncfile.createDimension('scalar', 1)  # scalar dimension
+
+    @staticmethod
+    def check_infinite_dimension(ncfile, name='iteration'):
```
Lnaden: I want to leave this in since people may actually use it as part of the standalone aspect.

Yank/storage/iodrivers.py
```diff
@@ -0,0 +1,1753 @@
...
+        ncfile : NetCDF File
+        name : string, optional, Default: 'iteration'
+            Name of the dimension
+
+        """
+        if name not in ncfile.dimensions:
+            ncfile.createDimension(name, 0)
+
+    @staticmethod
+    def check_iterable_dimension(ncfile, length=0):
```
Lnaden: There is a check for that in already, it has to be an int >= 0

andrrizzi: Sorry, I am saying _strictly_ greater than 0.

Lnaden:
I think there is something going on with `conda-build` right now. I'll check later to see if they fix it.

jchodera:
cc: https://github.com/omnia-md/conda-recipes/issues/669

Lnaden:
We have had some discussion about splitting this off as its own package. If we do that, we will want to add it to the `omnia` conda channel. What should we call this package and are there additional features we want to include such as other `StorageIODriver`s?

andrrizzi:
If this is something that we plan to always use with a some other package (e.g. `openmmmcmc` or `repex`) we could think about putting it there.

Lnaden:
I think it was more general usage by the lab itself, not really pinned to 1 particular package

andrrizzi:
What I meant is that if we think that the lab will always use `storage` and some other package such as some common sampling package, we could put them in the same repo.

Lnaden:
Im not sure that is so helpful. If we expect `storage` to be used in a large number of projects, can we ensure that all the projects will use whatever we tie `storage` to? I don't want to have people installing something like `openmmmcmc` just to access `storage`, which itself does not depend on the parent package.

andrrizzi:
> can we ensure that all the projects will use whatever we tie `storage` to?

I don't know (this is why I used the if-clause in proposing it). It depends on what's our strategy for common software infrastructure. I'm just trying to think about ways to avoid too much fragmentation, as each stand-alone repo adds maintenance overhead.

Lnaden:
Closing this as the code and all of its requirements have been moved to the OpenMMTools repo

Issue #611, opened 2016-12-22
Problems with Yank docs relating to parallel jobs


davidlmobley:
The link to running in MPI mode on this page is broken: http://getyank.org/latest/quickstart.html

davidlmobley:
Also, the link to `build-mpirun-configfile` from this page is broken: http://getyank.org/latest/running.html#id3

I'm not conda installing since I need the latest `solvent_dsl` stuff Andrea added; though now that I think of it I'll try `yank-dev` as presumably that has the latest.

davidlmobley:
Nope, I just tried conda installing yank-dev and this seems to be incorrect:

> build-mpirun-configfile is automatically installed with YANK when you use the conda installation route.

I'll go hunting for it.

( @Lnaden - I know you're very interested in issues with the docs; there seem to be a couple here. I'm trying to sort out how to get Yank to recognize multiple GPUs per node on my cluster under the SLURM queueing system; I've got Exxact nodes with 8x Nividia Titan II's and so far can only get it to see one at a time; I'm trying to sort out the MPI issues necessary.)

davidlmobley:
Well, I found the file, which has been moved to https://github.com/choderalab/clusterutils/blob/master/clusterutils/build_mpirun_configfile.py rather than the https://github.com/choderalab/clusterutils/blob/master/scripts/build-mpirun-configfile.py mentioned in the docs. However, superficially, it doesn't look useful to me as it seems to be PBS-specific and I'm on SLURM (on one cluster) and Torque (on the other) and this appears to raise an exception if it's not PBS... 

Any advice?

davidlmobley:
Also, I would note that the [http://getyank.org/latest/running.html#running-in-mpi-mode](documentation) lists the command for running Yank in MPI mode as identical to the command for running it in serial mode, which is probably not the case.

jchodera:
You want to conda install yank, not yank-dev. That should automatically install the cluster utilities.

In fact, you probably want to install yank-examples to bring in the examples, yank, and all dependencies. This contains example batch scripts for Torque.

We don't have a slurm cluster to test on, but if you can give us access to one, we can add slurm batch queue script examples as well.

davidlmobley:
I can get you access to our slurm cluster; I'll have to create a collaborator account at UCI for you guys. Who would be the best person, you, @Lnaden , or @andrrizzi , @jchodera ?

davidlmobley:
[...]

PR #618, opened 2016-12-27
Fixed broken links in the docs and some typos

Lnaden:
Partial resolution to #611. Rest of it requires additional information.


docs/running.rst
```diff
@@ -126,19 +128,25 @@ Alternatively, to run the simulation in MPI mode:
 
 .. code-block:: none
 
-   $ yank script --yaml=yank.yaml
+   $ mpirun -n # yank script --yaml=yank.yaml
```
jchodera: Let's not use `#` here, and this may not even be necessary. If you're using the conda-installed `mpirun` (hydra), it will autodetect the queue manager and number of processes, won't it?

jchodera: From [the docs](https://wiki.mpich.org/mpich/index.php/Using_the_Hydra_Process_Manager#Resource_Managers_and_Launchers):
> In most cases, you don't need to do anything special for Hydra to work with your slurm, pbs, loadleveler, lsf, sge or cobalt environment. It'll automatically detect these resource managers and interact with them. The following command will be sufficient to launch processes on all the allocated cores:

docs/running.rst
```diff
@@ -126,19 +128,25 @@ Alternatively, to run the simulation in MPI mode:
...
+
+where ``#`` is the number of parallel instances of YANK you want to run with. Keep in mind your system may have a
+different MPI executable name and/or syntax.
 
 On systems with multiple NVIDIA GPUs per node, it is necessary to perform masking using ``CUDA_VISIBLE_DEVICES``.
 
 On systems using the conda-installed ``mpi4py`` package, the `MPICH2 hydra mpirun <https://wiki.mpich.org/mpich/index.php/Using_the_Hydra_Process_Manager>`_ will be automatically installed for you.
-You can use the cluster utility script `build-mpirun-configfile.py <https://github.com/choderalab/clusterutils/blob/master/scripts/build-mpirun-configfile.py>`_ available in our `clusterutils <https://github.com/choderalab/clusterutils>`_ tools to generate an appropriate ``configfile``:
+You can use the cluster utility script
+`build-mpirun-configfile.py <https://github.com/choderalab/clusterutils/blob/master/clusterutils/build_mpirun_configfile.py>`_
```
jchodera: We don't use the Python executable directly---we use the conda-installable `clusterutils` package that exposes the command-line executable `build-mpirun-configfile`. I don't think we want to confuse people by specifying both the Python script and command-line form.

docs/running.rst
Lnaden: That was my thought as well. but based on @davidlmobley 's comments regarding his SLRUM server it did not appear to be working. So I added this in response to those comments, but it may be a mater of more debugging first.

jchodera: @davidlmobley: Can you get us access to test/debug? This should "just work".

jchodera:
Can we also add documentation for the option to extend the number of iterations?

Lnaden:
[Already added it](http://getyank.org/latest/yamlpages/options.html#yaml-options-number-of-extension-iterations) in the PR that added that option

jchodera:
Ah! We may want to add that to the [Running Yank](http://getyank.org/latest/running.html) page too, since it is probably one of the most common overrides someone will want to specify from the command line.

I hadn't noticed it before, but 
* `number_of_extension_iterations` is kind of verbose, and 
* it doesn't allow one to easily intuit what happens when you also specify `number_of_iterations`

Is there some way we can make it more transparent with a better name, or is this the best approach for now?

andrrizzi:
Just to be sure I understand, `number_of_extension_iterations` is just syntax sugar for increasing `number_of_iterations` by the same number, correct?

Lnaden:
`number_of_extension_iterations ` supersedes `number_of_iterations` by reading how many iterations have already been complete, then extending them by the integer argument given to `number_of_extension_iterations`

If `number_of_extension_iterations` is set, `number_of_iterations` is ignored.

Lnaden:
Here is an idea. What if I change `number_of_extension_iterations` to the boolean `extension_simulation` which modifies the `number_of_iterations` to behave like `number_of_extension_iterations` as its implemented right now, default `False`?

That shortens the length of the argument, and removes the ambiguity of what happens when both are specified.

andrrizzi:
I think simulations can already be extended with `number_of_iterations` if you just raise it. Is that not working?

Lnaden:
Yes, that does work, but it requires modifying the file each time you want to extend. If you want to automate that over say, chained PBS jobs, then you have to edit the file each time. This lets you just rerun with the same file and carry on.

Lnaden:
I put the modified extension work in #625

Lnaden:
This has been updated to include the changes to `clusterutils` for LSF, SLURM, and MPICH3

PR #620, opened 2016-12-28
CLI Cleanup and Extened Simulations

Lnaden:
This update is a massive Command Line Interface clean up along with a new feature to extend simulations

* CLI `prepare` and `run` functions have been removed
* `yank script` now has ability to override individual options though the `-o` flag
* `YamlBuild` can accept new, partial YAML data and update existing YAML data in place
* `repex.py` can now run an extension of simulations through a `number_of_extension_iterations`
* Some PEP 8 and documentation fixes
* Some conversions from string % formatting to string.format formatting.
* Fixed setup.py license

Will need review, @andrrizzi would be great to make sure I got the YAML bits right.


Lnaden:
If this breaks something in the YAML, we can fix it later. We won't cut a release until we can confirm nothing breaks.

andrrizzi:
Sorry for the delay! The YAML part looks good to me!

I'm a little afraid that the `override` keyword will get abused and penalize reproducibility, but maybe I'm missing the key use case.

andrrizzi:
> I'm a little afraid that the override keyword will get abused and penalize reproducibility

Ah, no, I guess that's not a problem because the minimal `yaml` file will be exported to the experiment folder. I withdraw my objection!

PR #622, opened 2016-12-30
Cleanup from running the tests.

Lnaden:
* Fixed restraints sometimes choosing numpy int instead of Python ints causing OpenMM crash
* Removed old --platform argument from the `yank script` CLI
* Updated docs to reflect CLI changes

### DIFF
--- a/Yank/commands/script.py
+++ b/Yank/commands/script.py
@@ -25,7 +25,7 @@ usage = """
 YANK script
 
 Usage:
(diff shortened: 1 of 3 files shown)


jchodera:
Looks good! I'd just add an example for the `override` help string---maybe selecting the platform this way?---just to give users a better idea of the most common use case.

jchodera:
Thanks!

PR #629, opened 2017-01-24
Serialize System, State, and Integrator on NaN fail during propagation, 0.15.0

jchodera:
This causes the `System`, `State`, and `Integrator` to be serialized when YANK fails due to `NaN`s in order to aid debugging.

Testing this now.

### DIFF
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -48,7 +48,7 @@ Written by John D. Chodera <jchodera@gmail.com> while at the University of Calif
 
 LICENSE
 
-This code is licensed under the latest available version of the GNU General Public License.
+This code is licensed under the latest available version of the MIT License.
 
 """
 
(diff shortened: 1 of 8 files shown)


jchodera:
I wonder if we should add a test that uses a too-large timestep to force failure?

jchodera:
@Lnaden : Looks like the test failure was stochastic. Let's merge this and get out a new minor version increment?

Lnaden:
We could just update the version number on this PR. Are you wanting to cut a new release based on this?

jchodera:
I think it would be useful to help other folks debug.

Can you update the version number in a separate PR?

Lnaden:
Yes, but is there any harm in doing it on this one? Since its a branch off the main repo, I can just pull down and edit it

jchodera:
Whatever floats your boat.

Lnaden:
Okay, this should be prepped for release if the changes I made look good

PR #639, opened 2017-02-14
New ability to create a Jupyter/Ipython notebook instead of the analyze script

Lnaden:
This does not auto-render, but it does auto generate based on the settings.

This commit includes:
* Template Jupyter Notebook used to construct new ones
* Example notebook fully rendered with real data (can be removed from repo once approval is done)
* Updated version to 0.15.1
* Automatically include the template ipython notebook in the installed repo
* Template is fully load even if YANK is installed as .egg


andrrizzi:
We could also pre-render the notebook if we let the user specify the mandatory settings through the CLI with [something like this](http://nbconvert.readthedocs.io/en/latest/execute_api.html#executing-notebooks-from-the-command-line).

Lnaden:
pre-rendering is something I wanted to have also, but figured it could wait until we further detailed what we wanted in this report. Besides the matplotlib-fu to get the pictures of what analyze is doing, are there other features you can think of that we want in the report on this initial commit?

andrrizzi:
> are there other features you can think of that we want in the report on this initial commit?

Not that I can think of.

jchodera:
This is great!

I'd move as much of the analysis code to `Yank/analysis.py` as possible in order to have the notebook really focus on the data presentation rather than the data analysis.

Lnaden:
I'm not quite sure I follow. All of the data crunching is handled by the `analyze.py` module. Unless you are referring to the last code cell where its running through a subset of commands the `analyze()` function does. I can refactor the `analyze.py` script to avoid some of that code duplicate if that is what you are referring to.

The rest of the code is matplitlib-fu and formatting.

jchodera:
Plotting and formatting code can also be included into analyze.py (or even notebook.py) to minimize the code needed in the notebook, couldn't it?

Lnaden:
That is a good point. I'll work to bring all that code into a `.py` file somewhere. to minimize what is in the notebook.

Lnaden:
@jchodera I have split all of the notebook functions into a new module class which handles all the formatting. Is this more what you had in mind for the notebook?

I also think that we can make improvements on this from here if this initial report looks good.

jchodera:
> @jchodera I have split all of the notebook functions into a new module class which handles all the formatting. Is this more what you had in mind for the notebook?

This is perfect! It looks *much* cleaner now.

> I also think that we can make improvements on this from here if this initial report looks good.

Agreed!

Lnaden:
Great! I also got some input from the people here for improvements we can make with additional details. For now though, Merging!

PR #655, opened 2017-03-14
Update dockerfile to NVIDIA image, Python 3.5

jchodera:
This updates the Dockerfile to use the NVIDIA image and Python 3.5.


jchodera:
Merging this, since it only affects the dockerhub build.

PR #656, opened 2017-03-17
New ReplicaExchange

andrrizzi:
Ready for review. This is a major rework of `repex`. Tests are not supposed to pass yet on Travis, but they pass locally and on the cluster. I'll temporarily merge to the `development` branch to not break `master`.

**Bugfixes**:
- Fix the problem where stored energies at iteration 0 are 0.
- ReplicaExchange simulations can be created in parallel without crashing.

**Features**:
- **Integrated the openmmtools MCMC and states framework into ReplicaExchange**. This is the main new feature. It means:
	- We can use all kinds of propagators and monitor their statistics for each thermodynamic state.
	- We can always compute free energies on a single context as long as thermodynamic states differ only by a portion of the state modeled by a `ComposableState`. The only limitation is that all systems must have the same number of particles.
- Created an `mpi` module with utility functions to remove branching code.
- Separated storage in Reporter class to remove code duplication for analysis.
- Bumped NetCDF format ConventionVersion to 0.2.
    - Generalized thermodynamic state storage. The class decide its own representation in the storage as a dict.
    - Backwards compatibility with previous versions is guaranteed for reading (i.e. analysis), not for writing.
    - The header includes now the proper Yank version.
    - Trying to write on a dataset with old conventions raises an error indicating the last compatible Yank version to use.
- Features of ModifiedHamiltonianExchange have been generalized in ReplicaExchange.
    - ReplicaExchange store the serialization of only 1 system per compatible state to save disk usage.
    - The serialization of the context for debug is performed through `IntegratorMoveError.serialize_error(path)`, which is raised when applying an IntegratorMove results in NaNs.
    - MC rotation/displacement are now independent MCMCMoves objects.
    - Generalized unsampled states that now can be in any number (i.e. not just the expanded fully interacting and non interacting).
    - I have kept `sampling.py` around because it still has a `randomize_ligand` static method which we'll have to move to another place. I'll remove it afterwards.
* Added a bunch of MPI tests. The `TestReplicaExchange` suite can be also tested with MPI on the cluster.

**API**:
The new API is very similar to the old one. You can check [this example](https://github.com/choderalab/yank/blob/new-repex/Yank/repex.py#L1053-L1108) for an overview. It mainly differ from the old one in that we don't pass all the options in the constructor, but they are distributed as parameters of the `MCMCMove`s and the new functions `minimize`/`equilibrate`/`extend`.

Analysis still needs reworking to remove code duplication with `analyze.py`.


Yank/mpi.py
```diff
@@ -0,0 +1,421 @@
...
+# TODO drop this when we drop Python 2 support
+from openmoltools.utils import wraps_py2
+
+logger = logging.getLogger(__name__)
+
+# ==============================================================================
+# GLOBAL VARIABLES
+# ==============================================================================
+
+disable_mpi = False
```
Lnaden: Will this ever be `True`?

Yank/repex.py
```diff
@@ -1162,1633 +1507,482 @@ def _display_citations(self):
...
-
-        Returns
-        -------
-        u_n : numpy array of numpy.float64
-        u   _n[n] is -log q(X_n)
+        # Determine fraction of swaps accepted this iteration.
+        n_swaps_proposed = self._n_proposed_matrix.sum()
+        n_swaps_accepted = self._n_accepted_matrix.sum()
+        swap_fraction_accepted = 0.0
+        if n_swaps_proposed
```
Lnaden: Doesnt this convert it back into unit'd energy?

Yank/mpi.py
```diff
@@ -0,0 +1,421 @@
...
+        The communicator for this node, None if the program doesn't run
+        with MPI.
+
+    """
+    # If MPI execution is forcefully disabled, return None.
+    if disable_mpi:
+        return None
+
+    # If we have already initialized MPI, return the cached MPI communicator.
+    if get_mpicomm._is_initialized:
```
andrrizzi: Yeah, this is the only way in Python to have some sort of "static" function variable. There is indeed an inefficiency bug actually now that you point it out, since I need to set `get_mpicomm._is_initialized = True` also in the if-block below this that handles the non-MPI case.

Yank/mpi.py
```diff
@@ -0,0 +1,421 @@
...
+        return get_mpicomm._mpicomm
+
+    # Check for environment variables set by mpirun. Variables are from
+    # http://docs.roguewave.com/threadspotter/2012.1/linux/manual_html/apas03.html
+    variables = ['PMI_RANK', 'OMPI_COMM_WORLD_RANK', 'OMPI_MCA_ns_nds_vpid',
+                 'PMI_ID', 'SLURM_PROCID', 'LAMRANK', 'MPI_RANKID',
+                 'MP_CHILD', 'MP_RANK', 'MPIRUN_RANK']
+    use_mpi = False
+    for var in variables:
```
andrrizzi: I'm not sure, actually. For now this has always work, but it's for weird cases like that that I've added the `disable_mpi` module global variable.

Yank/mpi.py
```diff
@@ -0,0 +1,421 @@
...
+
+    # Return None if we are not running on MPI.
+    if not use_mpi:
+        logger.debug('Cannot find MPI environment. MPI disabled.')
+        get_mpicomm._mpicomm = None
+        return get_mpicomm._mpicomm
+
+    # Initialize MPI
+    from mpi4py import MPI
+    MPI.COMM_WORLD.barrier()
```
andrrizzi: Hmm. Maybe not! I'll remove it and see if it's giving us problems, thanks!

Yank/mpi.py
```diff
@@ -0,0 +1,421 @@
...
+    mpicomm = MPI.COMM_WORLD
+
+    # Override sys.excepthook to abort MPI on exception
+    def mpi_excepthook(type, value, traceback):
+        sys.__excepthook__(type, value, traceback)
+        sys.stdout.flush()
+        sys.stderr.flush()
+        if mpicomm.size > 1:
```
andrrizzi: Good question! I may be wrong about this, and I honestly forgot the details, but when I implemented the `excepthook` I read somewhere that `MPI_Abort` shouldn't be used to gracefully exit the application, but it's there just to shut everything down when there's no hope for recovery. I added the `mpicomm.size > 1` check because if we have only 1 process, MPI should handle closure automatically, and we won't risk having MPI processes hanging around on the cluster.

Yank/mpi.py
```diff
@@ -0,0 +1,421 @@
...
+
+    # Report initialization
+    logger.debug("MPI initialized on node {}/{}".format(mpicomm.rank+1, mpicomm.size))
+
+    return mpicomm
+
+get_mpicomm._is_initialized = False  # Static variable
+
+
+def run_single_node(rank, task, *args, **kwargs):
```
andrrizzi: Agreed. Unfortunately that works only in Python 2.

Yank/mpi.py
```diff
@@ -0,0 +1,421 @@
...
+    algorithm guarantees that each node will be assigned to the same job_id
+    (i.e. the index of the argument in distributed_args) every time.
+
+    Parameters
+    ----------
+    task : callable
+        The task to be distributed among nodes. The task will be called as
+        task(distributed_args[job_id], *other_args, **kwargs), so the parameter
+        to be distributed must the the first one.
+    distributed_args : iterable
```
andrrizzi: It works for single arg items, but the "task" function can take as first argument a tuple/list, so you can still use this to distribute sets of arguments.

Yank/mpi.py
```diff
@@ -0,0 +1,421 @@
...
+        The return value of distribute depends on the value of this parameter
+        (default is None).
+    sync_nodes : bool, optional
+        If True, the nodes will be synchronized at the end of the
+        execution (i.e. the task will be blocking) even if the
+        result is not shared (default is False).
+
+    Other Parameters
+    ----------------
+    *other_args
```
andrrizzi: Nope! Only `distributed_args` are distributed, this are extra positional arguments that we want to maintain equal for all the parallelized tasks.

Yank/mpi.py
```diff
@@ -0,0 +1,421 @@
...
+
+    """
+    send_results_to = kwargs.pop('send_results_to', None)
+    sync_nodes = kwargs.pop('sync_nodes', False)
+    mpicomm = get_mpicomm()
+    n_jobs = len(distributed_args)
+
+    # If MPI is not activated, just run serially.
+    if mpicomm is None:
+        logger.debug('Running {} serially.'.format(task.__name__))
```
andrrizzi: That's the case if you set `send_results_to=node_you_care_about` (or `'all'` in case all node needs it). In this particular case (i.e. without MPI) we check `send_results_to` only to determine the shape of the returned value. I probably need to document this better.

Yank/repex.py
```diff
@@ -1,752 +1,1428 @@
...
 
 import numpy as np
 import mdtraj as md
 import netCDF4 as netcdf
 
-from .utils import is_terminal_verbose, delayed_termination
+import openmmtools as mmtools
+from openmmtools.states import ThermodynamicState
+
+from yank import utils, mpi, version
```
andrrizzi: I thought relative imports were discouraged. Are you sure this is a problem? I haven't had issues with Python 3 so far with absolute imports.

Yank/repex.py
```diff
@@ -1,752 +1,1428 @@
...
-        >>> [system, positions] = [testsystem.system, testsystem.positions]
-        >>> nvt_state = ThermodynamicState(system=system, temperature=100.0*unit.kelvin)
-        >>> npt_state = ThermodynamicState(system=system, temperature=100.0*unit.kelvin, pressure=1.0*unit.atmospheres)
+        # Restore the title in the metadata.
+        if name == 'metadata':
+            data['title'] = self._storage.title
+        return data
 
-        Test compatibility.
+    def write_dict(self, name, data):
```
andrrizzi: I'll patch this with your custom `YamlDumper/Loader` in a separate PR so that we can generate data that we'll be able to read later.

Yank/repex.py
```diff
@@ -1,752 +1,1428 @@
...
+    >>> T_min = 298.0 * unit.kelvin  # Minimum temperature.
+    >>> T_max = 600.0 * unit.kelvin  # Maximum temperature.
+    >>> temperatures = [T_min + (T_max - T_min) * (math.exp(float(i) / float(nreplicas-1)) - 1.0) / (math.e - 1.0)
+    ...                 for i in range(n_replicas)]
+    >>> thermodynamic_states = [states.ThermodynamicState(system=testsystem.system, temperature=T)
+    ...                         for T in temperatures]
+
+    Initialize simulation object with options. Run with a GHMC integrator.
+
+    >>> move = mcmc.GHMCMove(timestep=2.0*unit.femtoseconds, n_steps=50)
```
andrrizzi: It is! You can pass a list here and I made it so that they are tied to the thermodynamic states (not the replicas) since I thought we may want to slightly tweak the propagation with different degrees of decoupling.

Yank/repex.py
```diff
@@ -1,752 +1,1428 @@
...
+        if replica_mixing_scheme not in self._SUPPORTED_MIXING_SCHEMES:
+            raise ValueError("Unknown replica mixing scheme '{}'. Supported values are {}.".format(
+                replica_mixing_scheme, self._SUPPORTED_MIXING_SCHEMES))
+
+        # Handling default propagator.
+        if mcmc_moves is None:
+            # This will be converted to a list in create().
+            self._mcmc_moves = mmtools.mcmc.LangevinDynamicsMove(timestep=2.0*unit.femtosecond,
+                                                                 collision_rate=5.0/unit.picosecond,
+
```
andrrizzi: I'm not sure I follow. Did you meant this to be a comment of another line?

Yank/repex.py
```diff
@@ -1162,1633 +1507,482 @@ def _display_citations(self):
...
 
-        """
-        if self.platform is None:
-            return self.mm.Context(system, integrator)
-        else:
-            return self.mm.Context(system, integrator, self.platform)
+    @staticmethod
+    def _does_file_exist(file_path):
+        """Check if there is a file at the given path."""
+        return os.path.exists(file_path) and os.path.getsize(file_path) > 0
```
andrrizzi: I think this is for the case when something went wrong in between opening the file and write the first data. Probably irrelevant though.

jchodera:
This looks great! The examples are very nice and clean. Go ahead and merge when you and @Lnaden think this is ready.

andrrizzi:
Great! Merging this whole first round of modifications now.

PR #659, opened 2017-03-24
New restraints API

andrrizzi:
Ready for review. Tests are not expected to pass yet so I'll temporarily merge to `development`.

- The API allow us to specify manually the parameters of the restraints ([here is an example](https://github.com/choderalab/yank/blob/new-restraints/Yank/restraints.py#L1175-L1206).
- I've split the automatic determination of parameters into its own function, because I thought that users implementing their own `Restraint`s may not want to support that functionality.
- I've also implemented a new composable `RestraintState` class to control the `lambda_restraints` parameter, which I've removed from `AlchemicalState` in choderalab/openmmtools#145 ([here is an example](https://github.com/choderalab/yank/blob/new-restraints/Yank/restraints.py#L152-L192)).

### DIFF
--- a/Yank/restraints.py
+++ b/Yank/restraints.py
@@ -387,8 +387,8 @@ class ReceptorLigandRestraint(ABC):
         left undefined in the constructor.
 
         """
-        raise NotImplemented('{} does not support automatic determination of '
-                             'the restraint parameters'.format(self.__class__.__name__))
+        raise NotImplementedError('{} does not support automatic determination of the '
+                                  'restraint parameters'.format(self.__class__.__name__))
 
 
 # ==============================================================================
@@ -618,8 +618,8 @@ class RadiallySymmetricRestraint(ReceptorLigandRestraint):
         """
         # Raise exception only if the subclass doesn't already defines parameters.
         if self._bond_parameters is None:
-            raise NotImplemented('Restraint {} cannot automatically determine '
-                                 'bond parameters.'.format(self.__class__.__name__))
+            raise NotImplementedError('Restraint {} cannot automatically determine '
+                                      'bond parameters.'.format(self.__class__.__name__))
 
     # -------------------------------------------------------------------------
     # Internal-usage
--- a/Yank/restraints.py
+++ b/Yank/restraints.py
@@ -138,105 +138,64 @@ ABC = abc.ABCMeta('ABC', (object,), {})  # compatible with Python 2 *and* 3
 
 
 class ReceptorLigandRestraint(ABC):
-    """
-    Impose a single restraint between ligand and protein to prevent ligand from drifting too far
-    from protein in implicit solvent calculations.
+    """A restraint preventing a ligand from drifting too far from its receptor.
 
-    This restraint strength is controlled by a global context parameter called 'lambda_restraints'.
+    With replica exchange simulations, keeping the ligand close to the binding
+    pocket can enhance mixing between the interacting and the decoupled state.
+    This should be always used in implicit simulation, where there are no periodic
+    boundary conditions.
+
+    This restraint strength is controlled by a global context parameter called
+    'lambda_restraints'. You can easily control this variable through the
+    `RestraintState` object.
 
     Notes
     -----
     Creating a subclass requires the following:
-    * Perform any necessary processing in subclass `__init__` after calling base class `__init__`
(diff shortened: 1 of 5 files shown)


Yank/restraints.py
```diff
@@ -96,330 +107,392 @@ def available_restraint_types():
...
-    """
-    Impose a single restraint between ligand and protein to prevent ligand from drifting too far
-    from protein in implicit solvent calculations.
+    A `ComposableState` controlling the strength of a restraint
+    through its `lambda_restraints` property.
 
-    This restraint strength is controlled by a global context parameter called 'lambda_restraints'.
+    Parameters
+    ----------
+    lambda_restraints : float
```
Lnaden: I have a concern here. By splitting all the `lambda_X` into different `ComposableStates`, is there any way to query all possible `lambda` variables in a single place that does not require the user to loop through stuff?

andrrizzi: Not really without implementing a method in `CompoundThermodynamicState` that does the job. However, all the `lambda_X` parameters will stay in `AlchemicalState` with the exception of `lambda_restraints`. I thought it was cleaner to keep it separated since `alchemy` doesn't implement any restraint-related functionality.

Yank/restraints.py
```diff
@@ -96,330 +107,392 @@ def available_restraint_types():
...
+
+    def determine_missing_parameters(self, thermodynamic_state, sampler_state, topography):
+        """Automatically choose undefined parameters.
+
+        Optionally, a ReceptorLigandRestraint can support the automatic
+        determination of all or a subset of the parameters that can be
+        left undefined in the constructor.
 
         """
+        raise NotImplemented('{} does not support automatic determination of '
```
andrrizzi: Ah! Good catch!

Yank/restraints.py
```diff
@@ -96,330 +107,392 @@ def available_restraint_types():
...
         system = openmm.System()
         system.addParticle(1.0 * unit.amu)
         system.addParticle(1.0 * unit.amu)
         force = self._create_restraint_force(0, 1)
         # Disable the PBC if on for this approximation of the analytical solution
-        try:  # This was added in OpenMM 7.1
-            force.setUsesPeriodicBoundaryConditions(False)
-        except AttributeError:
-            pass
+        force.setUsesPeriodicBoundaryConditions(False)
```
andrrizzi: Right! I forgot about that. I set it back to `openmm>=7.0.1` few months ago to make it work on the cluster. I'll revert it in this PR.

Yank/restraints.py
```diff
@@ -766,89 +1168,367 @@ class Boresch(OrientationDependentRestraint):
...
+    ...     restraint.restrain_state(thermodynamic_state)
+    ...
+    There are undefined parameters. Choosing restraint parameters automatically.
+
+    Get standard state correction.
+
+    >>> correction = restraint.get_standard_state_correction(thermodynamic_state)
+
     """
+    def __init__(self, restrained_atoms=None,
```
andrrizzi: Not if the user specify the atoms manually. We could figure out a way to do that by inspecting the bonds in the `ThermodynamicState` system when `restrain_state(thermodynamic_state)` is called, and raising an error if the 6 atoms are connected (directly or indirectly) by bonds.

In the near future though, all the atom selections that the user needs to make will probably be easily handled by the `Topography` object with DSL and/or SMIRKS specification, so it may not be worth it.

Yank/restraints.py
```diff
@@ -766,89 +1168,367 @@ class Boresch(OrientationDependentRestraint):
...
+            """Cast the passed_unit into md unit system for integrand lambda functions"""
+            return passed_unit.value_in_unit_system(unit.md_unit_system)
+
+        # Shortcuts variables.
+        pi = np.pi
+        kT = thermodynamic_state.kT
+        p = self  # For the parameters.
+
+        # Radial
+        sigma = 1 / unit.sqrt(p.K_r / kT)
```
andrrizzi: Good idea! I checked on the interpreter and `unit.sqrt` raises a `ValueError("math domain error")` when called with a negative argument, so I'm tempted to leave the handling of eventual errors to it. The error trace is pretty clear and this way I won't have to add a bunch of `assert` statements.

andrrizzi:
Ok! Merging this to 'development' since it's blocking. I can make modifications in a separate PR in case.

PR #660, opened 2017-03-29
New extendable Analyze API

Lnaden:
This PR has the new API for the analyze module.

This is still a work in progress as there is old code left in for comparison, review, and debate with the new API. Tests still need to be written but strongly depend on the integration of other API actions.

Most of the work will require cleanup, and some functionality changes (like creating a factory for the `MultiPhaseAnalyzer` through a `.create` function), but the major API and scientific changes are in. I wanted to open this PR now so the majority of the API and design can be discussed before trying to finalize and integrate code.

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -17,18 +17,18 @@ import os
 import os.path
 
 import yaml
+import abc
+import copy
 import numpy as np
 
 from .repex import Reporter
 import netCDF4 as netcdf  # netcdf4-python
 
-from pymbar import MBAR  # multistate Bennett acceptance ratio
+from pymbar import MBAR  # multi-state Bennett acceptance ratio
 from pymbar import timeseries  # for statistical inefficiency analysis
 
 import mdtraj
 import simtk.unit as units
-
-import abc
 
 from . import utils
 
@@ -55,14 +55,38 @@ def generate_phase_name(current_name, name_list):
     elif current_name in name_list:
         name = current_name + str(counter)
         while name in name_list:
-            counter +=1
+            counter += 1
             name = current_name + str(counter)
     else:
         name = current_name
     return name
 
 
-class ObservablesRegitry(object):
+def get_analyzer(file_path):
+    """
+    Utility function to convert storage file to a Reporter and Analyzer by reading the data on file
+
+    For now this is mostly placeholder functions, but creates the API for the user to work with.
+    """
+    reporter = Reporter(file_path)  # Eventually extend this to get more reporters, but for now simple placeholder\
+    """
+    storage = infer_storage_format_from_extension('complex.nc')  # This is always going to be nc for now.
+    metadata = storage.metadata
+    sampler_class = metadata['sampler_full_name']
+    module_name, cls_name = sampler_full_name.rsplit('.', 1)
+    module = importlib.import_module(module_name)
+    cls = getattr(module, cls_name)
+    reporter = cls.create_reporter('complex.nc')
+    """
+    # Eventually change this to auto-detect simulation from reporter:
+    if True:
+        analyzer = RepexPhase(reporter)
+    else:
+        raise RuntimeError("Cannot automatically determine analyzer for Reporter: {}".format(reporter))
+    return analyzer
+
+
+class _ObservablesRegistry(object):
     """
     Registry of computable observables.
     """
@@ -72,9 +96,9 @@ class ObservablesRegitry(object):
         Set of observables which are derived from the subsets below
         """
         observables = set()
-        for subset in (ObservablesRegitry.observables_defined_by_two_states(),
-                       ObservablesRegitry.observables_defined_by_single_state(),
-                       ObservablesRegitry.observables_defined_by_phase()):
+        for subset in (_ObservablesRegistry.observables_defined_by_two_states(),
+                       _ObservablesRegistry.observables_defined_by_single_state(),
+                       _ObservablesRegistry.observables_defined_by_phase()):
             observables = observables.union(set(subset))
         return tuple(observables)
 
@@ -82,8 +106,8 @@ class ObservablesRegitry(object):
     @staticmethod
     def observables_with_error():
         observables = set()
-        for subset in (ObservablesRegitry.observables_with_error_adding_quadrature(),
-                       ObservablesRegitry.observables_with_error_adding_linear()):
+        for subset in (_ObservablesRegistry.observables_with_error_adding_quadrature(),
+                       _ObservablesRegistry.observables_with_error_adding_linear()):
             observables = observables.union(set(subset))
         return tuple(observables)
 
@@ -116,7 +140,7 @@ class ObservablesRegitry(object):
         Observables which are defined by the phase as a whole, and not defined by any 1 or more states
         e.g. Standard State Correction
         """
-        return 'restraints'
+        return 'standard_state_correction'
 
 
 class YankPhaseAnalyzer(ABC):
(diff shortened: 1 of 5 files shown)


Yank/repex.py
```diff
@@ -259,13 +259,15 @@ def write_thermodynamic_states(self, thermodynamic_states, unsampled_states):
                 # Write state as dictionary.
                 self.write_dict('{}/state{}'.format(state_type, state_id), serialized_state)
 
-    def read_sampler_states(self, iteration):
+    def read_sampler_states(self, iteration=slice(None)):
```
Lnaden: Im fine with converting it back into an `arg` and allowing for the `slice(None)` on demand from the users. When I added those in, I kind of blindly just went through, only realizing after some of them would not work. So lets revert that.

Yank/repex.py
```diff
@@ -431,30 +433,36 @@ def write_mcmc_moves(self, mcmc_moves):
...
         -------
         energy_thermodynamic_states : n_replicas x n_replicas numpy.ndarray
-            energy_thermodynamic_states[i][j] is the reduced potential computed at
-            SamplerState sampler_states[i] and ThermodynamicState thermodynamic_states[j].
+            energy_thermodynamic_states[iteration, i, j] is the reduced potential computed at
+            SamplerState sampler_states[iteration, i] and ThermodynamicState thermodynamic_states[iteration, j].
         energy_unsampled_states : n_replicas x n_unsampled_states numpy.ndarray
-            energy_unsampled_states[i][j] is the reduced potential computed at SamplerState
-            sampler_states[i] and ThermodynamicState unsampled_thermodynamic_states[j].
+            energy_unsampled_states[iteration, i, j] is the reduced potential computed at SamplerState
```
Lnaden: Yes, documentation will be tricky here since we can support a `slice(None)` and an int which changes the output

Yank/analyze.py
```diff
@@ -32,27 +35,1145 @@
...
+        self._sign = '+'
+        self._equilibration_data = None  # Internal tracker so the functions can get this data without recalculating it
+        # External properties
+        self._reference_states = None  # initialize the cache object
+        self.reference_states = reference_states
+        self._mbar = None
+        self._kT = None
+
+    @property
+    def name(self):
```
Lnaden: I disagree, the `name` is important with `MultiPhaseAnalyzers` since the name that is used is really the only way to make a unique ID to the `SinglePhase` that is not the object itself. Also, when several `MultiPhases` are combined, the names can get shuffled and adjusted based on order of operation (i.e. `+` is not a perfectly commutable operator) . By allowing the user to set the name, or by allowing the `read_directory` function to set the name, we provide a way to id the `SinglePhase` analyzers in a human readable.

Yank/analyze.py
```diff
@@ -32,27 +35,1145 @@
...
+        """User-readable name of the phase"""
+        return self._name
+
+    @name.setter
+    def name(self, value):
+        self._name = value
+
+    @property
+    def observables(self):
```
Lnaden: This does not return the `_ObervablesRegistry` object, it determines which observables it can compute from the registry, and stores that as an internal list. I do agree that we could hide this and use inspection in the `MultiPhase` analyzers to determine which are shared properties.

Yank/analyze.py
```diff
@@ -32,27 +35,1145 @@
...
+
+    @property
+    def mbar(self):
+        """Access the MBAR object tied to this phase"""
+        if self._mbar is None:
+            self._create_mbar_from_scratch()
+        return self._mbar
+
+    @property
+    def reference_states(self):
```
Lnaden: Yeah, we should rename this to remove name collision.

Yank/analyze.py
```diff
@@ -32,27 +35,1145 @@
...
+            where an energy drawn from sampled state [k] is evaluated in sampled state [l] at iteration [n]
+        unsampled_energy_matrix
+            Has shape (K, L, N) = (number of sampled states, number of UN-sampled states, number of iterations)
+            Indexed by [k,l,n]
+            where an energy drawn from sampled state [k] is evaluated in un-sampled state [l] at iteration [n]
+        """
+        raise NotImplementedError()
+
+    # This SHOULD be an abstract static method since its related to the analyzer, but could handle any input data
+    # Until we drop Python 2.7, we have to keep this method
```
Lnaden: Python 2 does not support
```python
@staticmethod
@abc.abstractmethd
def function():
...
```
It raises an error... for... reasons.

We did not talk about it.

Yank/analyze.py
```diff
@@ -32,27 +35,1145 @@
...
+        cast_data = np.asarray(data)
+        data_shape = cast_data.shape
+        # Since we already have g, we can just pass any appropriate shape to the subsample function
+        indices = timeseries.subsampleCorrelatedData(np.zeros(data_shape[axis]), g=subsample_rate)
+        subsampled_data = np.take(cast_data, indices, axis=axis)
+        return subsampled_data
+
+    # Private Class Methods
+    @abc.abstractmethod
+    def _prepare_mbar_input_data(self, *args, **kwargs):
```
Lnaden: Thats probably a good idea. I think this `abstractmethod` was leftover from when this was a public function in my original design.

Yank/analyze.py
```diff
@@ -32,27 +35,1145 @@
...
+        phase_pass = {'phases': phases, 'signs': signs, 'names': names}
+        return MultiPhaseAnalyzer(phase_pass)
+
+    def __add__(self, other):
+        return self._combine_phases(other, operator='+')
+
+    def __sub__(self, other):
+        return self._combine_phases(other, operator='-')
+
+    def __neg__(self):
```
Lnaden: Kind of, I could not find a way to pass the negative operator in a non-preserving manner without doing a deepcopy which would be super memory inefficient and break the ability for adjusting `SinglePhase` `reference_state`'s that propagate to all `MultiPhase` analyzers which use them, which is why all the functions which combine phases actually reverse them.

E.g.
```python
>>> x = RepexAnalyzer(...)
>>> print(x.sign)
  '+'
>>> -x
>>> print(x.sign)
  '-'
```
which is not the exact behavior we want. But I could not find a way to detect that both negative operator has been used in a more complex operation, and not as strait re-assignment. (can't overload the `=` operator)

Yank/analyze.py
```diff
@@ -32,27 +35,1145 @@
...
+        """Internally handle the internal sign"""
+        if self._sign == '+':
+            self._sign = '-'
+        else:
+            self._sign = '+'
+        return self
+
+
+class RepexPhase(YankPhaseAnalyzer):
```
Lnaden: Agreed

Yank/analyze.py
```diff
@@ -32,27 +35,1145 @@
...
+            str_row = ""
+            for j in range(nstates):
+                str_row += "{:8.3f}".format(dDeltaf_ij[i, j])
+            logger.info(str_row)
+
+        # Return free energy differences and an estimate of the covariance.
+        free_energy_dict = {'value': Deltaf_ij, 'error': dDeltaf_ij}
+        self._computed_observables['free_energy'] = free_energy_dict
+
+    def get_free_energy(self):
```
Lnaden: I like the idea of having both a `.free_energy` and a `.free_energies`, but I don't know how best make that work with the `ObservablesRegistry`

Yank/analyze.py
andrrizzi: Ah, I see, I didn't see how it was used in `MultiPhaseAnalyzers`. I didn't work out the implementation details, but we may still be able to remove it by changing the argument in the constructor to take a `list` (or something ordered) instead of a `dict`.

The human readable representation would indeed be nice to have, but I wouldn't tie API design decisions to this feature. One work around that pops into my head is to print print the extension-less name of the storage file used by the reporter, which is more or less what our top level functions will probably do manually.

Yank/analyze.py
andrrizzi: Ah, you're right! I didn't notice that. We should keep thinking about this.

andrrizzi:
Merging this to `development` so that we can keep things going.

Issue #643, opened 2017-02-23
For a series of mixed solvent systems, I have to minimize outside of Yank to avoid crashes


davidlmobley:
I've been troubleshooting a series of cases (I mentioned in passing previously) where I have systems set up via `packmol` (in our `SolvationToolkit`) which I can't simulate in Yank without them crashing as soon as I begin dynamics. *However*, if I add a minimization in OpenMM prior to going into Yank (without perturbing the system!), everything works fine. I am attaching input files to reproduce.

These are systems which always minimize fine in GROMACS for us without issue, but apparently Yank is having trouble minimizing them for some reason. Originally we thought perhaps it was a box vector issue, but I've checked and the box vectors and periodicity seem to be handled correctly. So as far as I can tell it really seems to be an issue with how Yank is minimizing (at least, if I add an "outside of Yank" minimization as a standard procedure in my runs, I avoid these crashes entirely). 

I haven't tracked back through the code to see how the minimization Yank is doing is different from what I'm doing in OpenMM outside of Yank, but perhaps you guys will have insights. We ARE running with openmm 7.1. 

`gaff_in_yank_gromacs_clean` is a clean setup I run which invariably crashes; `gaff_in_yank_gromacs_clean_minfirst` is a setup which adds minimization (see `minimize.py` there) first before running Yank. There is a `run_1.sh` shell script in each directory which does the run on our queueing system. 

One perhaps important note: Systems coming out of `packmol` are typically less well packed and under greater forces (and thus in greater need of minimization) than typical systems we might solvate in a pre-equilibrated box of water (e.g. systems from `tleap` or other mechanisms, I would guess), so this is probably a more stringent test of whether minimization is adequate than most systems you've looked at.

[gaff_in_yank_gromacs_clean.tar.gz](https://github.com/choderalab/yank/files/798004/gaff_in_yank_gromacs_clean.tar.gz)
[gaff_in_yank_gromacs_clean_minfirst.tar.gz](https://github.com/choderalab/yank/files/798005/gaff_in_yank_gromacs_clean_minfirst.tar.gz)

andrrizzi:
Thanks for the input files, this helps a lot. I'll try to figure out what's happening over the weekend.

andrrizzi:
[...]

PR #662, opened 2017-03-31
New application layer

andrrizzi:
This is the last big chunk of modifications! Tests are not supposed to pass until this is merged with the changes in the `new-repex` branch. I'll start running validation simulations as soon as we have gathered all the pieces into the branch `development`. Here's a list of changes in this PR:

**Bugfixes**
- Ions of any charge (instead of just single charge) are supported when figuring out which counterions to alchemically modify to keep the periodic solvation box neutral with charged ligands.

**New features**
- Implemented `AlchemicalPhase` that wraps around a generic sampler. The preliminary minimization of the positions in the reference state discussed in #643 is implemented in `AlchemicalPhase.minimize`.
- `YamlBuilder` now offer two methods `build_experiments` and `run_experiments`. The former sets up and iterates over all the possible experiments, which are provided as `YamlExperiment` objects. These can be further tweaked through the API to set things like the propagator, custom restraints, and others that is not possible to set through the YAML script. `run_experiments` does what `builds_experiments` was doing before.

**Changes**
- Removed options `show_mixing_statistics`, `show_energies`, `online_analysis`, `online_analysis_min_iterations` from YAML syntax which are not supported by `ReplicaExchange` anymore.
- Divided the `YamlBuilder` options in general options (e.g. output directory, verbose) from experiment options (e.g. temperature/pressure, timestep), where the former cannot be specified in the experiment section (and thus they cannot be combinatorially expanded). This came with a lot of code simplification and we drop support for use cases that don't actually make much sense.
- `ligand_dsl` in system files section is now used only for ligand-receptor systems and it has thus become an optional attribute. If a ligand is specified but receptor atoms are not found, an error is raised.
- Updated default values of `collision_rate`: set to 1/ps for production and 90/ps for equilibration.
- Two new options `switch_phase_every` (experiment option), and `switch_experiment_every` that allow to make progress on multiple phases/experiments together (name to be discussed).

### DIFF
--- a/Yank/__init__.py
+++ b/Yank/__init__.py
@@ -6,15 +6,14 @@ YANK
 """
 
 # Define global version.
-from . import version #Needed for yank 3.X.
+from . import version  # Needed for yank 3.X.
 __version__ = version.version
 
 # Self module imports
 from . import utils
 from . import repex
-from . import sampling
 from . import restraints
 from . import pipeline
 from . import yamlbuild
 from . import analyze
-from .yank import Yank
+from .yank import Topography, AlchemicalPhase
--- a/Yank/pipeline.py
+++ b/Yank/pipeline.py
@@ -16,26 +16,161 @@ Utility functions to help setting up Yank configurations.
 import os
 import inspect
 import logging
-logger = logging.getLogger(__name__)
+import itertools
 
 import mdtraj
+import numpy as np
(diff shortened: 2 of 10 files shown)


andrrizzi:
Merging this to `development` branch to keep things going. We can modify this in a separate PR.

PR #663, opened 2017-04-03
Fix development tests

andrrizzi:
The last bugfixes to prepare the `development` branch for production and to make Travis tests pass.

Currently, the only change in this PR is the new serialization of dictionaries as strings, and few bugfixes.


andrrizzi:
Ah it looks I can't trigger Travis tests unless I'm opening a PR to `master`. Beside the few bugfixes, the only thing changed is that the `Reporter` serializes `dict`s as strings now.

@Lnaden I changed a little bit your `Loader`/`Dumper` with an explicit constructor/representer for `ndarray`s because otherwise `pyyaml` uses its pickle representation, which may change with future versions of `numpy`. I thought this way we cab be in control of when we want to break backwards compatibility. It would be great if you could take a [look at it](https://github.com/choderalab/yank/blob/fix-development-tests/Yank/repex.py#L1835-L1883).

Lnaden:
> Ah it looks I can't trigger Travis tests unless I'm opening a PR to `master`

Yes you can. Add `- BRANCHNAME` to the `.travis.yaml` file in the `branches: only:...` section.

>  It would be great if you could take a look at it.

Will do

andrrizzi:
> Yes you can. Add - BRANCHNAME to the .travis.yaml file in the branches: only:... section.

Awesome. Will do that then, thanks for the tip.

Lnaden:
Why would what `numpy` care what `pyyaml` does with its arrays?

andrrizzi:
> Why would what numpy care what pyyaml does with its arrays?

The problem is exactly that they don't care. Right now `pyyaml` uses the `ndarray` pickle representation, but if the `numpy` folks decide to change that representation in the future, then our nc files will automatically become incompatible with previous versions, unless we pin `numpy` to the last version that works.

Lnaden:
Im not sure they will be breaking the ability to pickle any time soon based on the [save format docs](https://docs.scipy.org/doc/numpy/neps/npy-format.html). I don't see a problem either way except for the possible memory space consumed by storing arrays as lists, but overall it looks fine. Is the check for the `0` size dimension so we dont write an incompatible empty list to the zero-width array?

andrrizzi:
I'm not sure it's just a matter of saving format. This is the native pyYAML representation of an array:
```python
>>> import numpy as np
>>> import yaml
>>> a = np.array([1.0, 2, 3, 4])
>>> print(yaml.dump(a))
"""!!python/object/apply:numpy.core.multiarray._reconstruct
args:
- !!python/name:numpy.ndarray ''
- !!python/tuple [0]
- !!binary |
  Yg==
state: !!python/tuple
- 1
- !!python/tuple [4]
- !!python/object/apply:numpy.dtype
  args: [f8, 0, 1]
  state: !!python/tuple [3, <, null, null, null, -1, -1, 0]
- false
- !!binary |
  AAAAAAAA8D8AAAAAAAAAQAAAAAAAAAhAAAAAAAAAEEA="""
```
when loaded, I think `pyyaml` calls `numpy.core.multiarray._reconstruct()` to recreate the object, but that's marked as a private member so it may change its name/signature in the future versions.

> Is the check for the 0 size dimension so we don't write an incompatible empty list to the zero-width array?

Yep.

> except for the possible memory space consumed by storing arrays as lists

I agree. Hopefully we'll never have to store big arrays as dicts, but we can always convert the python list in its binary representation if we'll have to. For short arrays the `numpy` string representation turns out to be longer than the simple list (and definitely not human-readable).

Lnaden:
Looks fine then, if it causes problems going forward, we can affix a storage version on to it and call it good. Could you also edit the `clusterutils` storage module PR as well, or branch it and open your own PR into storage itself?

andrrizzi:
> Could you also edit the clusterutils storage module PR as well?

Sure!

andrrizzi:
Tests finally pass! I think we're ready to merge to `master`! I'm just waiting for `openmmtools 0.9.1` to be released which include a bugfix we need (I've been temporarily tested with `openmmtools-dev`).

@Lnaden This is mostly fixing tests. The only new bit I've added since the last time is the [code that adds the restraint to the fully interacting expanded state](https://github.com/choderalab/yank/blob/fix-development-tests/Yank/yank.py#L584-L592). Can you take a look at it when you have the chance before I merge everything?

Lnaden:
The new restraint code checks look fine.

Have we updated the `commands` to support the new framework yet?

andrrizzi:
> Have we updated the commands to support the new framework yet?

I think so! You fixed the `analyze` command and I fixed the `script`. Is there anything else I forgot? The `status` command doesn't use the new `Reporter` yet, but it should still works since it only read the positions dimensions directly from the nc file, and we didn't change how they are saved.

Lnaden:
Should be good then

PR #666, opened 2017-04-10
Add checks for individual OpenEye tools

Lnaden:
Every function which uses OpenEye tools now only checks for the OpenEye licenses it needs, not all of them blindly.

I noticed that the OEOmega tools are not actually used by anything. Some of the OpenMolTools we use check for OEOmega, but don't use them either and I think are actually typos. e.g. 
```python
if not oeiupac.OEIUPACIsLicensed(): raise(ImportError("Need License for OEOmega!"))
```

This check is not harming anything, but its an observation.

### DIFF
--- a/Yank/utils.py
+++ b/Yank/utils.py
@@ -11,6 +11,7 @@ import glob
 import shutil
 import inspect
 import logging
+import importlib
 import itertools
 import subprocess
 import collections
@@ -1090,16 +1091,24 @@ class Mol2File(object):
 
 # OpenEye functions
 # ------------------
-def is_openeye_installed():
+def is_openeye_installed(oetools=('oechem', 'oequacpac', 'oeiupac', 'oeomega')):
+    # Complete list of module: License check
+    tools_license = {'oechem':'OEChemIsLicensed',
+                     'oequacpac': 'OEQuacPacIsLicensed',
+                     'oeiupac': 'OEIUPACIsLicensed',
+                     'oeomega': 'OEOmegaIsLicensed'}
+    tool_keys = tools_license.keys()
+    # Cast oetools to tuple if its a single string
+    if type(oetools) is str:
+        oetools = (oetools,)
     try:
-        from openeye import oechem
-        from openeye import oequacpac
-        from openeye import oeiupac
-        from openeye import oeomega
-
-        if not (oechem.OEChemIsLicensed() and oequacpac.OEQuacPacIsLicensed()
-                and oeiupac.OEIUPACIsLicensed() and oeomega.OEOmegaIsLicensed()):
-            raise ImportError
+        for tool in oetools:
+            if tool in tool_keys:
+                # Try loading the module
+                module = importlib.import_module('openeye', tool)
+                # Check that we have the license
+                if not getattr(module, tools_license[tool])():
+                    raise ImportError
     except ImportError:
         return False
     return True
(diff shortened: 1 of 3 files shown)


Yank/yamlbuild.py
```diff
@@ -814,8 +816,8 @@ def _setup_molecules(self, *args):
             if 'antechamber' in mol_descr:
                 # Generate charges with OpenEye if requested
                 if 'openeye' in mol_descr:
-                    if not utils.is_openeye_installed():
-                        err_msg = ('Cannot find OpenEye toolkit to compute charges '
+                    if not utils.is_openeye_installed(oetools=('oechem', 'oequacpac')):
```
andrrizzi: We probably need to check for `oeomega` here.

Yank/utils.py
```diff
@@ -1090,16 +1091,32 @@ def net_charge(self, value):
...
+                     'oeomega': 'OEOmegaIsLicensed'}
+    tool_keys = tools_license.keys()
+    # Cast oetools to tuple if its a single string
+    if type(oetools) is str:
+        oetools = (oetools,)
+    tool_set = set(oetools)
+    valid_tool_set = set(tool_keys)
+    if tool_set & valid_tool_set == set():
+        # Check for empty set intersection
+        raise ValueError
```
andrrizzi: Not sure if this is intended, but this is outside of the `try ... except` clause so it want have the same message as below. Is there a reason I'm missing to why you're not just raising `ValueError` with the message here?

Lnaden: Logic typo, My brain had that inside the `try..except` so I was trapping it in the same block.

Yank/utils.py
```diff
@@ -1090,16 +1091,24 @@ def net_charge(self, value):
...
-        from openeye import oechem
-        from openeye import oequacpac
-        from openeye import oeiupac
-        from openeye import oeomega
-
-        if not (oechem.OEChemIsLicensed() and oequacpac.OEQuacPacIsLicensed()
-                and oeiupac.OEIUPACIsLicensed() and oeomega.OEOmegaIsLicensed()):
-            raise ImportError
+        for tool in oetools:
+            if tool in tool_keys:
```
Lnaden: If the input `oetools` is a single string, I would agree, but because you can feed in multiple tools, I have to iterate through each individual tool, then check them against the set of valid tools. I then need to act on each tool individually. If I only needed the check, I could just look for a non-strict subset of the two lists

Yank/yamlbuild.py
```diff
@@ -725,11 +725,13 @@ def _setup_molecules(self, *args):
             # we update the 'filepath' key also for OpenEye-generated molecules so
             # we don't need to keep track of the molecules we have already generated
             if extension is None or extension == '.smiles' or extension == '.csv':
-                if not utils.is_openeye_installed():
+                if not utils.is_openeye_installed(oetools=('oechem','oeiupac','oequacpac')):
```
Lnaden: Fixed with the OEOmega comment you gave broadly. I did not follow the checks of OpenMolTools through, I figured each function had its complete set of checks as headers, I was wrong apparently.

Yank/yamlbuild.py
Lnaden: Fixed as well

Lnaden:
I have one other thing I thought of I want to change to this first.

Lnaden:
@andrrizzi I made a change to the `selftest` feature to check for the specific individual licenses that we can use so users get a better sense of what they can and can't do with OpenEye

PR #669, opened 2017-04-25
Fix extract_trajectory from nc file

andrrizzi:
The function was broken when we broke backwards compatibility of ReplicaExchange storage format.

This is only a quick and dirty solution since we have other priorities. We should rewrite this after the analysis API will be stable.

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -21,6 +21,7 @@ import abc
 import copy
 import numpy as np
 
+import openmmtools as mmtools
 from .repex import Reporter
 import netCDF4 as netcdf  # netcdf4-python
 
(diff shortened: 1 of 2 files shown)


andrrizzi:
@Lnaden can you take a look at the error I'm getting with Python 3.4 tests? I think it's related to the changes in the openeye license checks in #666, and I wonder if you already know what is it about. Otherwise, I'll try to track it down.

Lnaden:
Ill take a look

Lnaden:
I think I have the fix, let me pull this PR and add it here

Lnaden:
Well that didn't work. Give me a bit, I'll try to fix it for you

andrrizzi:
Thanks!

Lnaden:
Okay, that should fix it

Lnaden:
Woo! That fixed it and should also fix the nightlies when this gets merged in

andrrizzi:
Awesome! Thanks!

PR #673, opened 2017-05-04
Add support for non-TIP3P solvation

### LINKED ISSUE #378: Add support for non-TIP3P explicit solvent models
Do we currently support explicit solvents other than TIP3P? If so, how are they specified?

andrrizzi:
Fix #378: Add support for non-TIP3P explicit solvent models.

Should be ready for review if tests pass.

### DIFF
--- a/Yank/tests/test_utils.py
+++ b/Yank/tests/test_utils.py
@@ -281,20 +281,20 @@ def test_TLeap_script():
 
     tleap = TLeap()
     tleap.load_parameters('oldff/leaprc.ff99SBildn', 'leaprc.gaff')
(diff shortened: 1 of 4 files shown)


jchodera:
Could we make sure to [update the docs with this change](https://github.com/choderalab/yank/tree/master/docs/yamlpages) as well? It's not clear to me from reading the code which section in the YAML is expanded and what the allowed tokens are.

andrrizzi:
Of course! I'm planning to do a docs update in a separate PR.

The section expanded and the allowed tokens are described in the issue I linked on top.

andrrizzi:
Website docs for this were added in #682 .

Thanks!

PR #674, opened 2017-05-04
Add support for temperatures at intermediate states in YAML.

### LINKED ISSUE #668: Intermediate states at increased temperature
Can you post suggestions on how to run an intermediate state at a different temperature than the end states?

andrrizzi:
Fix #668 .

This was so easy. Should be ready for review if tests pass.


Lnaden:
Can you add documentation to this in the docs?

andrrizzi:
> Can you add documentation to this in the docs?

I plan to do it soon in a separate PR. There are several changes I already need to make to the docs.

jchodera:
This looks great!

My only question is how much error checking we might want to add to make sure users don't have mismatched temperatures. In principle, the `lambda=0` state can have an arbitrarily high temperature as long as it is matched between all the alchemical phases, but it may be simpler just to add a check to ensure that the `lambda=0` and `lambda=1` temperatures/pressures match within each phase.

andrrizzi:
Thanks!

I can definitely add a temperatures/pressures check inside `AlchemicalPhase.create` before merging.

andrrizzi:
Done! Merging (tagging @daveminh).

Issue #41, opened 2014-04-11
Find ways to reduce file sizes


jchodera:
Michael Shirts mentioned that his datasets for T4 lysozyme using old YANK were 44GB total for about 20 ligands.  For his work on larger sets of proteins, he is generating ~4TB of data.

We should explore ideas for cutting down file sizes.  Some obvious ones:
- Enabling NetCDF compression by default
- Saving checkpoint data less frequently, but energy data more frequently

Lnaden:
Handle in post yank 1.0

Lnaden:
Do we think this can be closed now as #675 has added the checkpointing feature, or should we do more to reduce file sizes?

andrrizzi:
Maybe we should keep this around and rename it to something like "Quantize floats to reduce file size"? The updated link to the mdtraj issue is mdtraj/mdtraj#287 (the one above gives a not found error).

jchodera:
I believe #704 has greatly helped with reducing file sizes for the analysis file. We can reopen/revisit as needed.

Issue #535, opened 2016-10-20
Move Yank class away from parsing and forwarding options to other classes


andrrizzi:
Since all the relevant objects are initialized inside `Yank` (i.e. restraint, alchemical factory, sampling class), we need to pass all the parameters for all classes in a big `kwargs` dictionary. Then `Yank` has to use introspection to separate them by class to initialize all the objects correctly. I propose to make an effort to move the initialization of these objects outside `Yank`.

From a user perspective, the final interface would become something like

``` python
sampler = ModifiedHamiltonianExchange(**sampler_options)
factory = AlchemicalFactory(**alchemical_options)
restraint = Harmonic(**harmonic_options)
yank = Yank(store_directory, sampler, factory, **yank_options)
yank.create_phases(thermodynamic_state, restraint, *alchemical_phases)
```

This is of course only to give an idea, not an actual proposal, but I would implement the restraint part this way now, if we agree that the API should move towards this direction before 1.0 release. I realize that we may need to change a lot of things for this to happen, but I really believe this will turn out to be much easier to maintain and extend in the future.

**Question:** do we want to attach restraints to phases and then simply perform checks to determine if the restraint is appropriate for that type of phase (i.e. raise an exception if it's not a complex), or pass only a single `restraint` argument for all phases and adding the restraint automatically (as we currently do)?

jchodera:
I like it!  

However, to have `yank.create_phases()` be able to create all the phases, all of these things have to be turned into factories that can create, for example, a new set of restraints or a new sampler for each phase. So it would probably need to look more like
```python
sampler_factory = ModifiedHamiltonianExchangeFactory(**sampler_options)
alchemical_factory = AlchemicalFactory(**alchemical_options)
restraint_factory = HarmonicFactory(**harmonic_options)
yank = Yank(store_directory, sampler_factory, alchemical_factory, **yank_options)
yank.create_phases(thermodynamic_state, restraint_ractory, *alchemical_phases)
```
Maybe there's a way to avoid factory overload?

> Question: do we want to attach restraints to phases and then simply perform checks to determine if the restraint is appropriate for that type of phase (i.e. raise an exception if it's not a complex), or pass only a single restraint argument for all phases and adding the restraint automatically (as we currently do)?

Does `Yank` know that certain kinds of phases should be different from others? If so, it can probably determine whether or not a restraint needs to be created and added.

andrrizzi:
> all of these things have to be turned into factories

[...]

PR #675, opened 2017-05-04
Checkpointing Feature

Lnaden:
This PR provides the ability for YANK to checkpoint simulations to save space and time when writing to file, at the cost of writing less frequently

This commit does the following:
* Splits all storage files into 2 files, one for checkpointing and one for analysis
* The two files are used for infrequent, and frequent writes respectively
* All file paths are now on a "base" and the individual files are derived from that base
* Most choices about which file gets which data is handled in the Reporter class
* Added `checkpoint_interval` option to specify how frequently writes occur on the checkpoint relative to the iterations
* Simulations are resumed only from the most recent checkpoint
* Data accessed from checkpoint file which is not on checkpoint_interval returns None objects
* Data cannot be written to checkpoint file at non-checkpoint_interval iterations (logger raises a warning, no action taken)
* Fixed an un-documented bug where the Health Reports were not being picked up correctly

### DIFF
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -70,10 +70,10 @@ class Reporter(object):
 
     Parameters
     ----------
-    base_storage : str
-        The path to the files base name, this should not include a file suffix
-        The multiple storage files will be saved as two files named
-            "base_storage + _checkpointing" and "base_storage + _analysis"
+    storage : str
+        The path to the storage file for analysis.
+        A second checkpoint file will be determined from either checkpoint_storage_file or automatically based on
+            the storage option
         In the future this will be able to take Storage classes as well.
     open_mode : str or None
         The mode of the file between 'r', 'w', and 'a' (or equivalently 'r+').
@@ -85,14 +85,26 @@ class Reporter(object):
         Checkpoint information cannot be written on iterations which where iteration % checkpoint_interval != 0.
         Attempting to read checkpointing information results in a masked array where only entries which were written
             are unmasked
+    checkpoint_storage_file : str or None, optional
+        Optional name of the checkpoint point file. This file is used to save trajectory information and other less
+            frequently accessed data.
+        If None: the derived checkpoint name is the same as storage, less any extension, then "_checkpoint.nc" is added
+        The reporter internally tracks what data goes into which file, so its transparent to all other classes
+        In the future, this will be able to take Storage classes as well
+
 
     """
-    def __init__(self, base_storage, open_mode=None, checkpoint_interval=10):
+    def __init__(self, storage, open_mode=None, checkpoint_interval=10, checkpoint_storage_file=None):
         if type(checkpoint_interval) != int:
             raise ValueError("checkpoint_interval must be an integer!")
-        self._storage_file_base = base_storage
-        self._storage_file_checkpoint = base_storage + "_checkpointing.nc"
-        self._storage_file_analysis = base_storage + "_analysis.nc"
+        if checkpoint_storage_file is None:
+            dirname, filename = os.path.split(storage)
+            basename, ext = os.path.splitext(filename)
+            addon = "_checkpoint"
+            checkpoint_storage_file = os.path.join(dirname, basename + addon + ext)
+            logger.debug("Checkpoint file automatically chosen as {}".format(checkpoint_storage_file))
+        self._storage_file_analysis = storage
+        self._storage_file_checkpoint = checkpoint_storage_file
         self._storage_checkpoint = None
         self._storage_analysis = None
         self._checkpoint_interval = checkpoint_interval
@@ -101,19 +113,40 @@ class Reporter(object):
 
     @property
     def _storage(self):
-        """Return an iterable of the storage objects, avoids writing things like "for Z in [X,Y]" calls everywhere"""
-        return self._storage_checkpoint, self._storage_analysis
+        """Return an iterable of the storage objects, avoids having the [list, of, storage, objects] everywhere"""
+        return self._storage_analysis, self._storage_checkpoint
+
+    @property
+    def _storage_paths(self):
+        """Return an iterable of paths to the storage files"""
+        return self._storage_file_analysis, self._storage_file_checkpoint
 
     @property
     def _storage_dict(self):
         """Return an iterable dictionary of the self._storage_X objects"""
         return {'checkpoint': self._storage_checkpoint, 'analysis': self._storage_analysis}
 
+    def storage_files_exist(self):
+        """
+        Check if the storage files exist on disk. Reads information on the primary file to see existence of others
+
+        Returns
+        -------
+        files_exist : bool
+            If the primary storage file and its related subfiles exist, returns True
+            If the primary file or any subfiles do not exist, returns False
+        """
+
+
     def is_open(self):
         """Return True if the Reporter is ready to read/write."""
-        if self._storage[0] is None or self._storage[1] is None:
-            return False
-        return np.all([self._storage[0].isopen(), self._storage[1].isopen()])
+        open_check_list = []
+        for storage in self._storage:
+            if storage is None:
+                return False
+            else:
+                open_check_list.append(storage.isopen())
+        return np.all(open_check_list)
 
     def open(self, mode='r'):
         """Open the storage file for reading/writing.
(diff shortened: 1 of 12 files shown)


docs/yamlpages/options.rst
```diff
@@ -497,6 +497,28 @@ Timestep of Langevin Dynamics production runs.
...
+
+Specify how frequently checkpoint information should be saved to file relative to iterations. YANK simulations can be
+resumed only from checkpoints, so if something crashes, up to ``checkpoint_interval`` worth of iterations will be lost
+and YANK will resume from the most recent checkpoint.
+
+This option helps control write-to-disk time and file sizes. The fewer times a checkpoint is written, the less of both
+you will get. If you want to write a checkpoint every iteration, set this to ``1``.
+
+Checkpoint information includes things like full coordinates and box vectors, as well as more static information such
+as metadata, simulation options, and serialized thermodynamic states.
```
andrrizzi: This will need update if you end up moving stuff from the checkpoint storage.

Yank/repex.py
```diff
@@ -70,26 +70,50 @@ class Reporter(object):
...
+        of the iteration at which energies is written, hence why it must be greater than or equal to 1.
+        Checkpoint information cannot be written on iterations which where iteration % checkpoint_interval != 0.
+        Attempting to read checkpointing information results in a masked array where only entries which were written
+            are unmasked
 
     """
-    def __init__(self, storage, open_mode=None):
-        self._storage_file_path = storage
-        self._storage = None
+    def __init__(self, base_storage, open_mode=None, checkpoint_interval=10):
```
andrrizzi: I would make the constructor signature something like
```python
def __init__(self, storage, open_mode=None, checkpoint_interval=10, checkpoint_storage=None):
```
instead, and if `checkpoint_storage is None`, it will take the `storage` path and add a `_checkpoint` suffix to the name (plus append the correct extension). This way:
- We'd be able to pass a `Storage` objects to the constructor in the future in alternative to paths.
- We'd be able in the future to store positions and all the other informations in different formats by passing two different `Storage` objects or file paths with different extensions.

If a user wants to have a different name for the checkpoint than the default (probably a rare use case), he would pass directly a `Reporter` object instead of a file path. Something like:
```python
reporter = Reporter(storage='complex.nc', checkpoint_interval=50, checkpoint_storage='some_other_name.nc')
alchemical_phase.create(..., storage=reporter)
```

jchodera:
This sounds great!

How does resuming work? Does it back up to the last good checkpoint and *erase* the data from the non-checkpoint file? What would happen if, for example, `checkpoint_interval=100`, we fail at iteration 50 into the cycle, resume the simulation, and then terminate again before we reach iteration 100 in the cycle, and then try to analyze the data? Would we see a mix of old and new data?

Lnaden:
Whenever you resume, the file will always back up to the last good iteration it can, it then treats the rest of the data as garbage and overwrites it on the fly

Lnaden:
I'll double check that is what happens though

Lnaden:
Actually, the only way you would get junk data is to stop a simulation, resume it, stop it again before it makes it to the next checkpoint, then try to analyze that data

Lnaden:
Question: Do we want to allow the analysis file to move around (and be used for analysis) w/o the checkpoint file, or should they always be together?

The current implementation requires both files to be together. I can refactor it to allow the files to be separate for analysis (not resume),  but it will take some time.

jchodera:
It would be great if we could *analyze* the smaller analysis file without needing the checkpoint file.

andrrizzi:
I agree.

Lnaden:
Okay. I have re-refactored the code such that the analysis file can operate separate from the checkpoint file. Basic logic is:
```
Open analysis file
If write 
    Open new file for checkpoint
elif append or read 
    Try open checkpoint file
    Warn if no file
Functions acting on checkpoint file fail on their own
```

I still need to change the checkpoint behavior logic for resume, but what is here now will need looked over to make sure I didn't break something else

Looking over issues: Fixes #41

Lnaden:
Okay, I have added the ability for the reporter to ignore "junk" data, but still be properly indexed and sliced. I think this is probably ready for proper review again, @andrrizzi or @jchodera. 

What this does not do is allow the Reporter or Storage classes to be used as as part of the `create()` call. That might need to be done in a separate PR as this one is pretty feature rich already.

jchodera:
Awesome, thanks!

andrrizzi:
Whops! Somehow some of my comments got address to a line that changed. Sorry about that, they still refer to the last version of the code.

Yank/repex.py
```diff
@@ -159,21 +251,57 @@ def open(self, mode='r'):
...
+                nc_last_good_iter[0] = 0
+                return True
+            else:
+                return False
+
+        # Setup the primary file if needed
+        # Generate a unique UUID in case we need to assign it
+        # primary and subfiles are linked by UUID
+        # Uses uuid4 to avoid assigning hostname information
+        primary_uuid = str(uuid.uuid4())
```
Lnaden: I can't quite do that since the checkpoint file can move around and possibly be replaced. The UUID is really tied the primary file and so it can be safely written to as is. However, this takes into account the possibility of subfiles being left behind and the primary file being moved elsewhere post-creation. At that point, they would need the uuid from the existing file and not one created at run time, so I had to split the operation up into separate loops of "check the primary file" then "check the subfiles."

For the case of the checkpoint file, it could be reduced to that, but I'm thinking in case we use other subfiles later.

Yank/repex.py
```diff
@@ -159,21 +251,57 @@ def open(self, mode='r'):
...
+                # Assign the UUID to the subfile
+                ncfile.UUID = primary_uuid
             setattr(self, '_storage_' + nc_name, ncfile)
-        on_file_interval = self._storage_checkpoint.checkpoint_interval
-        if on_file_interval != self._checkpoint_interval:
-            logger.debug("checkpoint_interval != on-file checkpoint interval! "
-                         "Using on file checkpoint interval of {}.".format(on_file_interval))
-            self._checkpoint_interval = on_file_interval
+
+        if self._storage_analysis is not None:
```
Lnaden: No, it could earlier when I had it check the checkpoint file, but I realized that we may need the checkpoint information for non-checkpoint file operations and changed the target file, I'll remove it

Yank/repex.py
```diff
@@ -680,30 +806,29 @@ def write_timestamp(self, iteration):
...
         -------
         data : dict
             The restored data as a dict.
 
         """
+        # Leaving the skeleton to extend this in for now
+        if name in ['metadata']:  # store data needed for analysis in the primary file
+            storage = 'analysis'
+        else:
+            storage = 'checkpoint'
```
Lnaden: MCMC moves are not read by the analysis code
Acceptance rates are stored in their own variables (from `read_mixing_statistics`
```
n_accepted_matrix = self._storage_analysis.variables['accepted'][iteration, :, :].astype(np.int64)
        n_proposed_matrix = self._storage_analysis.variables['proposed'][iteration, :, :].astype(np.int64)```

Temperature is a problem. My big concern is that I don't like storing the massive serialized state strings in what is suppose to be the small analysis file, but if thats the only way we get Temperature, then I guess we'll have to

andrrizzi: Sorry if I wasn't clear, what I mean is that many individual `MCMCMove` objects store their own `n_accepted` and `n_proposed` variables (e.g. `GHMCMove`, `MCRotationMove`) that I'm sure we'll want to read from the analysis soon in the future.

Lnaden:
I added the feature where `Repex` and all the `create()` methods can now take a `Reporter` class object as well as a string. I removed all the passing of the `checkpoint_interval` and `checkpoint_storage_file` from being passed between all the classes. YAMLBuild now also constructs the Reporter object and passes it to the file.

Added a method on the Reporter to get the primary file path so it can be used in strings and what not.

@andrrizzi could you take a quick look over this feature if you have time? The tests should all pass.

Lnaden:
I also think this should be the end of this PR and most of what we wanted to do with the Reporter class for now.

Advances the goals laid out in #535

Lnaden:
I also reduced the number of states of the "test `yank script...` command to speed up the test, its a major choke point of the tests and sometimes causes travis to time out. Of course, that may be due to the setup as well

andrrizzi:
Looks good!

Lnaden:
Woo!

PR #677, opened 2017-05-11
Support SMILES files with comments and compress dictionaries in NetCDF file

andrrizzi:
As the title says.

The CSV standard doesn't have a definition of comment, but FreeSolv uses Python-style comments so I've added support for those.

### DIFF
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -907,7 +907,7 @@ class Reporter(object):
         try:
             nc_variable = self._resolve_variable_path(name, storage)
         except KeyError:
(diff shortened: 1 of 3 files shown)


andrrizzi:
Tests pass! This is ready for review.

jchodera:
Looks good, but do we actually have any tests of the CSV reader functionality in the yank repo?

andrrizzi:
Yep! https://github.com/choderalab/yank/blob/master/Yank/tests/test_yaml.py#L1082

PR #680, opened 2017-05-18
Fix #679: Optimization of the computation of the energy matrix

### LINKED ISSUE #679: When computing energy matrix get new context only if state is incompatible
Currently in [`ReplicaExchange._compute_replica_energies()`](https://github.com/choderalab/yank/blob/master/Yank/repex.py#L1920-L1944) we query the `ContextCache` for a compatible context and we set the positions every time.

A very simple optimization could be to use `state.is_state_compatible(previous_state)` to decide whether to do that (if incompatible) or do `current_therm_state.apply_to_context(context)` instead (if compatible).

andrrizzi:
See issue #679. Ready for review.

### DIFF
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -1930,12 +1930,20 @@ class ReplicaExchange(object):
         for energies, states in [(energy_thermodynamic_states, self._thermodynamic_states),
                                  (energy_unsampled_states, self._unsampled_states)]:
             for i, state in enumerate(states):
-                # Get the context, any Integrator works.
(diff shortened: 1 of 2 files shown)


jchodera:
Nice! Any idea what the performance impact is?

andrrizzi:
Not yet, but I'll test it soon! It shouldn't make things worse though since I think setting the positions is more expensive than setting context global parameters.

andrrizzi:
This is about 3 times faster than the old one for a solvated small molecule! For a protein in explicit solvent the difference is probably bigger.

PR #682, opened 2017-05-22
Update website docs

andrrizzi:
This changes the name of `switch_experiment/phase_every` options to `switch_experiment/phase_interval` as we discussed few weeks ago, and updates the website docs with the recent changes to the YAML syntax.

I don't think I'll have time to find a solution for `extend_simulation` before the next release. Should I comment it out from the docs until it's reintroduced?

### DIFF
--- a/Yank/tests/test_yaml.py
+++ b/Yank/tests/test_yaml.py
@@ -251,8 +251,8 @@ def test_yaml_parsing():
         experiments_dir: /path/to/output/experiments/
         platform: CPU
         precision: mixed
-        switch_experiment_every: -2.0
-        switch_phase_every: 32
+        switch_experiment_interval: -2.0
+        switch_phase_interval: 32
         temperature: 300*kelvin
         pressure: null
         constraints: AllBonds
--- a/docs/yamlpages/options.rst
+++ b/docs/yamlpages/options.rst
@@ -136,6 +136,7 @@ value (``auto``) is equivalent to ``mixed`` when the device support this precisi
 
 Valid options: [auto]/double/mixed/single
 
+
 .. _yaml_options_switch_experiment_interval:
 
 switch_experiment_interval
@@ -150,12 +151,13 @@ When running multiple experiments using the ``!Combinatorial`` tag, this allows 
 completing the specified ``number_of_iterations``. If 0 or less, YANK will complete the combinatorial calculations
 sequentially.
 
-Valid options (0): <int>
+Valid options (0): <Integer>
+
 
 .. _yaml_options_sys_and_sim_prep:
 
-System and Simulation Prepartion:
-=================================
+System and Simulation Preparation:
+==================================
 
 .. _yaml_options_randomize_ligand:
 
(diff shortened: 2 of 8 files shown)


Lnaden:
Yes, comment that out for now (leave in docs, but don't render). @jchodera was right, if the simulation can and is switching every so often, the extend simulation thing is less important. We can put that on the back burner for a while

andrrizzi:
Done! I hope I got the multiline comment syntax right.

Lnaden:
I'll take a look right now

Yank/tests/test_yaml.py
```diff
@@ -251,8 +251,8 @@ def test_yaml_parsing():
         experiments_dir: /path/to/output/experiments/
         platform: CPU
         precision: mixed
-        switch_experiment_every: -2.0
-        switch_phase_every: 32
+        switch_experiment_interval: -2.0
```
Lnaden: Why is this negative?

docs/yamlpages/options.rst
```diff
@@ -132,7 +136,21 @@ value (``auto``) is equivalent to ``mixed`` when the device support this precisi
...
+switch_experiment_interval
+--------------------------
+.. code-block:: yaml
+
+   options:
+     switch_experiments_interval: 0
+
+When running multiple experiments using the ``!Combinatorial`` tag, this allows to switch between experiments every
+``switch_experiments_interval`` iterations, and gather data about multiple molecules/conditions before
+completing the specified ``number_of_iterations``. If 0 or less, YANK will complete the combinatorial calculations
```
Lnaden: I'm not a big fan of the "0 *or less*" being explicitly documented. I would say just "0" and leave the code to handle less in case someone messes up. Otherwise they have to think about "What does a negative interval mean?"

Lnaden:
I found a few places I think need a bit more documentation, I'll fix them

Lnaden:
Okay, I have added docs on the switching in the running section, fixed some pointers that were missing from the `index.rst` for the YAML pages, and removed pointers which no longer existed for on the same page. Let me know what you think @andrrizzi

Yank/tests/test_yaml.py
andrrizzi: This is just a test to make sure that the parsing doesn't blow up if we give a negative number.

docs/yamlpages/options.rst
andrrizzi: Ok! I'll change that.

andrrizzi:
Thanks! I didn't notice that part of the docs. Looks good. I'll merge after addressing your comment.

Yank/tests/test_yaml.py
Lnaden: That is a good idea actually. Keep this in the test then, just change the public documentation then. I'll switch this to approve then

Lnaden:
I apparently found a few missing commits

andrrizzi:
Thanks!

PR #684, opened 2017-05-25
Speed up write Thermodynamic States

Lnaden:
This is an optimization to the `Reporter.write_thermodynamic_states` function to only serialize the `openmmtools.ThermodynamicState._standard_state` object if we have a new, incompatible `ThermodynamicState` with the states we already serialized. 

This results in about a 20% speedup to start time as a lower bound. The speed improvement is a function of both system size and number of states, although there does not appear to be a direct correlation as the `write_dict` and `serialize` functions become the rate-limiting steps.

This will also not pass tests until OpenMMTools 0.9.5 goes live

cc @andrrizzi for review

### DIFF
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -410,57 +410,28 @@ class Reporter(object):
                 for compare_state in stored_states:
                     if compare_state.is_state_compatible(thermodynamic_state):
                         serialized_thermodynamic_state = mmtools.utils.serialize(thermodynamic_state, skip_system=True)
+                        standard_system_hash = stored_states[compare_state]
                         reference_state_name, len_serialization = compatible_states_hashes[standard_system_hash]
                         serialized_thermodynamic_state['_Reporter__compatible_state'] = reference_state_name
                         found_compatible_state = True
                         break
+                # If no compatible state is found, do full serialization
                 if not found_compatible_state:
                     serialized_state = mmtools.utils.serialize(thermodynamic_state)
-                    serialized_thermodynamic_state = serialized_state
+                    serialized_thermodynamic_state = serialized_state['thermodynamic_state']
                     serialized_standard_system = serialized_thermodynamic_state['standard_system']
                     standard_system_hash = serialized_standard_system.__hash__()
                     reference_state_name = '{}/{}'.format(state_type, state_id)
                     len_serialization = len(serialized_standard_system)
+                    # Store new compatibility data
                     compatible_states_hashes[standard_system_hash] = reference_state_name, len_serialization
+                    stored_states[thermodynamic_state] = standard_system_hash
 
                     logger.debug("Serialized state {} is  {}B | {:.3f}KB | {:.3f}MB".format(
                         reference_state_name, len_serialization, len_serialization/1024.0,
                         len_serialization/1024.0/1024.0))
                 # Finally write the dictionary
                 self.write_dict('{}/state{}'.format(state_type, state_id), serialized_thermodynamic_state)
-
-                '''
-                serialized_state = mmtools.utils.serialize(thermodynamic_state)
-
-                # Find the serialized standard system.
-                serialized_thermodynamic_state = serialized_state
-                while 'thermodynamic_state' in serialized_thermodynamic_state:
-                    # The while loop is necessary for nested CompoundThermodynamicStates.
-                    serialized_thermodynamic_state = serialized_state['thermodynamic_state']
-                serialized_standard_system = serialized_thermodynamic_state['standard_system']
-
-                # We store the full standard system serialization only if we haven't
-                # store it yet, otherwise we just write a reference to a state containing
-                # the full system. We normally expect all the ThermodynamicStates to be
-                # compatible, which means we'll store only a single system in most cases.
-                standard_system_hash = serialized_standard_system.__hash__()
-                try:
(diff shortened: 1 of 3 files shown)


jchodera:
This doesn't require changing the reading of thermodynamic states too?

Yank/repex.py
```diff
@@ -401,39 +401,37 @@ def write_thermodynamic_states(self, thermodynamic_states, unsampled_states):
...
+                        serialized_thermodynamic_state = mmtools.utils.serialize(thermodynamic_state, skip_system=True)
+                        standard_system_hash = stored_states[compare_state]
+                        reference_state_name, len_serialization = compatible_states_hashes[standard_system_hash]
+                        serialized_thermodynamic_state['_Reporter__compatible_state'] = reference_state_name
+                        found_compatible_state = True
+                        break
+                # If no compatible state is found, do full serialization
+                if not found_compatible_state:
+                    serialized_state = mmtools.utils.serialize(thermodynamic_state)
                     serialized_thermodynamic_state = serialized_state['thermodynamic_state']
```
andrrizzi: I think this will raise `KeyError` for normal `ThermodynamicState`s (only `CompoundThermodynamicState`'s serialization has the `thermodynamic_state` key.

Yank/repex.py
```diff
@@ -401,39 +401,37 @@ def write_thermodynamic_states(self, thermodynamic_states, unsampled_states):
...
-                while 'thermodynamic_state' in serialized_thermodynamic_state:
-                    # The while loop is necessary for nested CompoundThermodynamicStates.
+                # Check if any compatible state has been found
+                found_compatible_state = False
+                for compare_state in stored_states:
+                    if compare_state.is_state_compatible(thermodynamic_state):
+                        serialized_thermodynamic_state = mmtools.utils.serialize(thermodynamic_state, skip_system=True)
+                        standard_system_hash = stored_states[compare_state]
+                        reference_state_name, len_serialization = compatible_states_hashes[standard_system_hash]
+                        serialized_thermodynamic_state['_Reporter__compatible_state'] = reference_state_name
```
andrrizzi: I think you'll have to unnest the `ThermodynamicState` before adding this field or the current implementation of `read_thermodynamic_states` won't work.

Simply adding `_Reporter__compatible_state` should work, but you could also decide to pop the `standard_system` key from the thermodynamic state serialization to avoid serializing a bunch of `standard_system: None` in the dictionary. I don't have a strong preference here though.

Yank/repex.py
```diff
@@ -401,39 +401,37 @@ def write_thermodynamic_states(self, thermodynamic_states, unsampled_states):
...
                         reference_state_name, len_serialization, len_serialization/1024.0,
                         len_serialization/1024.0/1024.0))
-                else:
-                    serialized_thermodynamic_state.pop('standard_system')
-                    serialized_thermodynamic_state['_Reporter__compatible_state'] = reference_state_name
-
-                # Write state as dictionary.
-                self.write_dict('{}/state{}'.format(state_type, state_id), serialized_state)
+                # Finally write the dictionary
+                self.write_dict('{}/state{}'.format(state_type, state_id), serialized_thermodynamic_state)
```
andrrizzi: I think you want to serialize `serialized_state` here, not `serialized_thermodynamic_state`.

devtools/conda-recipe/meta.yaml
```diff
@@ -23,7 +23,7 @@ requirements:
     - pymbar
     - ambermini >=16.16.0
     - docopt
-    - openmoltools
+    - openmoltools >=0.9.5
```
andrrizzi: In the end we went for `0.10.0` because the release breaks the API.

andrrizzi:
Unfortunately tests here won't pass anyway because in `openmmtools-0.10.0` `AlchemicalFactory` has been renamed, and that's fixed in #685 . If `test_repex.py:TestReporter` pass locally it should be fine, but let me know if you want me to merge that branch in here instead that in `master`.

Lnaden:
@andrrizzi Okay, I made the changes requested. I switched out a few of the variable names to try and make things less confusing* so if you could double check this again to make sure I did not break anything new.

*i.e. `thermodynamic_state` from loop -> serialized to `serialized state` which may or may not have a "thermodynamic_state" dict key which is then extracted to `serialized_thermodynamic_state`. 

Changed this to `state` in the loop -> `serialized_state` of which the `serialized_thermodynamic_state` is extracted (and may be the same object, but thats okay)

Lnaden:
> This doesn't require changing the reading of thermodynamic states too?

@jchodera no, because this does not actually change what is written, just reduces the number of expensive calls to serialize. Instead of serializing every system to string, then figuring out what is is unique and dropping the string, it figures out what is unique first, then only serializes the system as needed.

andrrizzi:
The two problems I found were that `unnest_thermodynamic_state` was returning the wrong variable taken from the outer scope, and that the hack I implemented in `read_thermodynamic_states` didn't work anymore with the way things are handled in `openmmtools` now.

I've also removed the storing of `len_serialization` and the need to compute the hash at all in `write_thermodynamic_states` which should give a slight speed up. I've also modified `read_thermodynamic_states` so that we'll avoid serializing a system every time we use another state as reference, which should give us a great speed up during restore.

Tests pass locally now. @Lnaden could you double check the modifications before merging?

PR #691, opened 2017-06-02
Automatic expanded cutoff and restraint parameters in YAML

### LINKED ISSUE #507: Add YAML Options to pass to the generalized restraint code
Follow up to #348

### LINKED ISSUE #690: Anisotropic dispersion cutoff: automatic determination(?) and support for non-rectangular box
Until #336 is implemented, we have no way to specify through the YAML script a different dispersion cutoff for the complex and solvent phase. This can be problematic when starting from system files prepared by other groups. Should we allow something like
```yaml
options:
  anisotropic_dispersion_cutoff: auto
```
that pushes the cutoff as much as it can based on the initial box_vectors (to a maximum of 16A)?

Also, currently we check that the given cutoff won't cause problems with
```python
box_vectors = system.getDefaultPeriodicBoxVectors()
min_box_dimension = min([max(vector) for vector in box_vectors])
if min_box_dimension < 2.25 * expanded_cutoff_distance:
  raise Exception
```
but I don't think this way of determining `min_box_dimension` works for non-rectangular boxes. I believe we should instead determine the minimum norm of the box vectors.

andrrizzi:
This is ready for review.

- Fix #690: Automatic determination of the expanded cutoff. Support for triclinic box.
- Fix #507: Support passing restraint parameters through YAML script.
- Updated main documentation.

### DIFF
--- a/Yank/yank.py
+++ b/Yank/yank.py
@@ -756,40 +756,53 @@ class AlchemicalPhase(object):
     @staticmethod
     def _expand_state_cutoff(thermodynamic_state, expanded_cutoff_distance):
         """Expand the thermodynamic state cutoff to the given one."""
+        # If we use a barostat we leave more room for volume fluctuations or
+        # we risk fatal errors. This is how much we allow the box size to change.
+        fluctuation_size = 0.8
+
         # Do not modify passed thermodynamic state.
         thermodynamic_state = copy.deepcopy(thermodynamic_state)
         system = thermodynamic_state.system
 
-        # Determine minimum box side dimension. This assumes
-        # that box vectors are aligned with the axes.
+        # Determine minimum box side dimension.
         box_vectors = system.getDefaultPeriodicBoxVectors()
-        min_box_dimension = min([max(vector) for vector in box_vectors])
+        min_box_dimension = min([np.linalg.norm(vector) for vector in box_vectors])
 
-        # If we use a barostat we leave more room for volume fluctuations or
-        # we risk fatal errors. If we don't use a barostat, OpenMM will raise
-        # the appropriate exception on context creation.
-        if (thermodynamic_state.pressure is not None and
-                min_box_dimension < 2.25 * expanded_cutoff_distance):
+        # Determine cutoff automatically if requested.
+        # We leave more space if the volume fluctuates.
+        if expanded_cutoff_distance == 'auto':
+            if thermodynamic_state.pressure is None:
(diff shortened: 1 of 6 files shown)


docs/yamlpages/options.rst
```diff
@@ -294,12 +294,13 @@ anisotropic_dispersion_cutoff
 .. code-block:: yaml
 
    options:
-     anisotropic_dispersion_cutoff: 16.0 * angstrom
+     anisotropic_dispersion_cutoff: auto
 
-Specify the expanded cutoff distance for YANK's :ref:`yaml_options_anisotropic_dispersion_correction` setting.
-Please see the main :ref:`yaml_options_anisotropic_dispersion_correction` option for details/
+Specify the expanded cutoff distance for YANK's :ref:`yaml_options_anisotropic_dispersion_correction` setting. If
+``auto`` the cutoff will be set to a little smaller than the minimum box size. Please see the main
```
jchodera: We should document `auto` more precisely. Suggest changing 
> the cutoff will be set to a little smaller than the minimum box size.
to
> the cutoff will be set to `0.99*min_box_vector/2` if no barostat is in use or `0.8*min_box_vector/2` if one is in use (to account for box size fluctuations), with `min_box_vector` denoting the smallest OpenMM box vector defining the initial triclinic cell volume.

Yank/yank.py
```diff
@@ -756,40 +756,53 @@ def extend(self, n_iterations):
...
+                    pass
+            else:
                 # We don't want to reduce the cutoff if it's already large.
-                if force.getCutoffDistance() < expanded_cutoff_distance:
+                if force_cutoff < expanded_cutoff_distance:
                     force.setCutoffDistance(expanded_cutoff_distance)
 
                     # Set switch distance. We don't need to check if we are
                     # using a switch since there is a setting for that.
                     switching_distance = expanded_cutoff_distance - 1.0*unit.angstrom
```
jchodera: Let's move the `1.0*unit.angstrom` up to a `switch_width` variable defined above just so we can more clearly see where this is defined. I don't think we need to expose this as a YAML argument, but it will make it easier to do if we need to.

andrrizzi:
Thanks! I'll address the comments before merging.

andrrizzi:
Actually, would it make sense to just preserve the same `switch_width` in the original system instead of specifying it as a parameter?

jchodera:
Great idea!

Issue #697, opened 2017-06-12
Nightly travis tests are failing


jchodera:
They appear to be [timing out on some tests](https://travis-ci.org/choderalab/yank/jobs/241806897#L1320).

Any idea how we can fix this? Should we exclude additional slow tests?

andrrizzi:
I think tests are taking longer because now we [compute and store the energy](https://github.com/choderalab/yank/blob/master/Yank/repex.py#L1593-L1597) at iteration 0. I can probably bring that calculation inside the while loop [here](https://github.com/choderalab/yank/blob/master/Yank/repex.py#L1609).

The easiest way to make that test pass is probably to test a solvation free energy instead of a receptor-ligand system (it's just testing that the command is running anyway).

PR #698, opened 2017-06-12
Fix maximum allowed cutoff of expanded states.

andrrizzi:
Should be ready for review.

The theoretical maximal allowed cutoff is given by half the norm of the smallest vector, but OpenMM limits it to the minimum diagonal element of the box vector matrix for efficiency.

Related to PR #691 and issue #690. See also pandegroup/openmm#1831.


andrrizzi:
I've also changed the test of the CLI option override to fix #697. The integration was making Travis timing out. Tests pass now.

jchodera:
Awesome, thanks!

Issue #582, opened 2016-11-25
Brainstorming ideas for speeding up YANK


jchodera:
I'm just collecting some ideas to consider after the 1.0 release.

* Optimize code based on profiling 
* Non-cubic/rectangular boxes (truncated octahedron, dodecahedron)
* Tune nonbonded cutoffs to balance performance and overlap
* Optimize PME parameters to maximize performance given error threshold
* Adjust number of steps per alchemical state to minimize discrepancy between end times for propagation steps
* Have separate thread write data to disk
* Only write full system positions checkpoints to disk every N iterations; only resume from these points
* Use better integrators that permit larger timestep, such as g-BAOAB
* Speed up ligand Monte Carlo using a CustomIntegrator or applying it only to weakly interacting replicas 
* optimize alchemical pathway to minimize number of thermodynamic states

jlerche:
Re: separate thread writing to disk. Just need to be aware that due to the global interpreter lock threads created with the `threading` library aren't actually run in a separate os process, the `multiprocessing` library would have to be used to spawn a separate process, with the overhead of loading a copy of the yank python files into memory.

jchodera:
Thanks! I've used MPI processes in the past to do this kind of thing with Python, but there are now plenty of options for truly parallel processing that avoids the GIL.

jchodera:
@Lnaden has implemented infrequent system checkpoints in #675!

jchodera:
Switching to g-BAOAB with hydrogen mass repartitioning is probably our easiest win in terms of getting a big speed boost for little effort. Once @maxentile's assessment of HMR is complete in a couple of weeks, we'll have an idea of what the optimal settings are here.

PR #704, opened 2017-06-27
Speedup and reduce size of storage [DNM]

### LINKED ISSUE #702: Fix netcdf string compression
Just opening an issue to record my experiments for reference.

With the current way we save strings, the `zlib` compression doesn't work. I was able to make it work by saving strings the old way as a character array.

```python
str_out = netcdf.stringtochar(np.array(python_string, 'S4'))

dimension = len(str_out)
dimension_name = 'string' + str(dimension)
if dimension_name not in self._storage_dict[storage].dimensions:
    self._storage_dict[storage].createDimension(dimension_name, dimension)

nc_variable = self._storage_dict[storage].createVariable(name, 'S1', (dim_name,), zlib=True)

nc_variable[:] = str_out
```
With this code the resulting nc file that contained the serialization of alchemically modified Src got reduced from 52M to 32K!!

I didn't test the restoring so we may have to tweak it a little. Also, with this implementation it may be impossible to update strings since I'm not sure you can change the dimension of a variable after its creation. The point is that maybe netcdf doesn't support the compression of the new variable-length `str` type variables.

Worse case scenario, we could save permanently the expensive bits (thermodynamic states, topology, ...), and save as vlen strings the things that need to be updated (i.e. mcmcmoves statistics).

Lnaden:
This PR tries to accomplish 2 things:
* Reduce the file size of the storage files by compressing the serialized system strings
* Reduce the boot up time of YANK by cutting down on how much processing of the system strings is done.

This PR does that, but currently has some problems (**D**o **N**ot **M**erge yet!)

The serialized systems are not processed through NetCDF anymore, and instead are stored in GZIP'd picked objects as an extra storage file. I mange to get 120MB of strings down to 5.6 MB in another file. It decompresses lossless.

There is speed up since the serialized systems are no longer run through the YAML dumper which appears to recursively search through the string, so that is a huge time save (no quantitative measurement, but it is very noticeable for me).

There is still something being saved to the main analysis file which is occupying a large amount of space. For instance, the freshly initialized explicit solvent simulation, the analysis file is still 32 MB. At least its not 200MB, but that number still isn't as small as it should be. Some other object is taking up a large amount of space.

This is not ready to merge as the tests are not passing (due to a change in how the file is saved from the test). And I don't know why the file is still so large.

@andrrizzi or @jchodera if you could look over this before I spend too much time trying to fix the rest of the problems and debugging what else is happening that takes up so much space.

Fixes #702 
Progress towards #582 and #41

### DIFF
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -204,7 +204,10 @@ class Reporter(object):
             if storage is None:
                 return False
             else:
-                open_check_list.append(storage.isopen())
+                try:
+                    open_check_list.append(storage.isopen())
+                except AttributeError:
+                    open_check_list.append(not storage.closed())
         return np.all(open_check_list)
 
     def open(self, mode='r'):
@@ -937,8 +940,12 @@ class Reporter(object):
         """
         # Create variable if needed.
         for storage in self._storage:
-            if 'timestamp' not in storage.variables:
-                storage.createVariable('timestamp', str, ('iteration',), zlib=False, chunksizes=(1,))
+            try:
+                if 'timestamp' not in storage.variables:
+                    storage.createVariable('timestamp', str, ('iteration',), zlib=False, chunksizes=(1,))
+            except AttributeError:
+                # Don't worry about timestamp on non-nc files
+                pass
         timestamp = time.ctime()
         self._storage_analysis.variables['timestamp'][iteration] = timestamp
         checkpoint_iteration = self._calculate_checkpoint_iteration(iteration)
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -333,9 +333,9 @@ class Reporter(object):
         # Handle Read mode
         if mode in ['r', 'a']:
             try:
-                self._storage_systems = gzip.open(self._storage_file_systems, 'r')
-                UUID = pickle.load(self._storage_systems)
-                assert UUID == primary_uuid
+                with gzip.open(self._storage_file_systems, 'r') as self._storage_systems:
+                    UUID = pickle.load(self._storage_systems)
+                    assert UUID == primary_uuid
             except IOError:  # Trap not on disk
                 logger.debug('Could not locate systems subfile. This is okay for certain append and read operations, '
                              'but not for production simulation!')
@@ -345,9 +345,9 @@ class Reporter(object):
                               "Analysis UUID: {}"
                               "Systems UUID: {}".format(primary_uuid, UUID))
         else:
-            self._storage_systems = gzip.open(self._storage_file_systems, mode)
-            # Write UUID to file as the first entry
-            pickle.dump(primary_uuid, self._storage_systems)
+            with gzip.open(self._storage_file_systems, mode) as self._storage_systems:
+                # Write UUID to file as the first entry
+                pickle.dump(primary_uuid, self._storage_systems)
         self._open_mode = mode
 
     def close(self):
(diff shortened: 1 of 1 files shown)


jchodera:
All variables are initialized to their initial chunksize when the file is initially created. Could that explain the 32MB?

Lnaden:
Maybe. But each of the itterable dimensions auto size as we add iterations. so the only thing that `should` be occupying initial space is the energies, and even that isnt that large. I'd bet its another uncompressed string somewhere somehow (like repex options/metadata or something), possibly the reference system?

Lnaden:
And if it is the reference system, we can just force that into a netcdf fixed length char dim and let the netcdf compress that

jchodera:
> Maybe. But each of the itterable dimensions auto size as we add iterations. so the only thing that should be occupying initial space is the energies, and even that isnt that large

One chunk worth of *every* variable is allocated initially. Are you saying no variables but the energies are created when the file is created initially?

andrrizzi:
Just reporting this here as well for the records. I run some `timeit` on my laptop of

1)
```python
res = pickle.dumps({'system': alch_ser})
```
2)
```python
res = yaml.dump({'system': alch_ser})
```
and 3)
```python
byte_alch_ser = alch_ser.encode('utf-8')
compressed = zlib.compress(byte_alch_ser, 4)
res = yaml.dump({'system': compressed})
```
here is the result and the dimension of the `res` string.
1) 0.016s, 38.4MB
2) 65.13s, 50.5MB
3) 4.759s, 3.1MB

For the 32MB, I guess both the `reference_state` and the `topography` (which contains the `Topology` object), which are saved as `metadata`, are pretty big.

For the state serialization of the states, we could add the first two lines of solution 3) directly into `ThermodynamicState.__getstate__(..., zlib=False)` through an extra parameter (similarly to what we did with `skip_system`) to make it available to everyone in the group. Since it's so fast, we could even increase the compression level (4 in this example) and reduce the compressed string dimension some more, or make the `zlib` argument an integer indicating the compression level instead of a boolean. This has also the advantage that it'll be already compatible with the `storage` layer in `openmmtools`.

For the topology, I don't have an idea on the top of my head since neither MDtraj nor OpenMM support string serialization, and we currently have to go through a dataframe representation. It might be good to check how much space it takes on disk and how much `yaml.dump(topography_serialization)` takes.

andrrizzi:
Worth mentioning also that I `netcdf` probably adds few MB of overhead. During my experiments with compression, I noticed that the `nc` file size was always few MBs bigger than the string I was saving.

andrrizzi:
Although now that I remember, with a fixed length string and same level of `zlib` compression, the final dimension on disk was much smaller, so that overhead could affect only variable length strings.

Lnaden:
@andrrizzi okay, so what do you recommend as far as an approach here. I get the various timings and what not, but I'm not following what your suggestion is.

Lnaden:
Okay. So we cannot store the pre-compressed string directly (improper characters when netcdf tries to read it), but we can do the following, which is really dumb, but lets us keep the framework with very minimal edits:

1. Encode the strings into byte types (Python 3 only)
2. Compress the strings with `zlib`
3. Run the compressed string through the `yaml.dump` to yield a string with the compressed data that 
4. Save the "string" amalgamation to the netcdf variable

On load:
1. Read variable, yaml.load it, decompress, decode.

If that looks like a solution (since it would not break anything), let me know and I'll put it in. It takes a 64MB string down to 4MB, so its not as high of compression as the gzip/pickle in another file approach, but its better

andrrizzi:
To keep the thread synced, I've updated #702 with a snippet that should use both python and netcdf compression.

Lnaden:
That's fine, I may just close this down depending on the choices in that thread

Lnaden:
Closing this down and going with a cleaner approach as per #702

PR #708, opened 2017-07-06
Use C-optimized YAML dumper and loader for Storage

### LINKED ISSUE #706: Consider using CDumper and CLoader for yaml dumping/loading
The optimized-C dumper/loader seem to be about 7-8 times faster than the python-implemented ones. When writing dictionaries in the netcdf storage, we use custom Dumper/Loader to handle tags though, so we'll have to check if it is possible to customize CDumper/CLoader.

Useful links
- Why is pyyaml slower than json: https://stackoverflow.com/questions/27743711/can-i-speedup-yaml
- https://stackoverflow.com/questions/16864132/is-there-an-up-to-date-fast-yaml-parser-with-python-bindings

andrrizzi:
Closes #706 .

Using `CDumper` gives me a x15 speedup on my laptop when storing the serialization of `openmmtools.testsystems.SrcExplicit` with `Reporter.write_dict()`.

I'll wait for the new `openmmtools` to be released before merging to be sure that the recent changes in the two repos are compatible.

### DIFF
--- a/Yank/tests/test_repex.py
+++ b/Yank/tests/test_repex.py
@@ -324,7 +324,7 @@ class TestReporter(object):
                 unsampled_serialized.append(unsampled_dict)
 
             # Two of the three ThermodynamicStates are compatible.
-            assert isinstance(states_serialized[0]['standard_system'], str)
+            assert 'standard_system' in states_serialized[0]
             assert 'standard_system' not in states_serialized[1]
             state_compatible_to_1 = states_serialized[1]['_Reporter__compatible_state']
             assert state_compatible_to_1 == 'thermodynamic_states/0'
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -2257,7 +2257,7 @@ class ParallelTempering(ReplicaExchange):
 # MODULE INTERNAL USAGE UTILITIES
 # ==============================================================================
 
-class _DictYamlLoader(yaml.Loader):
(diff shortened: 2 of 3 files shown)


Yank/tests/test_repex.py
```diff
@@ -324,7 +324,7 @@ def decompact_state_variable(variable):
                 unsampled_serialized.append(unsampled_dict)
 
             # Two of the three ThermodynamicStates are compatible.
-            assert isinstance(states_serialized[0]['standard_system'], str)
+            assert 'standard_system' in states_serialized[0]
```
Lnaden: This test will always be true. This will only check that the key is in the dict. When no system is returned, the `None` object occupies the value, but the `in <dict>` operator is only over the keys.

andrrizzi: Actually we pop `'standard_system'` from the `dict` when we add `'_Reporter__compatible_state'` (see [here](https://github.com/choderalab/yank/blob/master/Yank/repex.py#L421)). I got it wrong the first time.

Lnaden: Ah, yes. then in that case its fine

Lnaden:
The OpenMMTools hotfix is uploaded on the linux builds. Restarting the tests

andrrizzi:
Tests finally pass! Is it ok if I merge this @Lnaden ?

Lnaden:
Punch it!

PR #712, opened 2017-07-10
Online analysis feature

Lnaden:
This PR adds online analysis to the YANK simulations, handled primarily at the sampler level.

* Adds 3 options to the Repex API (and YAML options)
* Includes error and sanity checks common to MBAR without crashing the simulation.
* Stores the last online analysis attempt's MBAR f_k to use as the first guess of the new online analysis computations
* Should be backwards compatible?

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -176,7 +176,7 @@ class YankPhaseAnalyzer(ABC):
     Analyzer works in units of kT unless specifically stated otherwise. To convert back to a unit set, just multiply by
     the .kT property.
     """
-    def __init__(self, reporter, name=None, reference_states=(0, -1)):
+    def __init__(self, reporter, name=None, reference_states=(0, -1), analysis_kwargs=None):
         """
         The reporter provides the hook into how to read the data, all other options control where differences are
         measured from and how each phase interfaces with other phases.
@@ -199,6 +199,9 @@ class YankPhaseAnalyzer(ABC):
                 O[j] is not used
             For O where differences between states are required (e.g. Free Energy):
                 O[i,j] = O[j] - O[i]
+        analysis_kwargs : None or dict, optional
+            Dictionary of extra keyword arguments to pass into the analysis tool, typically MBAR.
+            For instance, the initial guess of weights to give to MBAR would be something like {'f_k':[0,1,2,3]}
         """
         if not reporter.is_open():
             reporter.open(mode='r')
@@ -224,6 +227,9 @@ class YankPhaseAnalyzer(ABC):
         self.reference_states = reference_states
         self._mbar = None
         self._kT = None
+        if analysis_kwargs is not None and type(analysis_kwargs) is not dict:
+            raise ValueError('analysis_kwargs must be either None or a dictionary')
+        self._extra_analysis_kwargs = analysis_kwargs
 
     @property
     def name(self):
@@ -473,8 +479,10 @@ class YankPhaseAnalyzer(ABC):
 
         # Initialize MBAR (computing free energy estimates, which may take a while)
         logger.info("Computing free energy differences...")
-        mbar = MBAR(energy_matrix, samples_per_state)
-
+        if self._extra_analysis_kwargs is not None:
+            mbar = MBAR(energy_matrix, samples_per_state, **self._extra_analysis_kwargs)
+        else:
+            mbar = MBAR(energy_matrix, samples_per_state)
         self._mbar = mbar
 
     def _combine_phases(self, other, operator='+'):
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -227,9 +227,9 @@ class YankPhaseAnalyzer(ABC):
         self.reference_states = reference_states
         self._mbar = None
         self._kT = None
-        if analysis_kwargs is not None and type(analysis_kwargs) is not dict:
+        if type(analysis_kwargs) not in [type(None), dict]:
             raise ValueError('analysis_kwargs must be either None or a dictionary')
-        self._extra_analysis_kwargs = analysis_kwargs
+        self._extra_analysis_kwargs = analysis_kwargs if (analysis_kwargs is not None) else dict()
 
     @property
     def name(self):
@@ -479,10 +479,7 @@ class YankPhaseAnalyzer(ABC):
 
         # Initialize MBAR (computing free energy estimates, which may take a while)
         logger.info("Computing free energy differences...")
-        if self._extra_analysis_kwargs is not None:
-            mbar = MBAR(energy_matrix, samples_per_state, **self._extra_analysis_kwargs)
-        else:
-            mbar = MBAR(energy_matrix, samples_per_state)
+        mbar = MBAR(energy_matrix, samples_per_state, **self._extra_analysis_kwargs)
         self._mbar = mbar
 
     def _combine_phases(self, other, operator='+'):
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -201,7 +201,8 @@ class YankPhaseAnalyzer(ABC):
                 O[i,j] = O[j] - O[i]
         analysis_kwargs : None or dict, optional
             Dictionary of extra keyword arguments to pass into the analysis tool, typically MBAR.
-            For instance, the initial guess of weights to give to MBAR would be something like {'f_k':[0,1,2,3]}
+            For instance, the initial guess of relative free energies to give to MBAR would be something like:
+                {'f_k':[0,1,2,3]}
         """
         if not reporter.is_open():
             reporter.open(mode='r')
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -949,8 +949,9 @@ class Reporter(object):
     def read_mbar_free_energies(self, iteration):
         """
         Read the MBAR dimensionless free energy at a given iteration from file. These free energies are relative to
-        the first state. These are computed through self-consistent iterations from an initial guess. The initial
-        guess is often 0 for all states, so any state not written is returned as zeros.
+        the first state. These are computed through self-consistent iterations from an initial guess.
+        The initial guess is often 0 for all states, so any state not written is returned as zeros for f_k, and
+            infinity for
 
         Used primarily in online analysis, and should be used in tandem with an YankPhaseAnalysis object from the
         analyze module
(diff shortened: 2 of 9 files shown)


Yank/repex.py
```diff
@@ -2099,6 +2215,65 @@ def _attempt_swap(self, replica_i, replica_j):
...
+        try:
+            mbar = analysis.mbar
+            free_energy, err_free_energy = analysis.get_free_energy()
+            output_fe = free_energy[idx, jdx]
+            output_dfe = err_free_energy[idx, jdx]
+            logger.debug("Current Free Energy Estimate is {} +- {} kT".format(output_fe, output_dfe))
+            self._last_mbar_f_k = mbar.f_k
+            if np.isnan(output_dfe):  # Trap a case when errors don't converge (usually due to under sampling)
+                output_dfe = np.inf
+        except ParameterError as e:
```
jchodera: What kind of errors are we expecting here?

Yank/repex.py
```diff
@@ -1102,6 +1145,19 @@ class ReplicaExchange(object):
     replica_mixing_scheme : 'swap-all', 'swap-neighbors' or 'none'
         The scheme used to swap thermodynamic states between replicas
         (default is 'swap-all').
+    online_analysis_interval : None or Int >= 1, optional, default None
+        Choose the interval at which to perform online analysis of the free energy.
+        After every interval, the simulation will be stopped and the free energy estimated. If the error in the free
+            energy estimate is at or below online_analysis_target_error, then the simulation will be considered
+            completed.
+    online_analysis_target_error : float >= 0
```
jchodera: List default.

Yank/repex.py
```diff
@@ -1181,7 +1237,9 @@ class while the simulation is running. This reads the SamplerStates of every
     # Constructors.
     # -------------------------------------------------------------------------
 
-    def __init__(self, mcmc_moves=None, number_of_iterations=1, replica_mixing_scheme='swap-all'):
+    def __init__(self, mcmc_moves=None, number_of_iterations=1, replica_mixing_scheme='swap-all',
+                 online_analysis_interval=None, online_analysis_target_error=1.0,
```
jchodera: This is the target error per *phase* right? If we want errors ~0.2 kcal/mol by default, that's allowing something like 0.2 kT (0.14 kcal/mol) error in each phase.

jchodera: So maybe we should reduce this to 0.2.

Yank/analyze.py
```diff
@@ -199,6 +199,9 @@ def __init__(self, reporter, name=None, reference_states=(0, -1)):
                 O[j] is not used
             For O where differences between states are required (e.g. Free Energy):
                 O[i,j] = O[j] - O[i]
+        analysis_kwargs : None or dict, optional
+            Dictionary of extra keyword arguments to pass into the analysis tool, typically MBAR.
+            For instance, the initial guess of weights to give to MBAR would be something like {'f_k':[0,1,2,3]}
```
Lnaden: You're right, "weights" is imprecise and confusing since the weights are per-sample. I'll update this everywhere

Most the variable names and defaults in general I chose for this could use some feedback, so I'm happy to fix any that dont sound right

Yank/repex.py
Lnaden: I agree. This is per-phase, so the compound error could be very large. I knew that 1kT was going to be too large, but I wanted to get this feature up as a PR for discussion and refine a bit after (like now). 

I'm still wanting to put tests in, but I cant think of how to easily do that since this relies on the `analyze` module which we have no tests for anyways because they will all be slow and require real simulation data.

I have some ideas for a very small, simple dataset which can be generated at test run-time. We would be able to use the API and just feed it really, really simple harmonic oscillator, tens of LJ fluid system, or something, make X time steps at very few states, save and analyze. We would want this instead of a static dataset since we are testing functionality and constantly changing that at the moment.

Yank/repex.py
Lnaden: With poorly overlapping data that is also poorly sampled, MBAR can throw odd numerical issues when it tries to solve itself. There are 2 main ones I know of, the case where the W_kn sums dont add to 1, throwing the parameter error, and the case where the error estimate does not converge and throws either a nan or an inf. The only times I really see this is in very, very poorly overlapped data, which may happen.

Because poor overlapping data (due to poorly sampled data) should not be a crash criteria, I trap it a few times to see if it clears up after a few iterations. I don't let it run on forever though and eventually crash it, reporting the stack of errors it banked along the way.

Yank/repex.py
jchodera: > I'm still wanting to put tests in, but I cant think of how to easily do that since this relies on the analyze module which we have no tests for anyways because they will all be slow and require real simulation data.

What about taking a small system in implicit solvent or vacuum and pre-generating some data for it that gets checked into the repo? You can test (1) offline analysis, and (2) online analysis (to make sure it terminates quickly if the simulation has already converged to the specified tolerance).

Yank/repex.py
jchodera: > There are 2 main ones I know of, the case where the W_kn sums dont add to 1, throwing the parameter error, and the case where the error estimate does not converge and throws either a nan or an inf.

Ugh. We should really make pymbar more robust...

Yank/repex.py
Lnaden: That's also a fine option. We just have not really had the ability to easily make a system like that for YANK tests pre-API. Since we have the API in now, its worth putting in. I'll come up with something for tomorrow and see if it works. I really don't want to upload more than a few megs, and maybe it should not even be distributed with the package on conda, kept on a `dev` side of things GitHub or on its own label

jchodera: > I really don't want to upload more than a few megs, and maybe it should not even be distributed with the package on conda, kept on a dev side of things GitHub or on its own label

If it's a really tiny vacuum system (e.g. [`HostGuestVacuum`](http://openmmtools.readthedocs.io/en/latest/api/generated/openmmtools.testsystems.HostGuestVacuum.html#openmmtools.testsystems.HostGuestVacuum)) it should be very small.

Yank/repex.py
```diff
@@ -1201,8 +1259,43 @@ def __init__(self, mcmc_moves=None, number_of_iterations=1, replica_mixing_schem
...
+
+        # Store all the options
+        self._online_analysis_interval = online_analysis_interval
+        self._online_analysis_target_error = online_analysis_target_error
+        self._online_analysis_minimum_iterations = online_analysis_minimum_iterations
+        self._online_error_trap_counter = 0  # Counter for errors in the online estimate
+        self._online_error_bank = []
+
+        if self._number_of_iterations == np.inf:
+            if self._number_of_iterations == np.inf:
```
andrrizzi: Did you mean to check something else here?

Yank/repex.py
```diff
@@ -1201,8 +1259,43 @@ def __init__(self, mcmc_moves=None, number_of_iterations=1, replica_mixing_schem
...
+        if online_analysis_interval is not None and (
+                        type(online_analysis_interval) != int or online_analysis_interval < 1):
+            raise ValueError("online_analysis_interval must be an integer 1 or greater")
+        if online_analysis_interval is not None:
+            if online_analysis_target_error < 0:
+                raise ValueError("online_analysis_target_error must be a float >= 0")
+            elif online_analysis_target_error == 0:
+                logger.warn("online_analysis_target_error of 0 may never converge.")
+            if type(online_analysis_minimum_iterations) is not int or online_analysis_minimum_iterations < 0:
+                raise ValueError("online_analysis_minimum_iterations must be an integer >= 0")
```
andrrizzi: It may be good to move these checks as part of the relative properties so that they are checked every time they are modified. You could extend the `_StoredProperty` class below to take a `condition` argument of type `function` in `__init__` that check the value and eventually raise the exception. That's a class that helps keeping the options up-to-date with the options in the storage.

Yank/repex.py
```diff
@@ -1201,8 +1259,43 @@ def __init__(self, mcmc_moves=None, number_of_iterations=1, replica_mixing_schem
...
+            if self._number_of_iterations == np.inf:
+                logger.warning("WARNING! You have specified an unlimited number of iterations and a target error "
+                               "for online analysis of 0.0! Your simulation may never reach 'completed' state!")
+            else:
+                logger.warning("WARNING! This simulation will never be considered 'complete' since there is no "
+                               "specified maximum number of iterations nor target uncertainty in free energy estimate!")
+        # Handle online analysis
+        self._online_analysis_interval = None
+
+        self._online_analysis_target_error = None
```
andrrizzi: Typo?

Yank/repex.py
```diff
@@ -1649,6 +1751,20 @@ def run(self, n_iterations=None):
             # Write iteration to storage file.
             self._report_iteration()
 
+            # Compute online free energy
+            if (self._online_analysis_interval is not None and
+                    self._iteration > self._online_analysis_minimum_iterations and
+                    self._iteration % self._online_analysis_interval == 0):
+                free_energy_error = self._run_online_analysis()
+            free_energy_error = None
```
andrrizzi: Typo?

Yank/repex.py
```diff
@@ -1609,6 +1709,8 @@ def run(self, n_iterations=None):
         n_iterations : int, optional
            If specified, only at most the specified number of iterations
            will be run (default is None).
+        continue_past_online_target : bool, optional, default False
```
andrrizzi: Do we need this extra argument? Maybe we can just set `online_analysis_target_error` or `online_analysis_target_error` to `None`, or use some of the existing options anyway?

Yank/repex.py
```diff
@@ -2099,6 +2215,65 @@ def _attempt_swap(self, replica_i, replica_j):
...
+    # Internal-usage: Online Analysis.
+    # -------------------------------------------------------------------------
+
+    @mpi.on_single_node(rank=0, broadcast_result=False, sync_nodes=False)
+    @mpi.delayed_termination
+    @mmtools.utils.with_timer('Computing online free energy estimate')
+    def _run_online_analysis(self):
+        # Prep the reporter
+        self._reporter.close()
+        self._reporter.open('r')
```
andrrizzi: Hmm. I don't understand why we need to do this. Could you elaborate/add some comments?

Yank/yank.py
```diff
@@ -411,6 +411,14 @@ def number_of_iterations(self):
     def number_of_iterations(self, value):
         self._sampler.number_of_iterations = value
 
+    @property
+    def is_complete(self):
+        """bool: is the sampler complete by some other mechanism. If no method is present in sampler, returns False"""
+        try:
+            return self._sampler.is_complete
+        except AttributeError:
+            return False
```
andrrizzi: Alternatively, you could check `self.iteration == self.number_of_iterations` if sampler doesn't expose a `is_complete` attribute.

Yank/repex.py
Lnaden: Hmm. I had it in my head that a dataset open in 'w' mode could not be read from. That was wrong. I have removed those lines

Lnaden:
Made the requested changes.
@andrrizzi can you look at the `_StoredOptions` changes to make sure I did it right

Lnaden:
@andrrizzi okay, tests are passing now. Could you review? I'll get started on the analyze module testing

PR #714, opened 2017-07-12
Add Analyze Tests, Fix MultiPhaseAnalyze

Lnaden:
* Adds a proper set of tests for the analyze module.
* Fixes the previously untested `MultiPhaseAnalyze `code
* Cleans up analyze API docs
* Removes old `analyze.py` legacy code that is no longer used
* Tries to speed up some tests by cutting down the number of states used
* Reduces citation spam, provides internal way to mute citations
    * Very experimental, we can remove this feature if desired.

I have not set this main one up to merge yet, because I want to do this in fragments, so I don't think the tests will run.

@andrrizzi do you think you would have time to look over this? There is quite a lot here. I can also merge the `online-analysis` branch first, then come back and open a new PR to merge this into master after.

### DIFF
--- a/Yank/tests/test_analyze.py
+++ b/Yank/tests/test_analyze.py
@@ -1,0 +1,295 @@
+#!/usr/local/bin/env python
+
+"""
+Test analyze.py facility.
+
+"""
+
+# =============================================================================================
+# GLOBAL IMPORTS
+# =============================================================================================
+
+import os
+import math
+import copy
+import shutil
+import tempfile
+import contextlib
+
+import yaml
+import pymbar
+import numpy as np
+import scipy.integrate
+from simtk import openmm, unit
+from nose.plugins.attrib import attr
+from nose.tools import assert_raises
+
+import openmmtools as mmtools
+from openmmtools import testsystems
+
+from yank import mpi, utils, analyze
+from yank.repex import Reporter, ReplicaExchange, _DictYamlLoader
+
+# ==============================================================================
+# MODULE CONSTANTS
+# ==============================================================================
+
+kB = unit.BOLTZMANN_CONSTANT_kB * unit.AVOGADRO_CONSTANT_NA  # Boltzmann constant
+
+
+# ==============================================================================
+# TEMPLATE PHASE CLASS WITH NO OBSERVABLES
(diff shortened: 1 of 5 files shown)


jchodera:
Awesome!

This is a great opportunity to also give the neglected [analysis module docs](http://getyank.org/0.16.2/analysis.html) some love.

andrrizzi:
Sure! I can take a look at this tomorrow!

Yank/tests/test_analyze.py
```diff
@@ -0,0 +1,295 @@
...
+        cls.hostguest_test = (hostguest_compound_states, hostguest_sampler_states, hostguest_unsampled_states)
+
+        # Run a quick simulation
+        thermodynamic_states, sampler_states, unsampled_states = copy.deepcopy(cls.hostguest_test)
+        n_states = len(thermodynamic_states)
+
+        # Remove one sampler state to verify distribution over states.
+        sampler_states = sampler_states[:-1]
+
+        repex = ReplicaExchange()
```
andrrizzi: Unused?

Lnaden: Oh, yeah, bit of the copy leftover I missed. I'll clean that up

Lnaden:
I suppose I could move a number of those static functions into module functions. I just liked the notion of not really needing anything out of the module directly per-se, and you could do everything by class. But I suppose a fair amount of that might be lost in the MultiPhase anyways.

The one function I am torn on is the `get_timeseries` function which really is a `classmethod` of the implemented `YankPhaseAnalyzer` class, but does not need the class object at all since the only thing it should depend on is the sampler. Right now its just a static method which raises a `NotImplementedError` if you fail to implement, but I dont know how else to classify it

andrrizzi:
As far as I understand, this is an old debate in Python. I tend to follow [these suggestions](https://stackoverflow.com/questions/18262595/python-should-i-use-static-methods-or-top-level-functions) when possible, but we can definitely postpone the decision.

Lnaden:
Ugh, that was an annoying bug to track down.
I'll get this in and work on the analyze docs in a separate PR in case @andrrizzi needs/wants some of the functionality from this

PR #715, opened 2017-07-13
Remove c_rf term in expanded cutoff state with reaction field

### LINKED ISSUE #670: Handling reaction-field correctly
I just realized that our experiments with setting`c_rf = 0` for reaction-field electrostatics were flawed because we still reweight to standard reaction-field (with `c_rf != 0`) for both fully interacting and fully decoupled/annihilated states, which allows the contribution we wanted to eliminate to sneak back in. 

We will need to figure out how to also make sure we recode reaction-field electrostatics as a `CustomNonbondedForce` in these cases, or simply allow OpenMM to set `c_rf = 0` (which may be far easier).

andrrizzi:
Fix #670 .

We'll need to cut a new openmmtools release before tests can pass. I've also run hydration free energy calculations with RF on the cluster to test if this makes it work.

### DIFF
--- a/Yank/tests/test_yank.py
+++ b/Yank/tests/test_yank.py
@@ -21,6 +21,7 @@ This code is licensed under the latest available version of the MIT License.
 # GLOBAL IMPORTS
 # ==============================================================================
 
+import os
 import functools
 import contextlib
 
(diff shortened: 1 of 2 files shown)


jchodera:
Looks good to me!

Issue #720, opened 2017-07-17
MC states not mixing in 0.16


spadavec:
For nearly the same experiment (only difference is slight change in `clearance` and existence of `switch_distance`), MC states seem to not mix in ```0.16```, while it seems to work in ```0.15``` (see attached YAML files)--i doubt these differences contribute to the overall result though. 
[yank_mc.zip](https://github.com/choderalab/yank/files/1153188/yank_mc.zip)

During `yank analyze` in `0.15`, the reported state mixing matrix is (relatively) populated (see attached `JQ1_15_analyze.out`), while in `0.16`, it is completely empty (see attached `JQ1_16_analyze.out`).  Furthermore, the following is also reported in ```0.16```: 

> 2017-07-17 07:59:50,232: Perron eigenvalue is   0.00000; state equilibration timescale is ~ 1.0 iterations

Unsurprisingly, the dG estimate is very, very wrong in ```0.16```  (~ +90 kcal/mol), and relatively correct in `0.15` (~ -15 kcal/mol). Interestingly, `yank analyze` takesmuch longer in ```0.16```  (currently ~5 min vs. 20 seconds). 

[yank_mc.zip](https://github.com/choderalab/yank/files/1153200/yank_mc.zip)

andrrizzi:
Hmm. That's weird! From the log `JQ1_16.out` it looks like the acceptance rate of the swap is around 6-8% at each iteration
```
2017-07-14 19:11:50,106: Accepted 390584/5120000 attempted swaps (7.6%)
```
will investigate what's happening.

Lnaden:
I'm seeing about 6-7% exchange between  all states in the complex phase, and about 10-12% exchange in the solvent phase.

That output seems to indicate that *mixing is ideal* to machine precision and everything mixes well, not that its not mixing. The reason you are not seeing mixing in the transition state table is that its below the threshold to show (0.05), not because nothing is mixing, but because there are so many swaps that no one state has much occupation. 

This seems very odd to me since the 0.16 version seems to indicate your complex phase deltaG is very small (backed up by the very high level of mixing). 

Could you possibly run `yank analyze report` to create an ipython notebook, run all the cells and post/send that? You can use the command line help to see the flags (has a `-s` for store directory and `-o` for output file name). If you run the cells locally, we can actually look at what the analyze code is doing.

I'm not sure why its suddenly so much slower though. 500 iterations should not take that long.

[...]

PR #721, opened 2017-07-17
Raise error when number of states in protocol differ

andrrizzi:
Address two of the points raised in #720 .
- Log the atoms indices automatically picked by the `Boresch` restraint.
- Raise an error when the user selects a protocol with different number of states.

### DIFF
--- a/Yank/yank.py
+++ b/Yank/yank.py
@@ -16,7 +16,6 @@ Interface for automated free energy calculations.
 # GLOBAL IMPORTS
 # ==============================================================================
 
-import os
 import abc
 import copy
 import time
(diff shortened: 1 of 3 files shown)


Yank/yank.py
```diff
@@ -473,6 +472,12 @@ def create(self, thermodynamic_state, sampler_states, topography, protocol,
             Simulation metadata to be stored in the file.
 
         """
+        # Check that protocol has same number of states for each parameter.
+        len_protocol_parameters = {par_name: len(path) for par_name, path in protocol.items()}
+        if len(set(len_protocol_parameters.values())) != 1:
+            raise ValueError('The protocol parameters have a different number '
+                             'of states: {}'.format(len_protocol_parameters))
```
Lnaden: What does this look like when you format a dict to print? is it just `dict.__str__()`?

andrrizzi: Yeah, it looks something like `{lambda_sterics: 22, lambda_electrostatics: 20}`. Not the prettiest, but it's a very short dict, so it should be readable.

Lnaden: Eh, errors dont need to be "pretty" so long as they are accurate

Lnaden: And readable

andrrizzi:
Tests pass!

Issue #119, opened 2014-11-15
Added 'autotune' command for automated tuning of alchemical protocol


jchodera:
We should add an `autotune` command to the cli that would attempt to optimize the alchemical protocol to minimize the discrete analog to thermodynamic length by reweighting existing data in the repex stores.

Lnaden:
Post 1.0 issue

andrrizzi:
Resurrecting this thread as we'll have to add this for YANK 1.0 paper.

As I see it, we have two options (not necessarily exclusive) to expose this feature

1. A new command, something like `yank autotune protocol --yaml=yank.yaml` or `yank optimize protocol --yaml=yank.yaml`.
2. Add this directly to the YAML syntax with something like
```yaml
protocols:
  absolute-binding:
    complex:
      alchemical_path: auto
    solvent:
      alchemical_path: auto
```

In both cases, I think the output should be multiple YAML scripts, one for each experiment, rather than a single monolithic, unreadable new script, especially for big combinatorial experiments.

For this to work, I would have to change a little the behavior of `ExperimentBuilder` to make it perform a recursive search into the experiment folders looking for YAML scripts to run. Currently, the single-experiment scripts are saved only for backup purposes. This structure could also simplify parallelizing multiple experiments in the future.

jchodera:
I think there should be two distinct features:
* (1) Construct an initial guess for the alchemical protocol (implemented for yank 1.0)
* (2) Update the alchemical protocol based on an existing simulation (implemented sometime in the future)

For (1), I like the idea of adding the `auto` keyword directly to the protocol description in the YAML. Is it possible that, as YANK *generates* the explicitly substituted YAML file for the experiment it is about to run, it could detect whether `auto` is used and carry out the fast lambda scan to produce an initial protocol guess at that point, and then fill in the explicit protocol in the experiment directory? Or would this make early termination unrecoverable?

andrrizzi:
I agree about the two distinct features.

For (1), I was thinking to implement it exactly that way, possibly in parallel using the `mpi` library. In other words, `ExperimentBuilder` would do sequentially
1. Run the setup pipeline on all molecules/systems that needed it (possibly in parallel in the future).
2. Create a single-experiment YAML script for each combination and save it in its experiment folder. In this phase, YANK would run the automatic alchemical path determination for each experiment that specifies the `auto` keyword (in parallel).
3. Run all the experiments using the single-experiment YAML scripts generated in step 2).

[...]

PR #722, opened 2017-07-19
API, Documentation, and Dropdowns (Oh My!)

Lnaden:
This PR is the first of a series dedicated to the website. This particular one contains three major parts:

1. Fully documented the entire public API in docstrings.
2. Automatic documentation generation on the website in the `API Reference` category
3. Collapsable options in all the YAML pages to clean make things easier to read on not sift through a wall of text.

The text changes can be reviewed here, but visual changes (including autodoc generation) must be seen by checking out this branch, *installing yank*, then generating the docs yourself. The autodoc feature of Sphinx looks at the currently installed version by import, not what is in the repo itself (if you know how to change that, please comment).

@andrrizzi Some of the docstrings I added were for code you wrote, so if you could at minimum review those, that would be very helpful. Thanks in advance!

PR #726, opened 2017-07-24
Automatic determination of protocol and parallel experiments

### LINKED ISSUE #725: Run combinatorial experiments in parallel
Since the YANK 1.0 paper will involve running a lot of molecules, I'm also thinking of adding a way to run multiple experiments on single nodes. This way we won't have tens of nodes in idle while we write on the disk and/or perform the online analysis.

With the new feature in the MPI library (PR coming) it will be very easy to implement both of these:
- add a YAML option `options:nodes_per_experiment:2` that allows to spawn a job on multiple MPI processes and allocate (in this case) 2 nodes per experiments by splitting the mpicomm.
- add two options to the `yank script` command `--job_id=2` and `--n_jobs=10` that (in this case) will run only one tenth of the experiments on whatever number of processes.

Personally, I would implement them both (it would take me only an hour at this point). The second one could be very helpful for having parallelization even on systems that with no MPI, or simply in times where the cluster is busy and it's not possible to gather a lot of gpus at the same time.

What do you think?

andrrizzi:
Still working on it. Should be able to finish the PR as soon as I get feedback on #725 (and eventually fix the tests).

- [x] Adds to `mpi.distribute` a `group_nodes` argument that allows to split the `mpicomm` automatically and distribute a task that also coordinates multiple MPI nodes.
- [x] Implements the `auto` keyword to determine the `alchemical_path` of a phase as described in #119 .
- [x] Fix #725: run combinatorial experiments in parallel.


andrrizzi:
Done! Should be ready for review. I just have to update the docs after #722 will be merged.

Not sure what to do about the cyclomatic complexity issues in codeclimate right now.

@Lnaden I think I've fixed a couple of bugs in repex and made a small modification of the online analysis variables to have `ReplicaExchange` read data from the reporter only in `from_storage()`. This way I was able to use an [extremely dummy `DummyReporter`](https://github.com/choderalab/yank/blob/auto-path/Yank/experiment.py#L1622-L1628) to build the experiments for path optimization (writing on disk takes a lot of time and we don't need the nc files for this). Could you check that I didn't mess anything up (not urgent)?

Yank/experiment.py
```diff
@@ -353,6 +359,7 @@ class ExperimentBuilder(object):
         'platform': 'fastest',
         'precision': 'auto',
         'switch_experiment_interval': 0,
+        'nodes_per_experiment': None
```
jchodera: Should we document the YAML option `nodes_per_experiment` in the docs?

Does it represent *nodes* or *MPI processes = GPUs* used per experiment?

Yank/pipeline.py
```diff
@@ -1417,6 +1417,182 @@ def _setup_system(self, system_file_path, pack, alchemical_charge,
             logger.warning('TLeap: ' + warning)
 
 
+# ==============================================================================
+# FUNCTIONS FOR ALCHEMICAL PATH OPTIMIZATION
+# ==============================================================================
+
+def restrain_atoms(thermodynamic_state, sampler_state, mdtraj_topology,
```
jchodera: We should move this to `openmmtools.forces` or `openmmtools.forcefactory` if possible.

Yank/experiment.py
```diff
@@ -1479,7 +1575,167 @@ def _setup_experiments(self):
...
+                state_parameters.append(('lambda_sterics', [1.0, 0.0]))
+
+            # We only need to create a single state.
+            phase.protocol = {par[0]: [par[1][0]] for par in state_parameters}
+
+            # Remove unsampled state that we don't need for the optimization.
+            phase.options['anisotropic_dispersion_correction'] = False
+
+            # If default argument is used, determine number of equilibration iterations.
+            # TODO automatic equilibration?
```
jchodera: It would be nice to move away from hard-coded number of equilibration iterations here, at least allowing flexibility in selecting these via some appropriate key in the YAML file. 

If we can at least specify in the YAML how many *total* iterations to use for auto determination, it would be very easy to use automatic equilibration detection to compute the optimal estimates of <DeltaU>---that's literally a single additional line, so I'd suggest we add that now at the very least, which goes into `find_alchemical_protocol`:
```python
denergies = reweighted_energies - simulated_energies
[t0, g, Neff_max] = pymbar.timeseries.detectEquilibration(denergies)
std_energy = np.std(denergies[t0:])
```

Yank/pipeline.py
```diff
@@ -1417,6 +1417,182 @@ def _setup_system(self, system_file_path, pack, alchemical_charge,
...
+    for index in restrained_atoms:
+        parameters = sampler_state.positions[index,:].value_in_unit_system(unit.md_unit_system)
+        restraint_force.addParticle(index, parameters)
+
+    # Update thermodynamic state.
+    system.addForce(restraint_force)
+    thermodynamic_state.system = system
+
+
+def find_alchemical_protocol(thermodynamic_state, sampler_state, mcmc_move, state_parameters,
```
jchodera: Maybe we can call this `trailblaze_alchemical_protocol` after David Minh's suggestion?

Yank/experiment.py
andrrizzi: Ah, right! I'll call it `processes_per_experiment` and add it to the website docs.

Yank/experiment.py
andrrizzi: Hmm, sorry I'm not sure I understand. The equilibration iterations are run before computing any energy and before restraining the receptor atoms. Then at each iteration we generate a fixed number of samples to compute `denergies`, but we don't perform equilibration every time we compute it.

Yank/pipeline.py
andrrizzi: I'll move it to `openmmtools.forcefactories`. We'll need to make an `openmmtools` release before merging this then and update the minimum version in the conda recipe.

Yank/pipeline.py
andrrizzi: Will do!

Yank/experiment.py
jchodera: > Hmm, sorry I'm not sure I understand. The equilibration iterations are run before computing any energy and before restraining the receptor atoms. Then at each iteration we generate a fixed number of samples to compute denergies, but we don't perform equilibration every time we compute it.

I'm suggesting that instead, we simply run some specified number of iterations `niterations` iterations, computing `denergies` for each iteration, storing the whole array. After that, we can determine what the optimal fraction of data to use is via `detectEquilibration`. If computation of the `denergies` array is fast, this should be fine; if it's really slow, this might slow things down a lot.

andrrizzi: Ah, I see, but shouldn't we run a preliminary equilibration even in that case? Otherwise, we'll probably have to push `niterations` to a very high number to get some decent info from the first iterations of the algorithm (that are used to determine the `lambda_electrostatics` path), and from the idea I got by playing around with the script, I think that would slow down things down a lot.

jchodera: With CA restraints, the only real requirement is to equilibrate within the local well at the current lambda value which should be relatively rapid. On the other hand, without CA restraints, a decent amount of equilibration could be required. Adding computation at a variety of lambda values can also slow things down a great deal. 

The idea of adding CA restraints and simply using `detectEquilibration` to decide which part to analyze is an attempt to get beyond arbitrarily specified equilibration/production numbers of iterations, but it may be computationally infeasible to do so. As long as these numbers are exposed as YAML-controllable, it's likely OK for now---we can do a more systematic study of automated schemes for another paper. I just object to the hard-coding of these values for now.

andrrizzi: > I just object to the hard-coding of these values for now

Yeah, I understand, I don't like it either. The reason I did it is that everything is still very experimental, and I didn't want to expose YAML options in YANK 1.0 that we would have to break after few weeks.

If you agree, I'd leave it as it is now, and postpone this decision to after we have generated the first paths for the YANK 1.0 paper simulations, so that we'll have a better idea of what we actually need before engineering stuff in. We can use the API to make experiments.

jchodera: OK, let's capture that TODO in an issue and move ahead then.

andrrizzi: Ok, thanks! I'll open an issue.

andrrizzi:
With the new openmmtools release, tests pass! Merging.

PR #733, opened 2017-07-26
Fix casting jobid and njobs CLI parameters to int

andrrizzi:
See title. Should be ready for review if tests pass.


andrrizzi:
Thanks!

PR #734, opened 2017-07-28
Add possibility to specify ionic strength

andrrizzi:
Should be ready for review!

Only monovalent ions are supported right now (I've added an assert statement to check this in the code).

### DIFF
--- a/Yank/pipeline.py
+++ b/Yank/pipeline.py
@@ -24,6 +24,7 @@ import logging
 import itertools
 import collections
 
+import mdtraj
 import numpy as np
 import openmmtools as mmtools
 import openmoltools as moltools
@@ -1275,7 +1276,7 @@ class SetupDatabase:
             self._processed_mols.add(mol_id)
 
     def _setup_system(self, system_file_path, pack, alchemical_charge,
-                      system_parameters, solvent_id, *molecule_ids):
+                      system_parameters, solvent_id, *molecule_ids, **kwargs):
         """Setup a system and create its prmtop/inpcrd files.
 
         IMPORTANT: This function does not check if it's about to overwrite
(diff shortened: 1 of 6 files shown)


Yank/pipeline.py
```diff
@@ -1424,6 +1457,22 @@ def _setup_system(self, system_file_path, pack, alchemical_charge,
...
+        with mmtools.utils.temporary_directory() as tmp_dir:
+            system_file_path = os.path.join(tmp_dir, 'temp_system.prmtop')
+            self._setup_system(system_file_path, *args, ignore_ionic_strength=True,
+                               save_amber_files=False)
+
+            # Count number of waters of created system.
+            system_file_path = os.path.join(tmp_dir, 'temp_system.pdb')
+            system_traj = mdtraj.load(system_file_path)
+            n_waters = 0
+            for res in system_traj.topology.residues:
```
jchodera: You can probably use a generator idiom here, like:
```python
n_waters = sum(int(res.is_water) for res in system_traj.topology.residues)
```

docs/yamlpages/solvents.rst
```diff
@@ -323,13 +323,38 @@ Specifies the negative counter ions that will be added as needed.
 No negative counter ions will be added if this option is not specified. Note that the name must match a known atom type
 in LEaP based on the parameter files you specified to load.
 
-Nonbonded Methods: ``CuttoffPeriodic``, ``Ewald``, ``PME``
+Nonbonded Methods: ``CutoffPeriodic``, ``Ewald``, ``PME``
```
jchodera: I think we want to remove ``CutoffPeriodic`` from the docs for now!

docs/yamlpages/solvents.rst
```diff
@@ -323,13 +323,38 @@ Specifies the negative counter ions that will be added as needed.
...
+       ionic_strength: 0.0*molar
+
+The ionic strength of the ions.
+
+Both ``positive_ion`` and ``negative_ion`` must be specified with this, and only monovalent ions are supported. Note
+that this does not include the ions that are used to neutralize the periodic box.
+
+Nonbonded Methods: ``CutoffPeriodic``, ``Ewald``, ``PME``
+
+Valid Options (0 * molar) <Quantity Length> [1]_
```
jchodera: I'm confused about what this line means, and why there's a `<Quantity Length>`.

jchodera:
Looks good to me except for the documentation bit!

docs/yamlpages/solvents.rst
andrrizzi: Right! I'll take a pass.

docs/yamlpages/solvents.rst
andrrizzi: Whops! Copy-paste error. I'll fix it.

The value in-between parenthesis is the default value. It's the same syntax used throughout the docs.

andrrizzi:
Tests pass!

PR #735, opened 2017-07-28
Update site: Algorithm Docs Update

Lnaden:
This PR finishes out the algorithms page of the docs. There is a fair amount here, so I wanted to make the PR now so we can fix anything I got wrong.

@andrrizzi Do you think you could add to the "Autotuning the alchemical protocol" section since you're the one who wrote that feature? If not, I can take a stab at it after going over your code.

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -1026,7 +1026,7 @@ class MultiPhaseAnalyzer(object):
 
         ``PhaseB`` has ``.get_free_energy`` and ``.get_enthalpy``,
 
-        ``PhaseAB = PhaseA + PhaseB`` will only have a .get_free_energy method
+        ``PhaseAB = PhaseA + PhaseB`` will only have a ``.get_free_energy`` method
 
     Because each Phase may have a different number of states, the ``reference_states`` property of each phase
     determines which states from each phase to read the data from.
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -1016,10 +1016,7 @@ class MultiPhaseAnalyzer(object):
     """
     Multiple Phase Analyzer creator, not to be directly called itself, but instead called by adding or subtracting
     different implemented YankPhaseAnalyzer or other MultiPhaseAnalyzers's. The individual Phases of the
-    MultiPhaseAnalyzer are only references to existing Phase objects, not copies.
-
-    The user themselves should not attempt to make this class by calling it directly, but instead through doing
-    addition and subtraction of YankPhaseAnalyzer objects and/or other MultiPhaseAnalyzer objects. All YankPhaseAnalyzer
+    MultiPhaseAnalyzer are only references to existing Phase objects, not copies. All YankPhaseAnalyzer
     and MultiPhaseAnalyzer classes support ``+`` and ``-`` operations.
 
     The observables of this phase are determined through inspection of all the passed in phases and only observables
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -95,8 +95,8 @@ def get_analyzer(file_base_path):
     """
     Utility function to convert storage file to a Reporter and Analyzer by reading the data on file
 
-    For now this is mostly placeholder functions since there is only the implemented RepexPhase, but creates the
-    API for the user to work with.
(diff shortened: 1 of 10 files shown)


Yank/analyze.py
```diff
@@ -1016,10 +1016,7 @@ class MultiPhaseAnalyzer(object):
     """
     Multiple Phase Analyzer creator, not to be directly called itself, but instead called by adding or subtracting
     different implemented YankPhaseAnalyzer or other MultiPhaseAnalyzers's. The individual Phases of the
-    MultiPhaseAnalyzer are only references to existing Phase objects, not copies.
-
-    The user themselves should not attempt to make this class by calling it directly, but instead through doing
-    addition and subtraction of YankPhaseAnalyzer objects and/or other MultiPhaseAnalyzer objects. All YankPhaseAnalyzer
+    MultiPhaseAnalyzer are only references to existing Phase objects, not copies. All YankPhaseAnalyzer
```
jchodera: Should we be using the Sphinx markup when mentioning classes?

docs/algorithms.rst
```diff
@@ -112,28 +112,21 @@ Explicit solvent
 Solvent model
 ^^^^^^^^^^^^^
 
-.. warning:: Only TIP3P is supported through the :ref:`YAML method of setting up solvents <yaml_head>` at this moment. We are working on adding back in support for other water models before the YANK 1.0 release, but after the preview build
+Any explicit solvent model that can be constructed via AmberTools or that is distributed along with OpenMM is supported.
```
jchodera: Do we really support addition of solvent from OpenMM at this time? I thought it was only those solvents present in AmberTools.

docs/algorithms.rst
```diff
@@ -347,27 +340,231 @@ This speeds up mixing and reduces the total number of samples you need to get un
 Markov chain Monte Carlo
 ========================
 
+YANK operates with additional sampling enhancements based on Markov Chain Monte Carlo (MCMC) techniques, and implements
```
jchodera: Awkward. Maybe:
"YANK uses a replica exchange scheme with Gibbs sampling (CITE) to sample multiple thermodynamic states in a manner that improves mixing of the overall Markov chain.
In between replica exchange attempts, YANK uses compositions of MCMC moves from the ```openmmtools.mcmc`` module <http://openmmtools.readthedocs.io/en/latest/mcmc.html>`_ to sample from the equilibrium distribution at each thermodynamic states."

docs/algorithms.rst
```diff
@@ -347,27 +340,231 @@ This speeds up mixing and reduces the total number of samples you need to get un
...
+
+    u_*(X) \equiv \sum_{k=1}^K u_{s_k}(x_k)
+
+That is, :math:`u_*(X)` is the sum of the reduced potential energies of each replica configuration at the current
+thermodynamic state it is visiting.
+
+Determining equilibration and autocorrelation in a replica-exchange simulation
+------------------------------------------------------------------------------
+
+Now that we have the observable, we now need to evaluate the equilibration, and autocorrelation times.
```
jchodera: Suggest: "YANK automatically determines an optimal partitioning between the initial equilibration phase of the simulation (which is discarded) and the production phase of the simulation (which is analyzed to estimate free energies)."

docs/algorithms.rst
```diff
@@ -347,27 +340,231 @@ This speeds up mixing and reduces the total number of samples you need to get un
...
+presented. However, we really only care about differences in free energies between states, so one state is chosen as
+the reference, so :math:`f_0 = 0`. This equation is also not conditioned on state :math:`i` being part of the sampled
+states :math:`K`, allowing estimation of the free energy at any thermodynamic state such as the unsampled states with
+expanded cutoffs that YANK supports.
+
 Automated convergence detection
 ===============================
 
-Will extract information from `here <http://nbviewer.ipython.org/github/choderalab/simulation-health-reports/blob/master/examples/yank/YANK%20analysis%20example.ipynb>`_.
+YANK has the ability to run a simulation until the free energy difference in a phase reaches a target uncertainty.
```
jchodera: "reaches a user-specified target uncertainty"

PR #738, opened 2017-08-01
Generalize Harmonic and FlatBotton restraints to be referenced to centroids

### LINKED ISSUE #736: Implement restraints referenced to center of geometries of atom selections
This is just a placeholder to remind us that we will need to implement variants of the `Harmonic` and `FlatBottom` restraints that use the centers of geometry of DSL-specified atom selections in order to carry out host-guest (and like T4 lysozyme) binding free energies for the paper.

We can use [`CustomCentroidBondForce`](http://docs.openmm.org/7.1.0/api-python/generated/simtk.openmm.openmm.CustomCentroidBondForce.html#simtk.openmm.openmm.CustomCentroidBondForce) for this purpose. For example, `Harmonic` would use something like this
```
# Harmonic restraint between two groups of atoms
num_groups = 2
force = openmm.CustomCentroidBondForce(num_groups, "(K/2)*distance(g1,g2)^2")
force.addGlobalParameter("K", K)
force.addGroup(particles1); # create group 0
force.addGroup(particles2); # create group 1
force.addBond([0,1], []); # add bond between groups 0 and 1
system.addForce(force)
```

andrrizzi:
Fix #736 . Should be ready for review if tests pass!

Instead of taking an integer to specify the receptor and ligand atoms, `RadiallySymmetricRestraint` now takes either a list of particles indices or a DSL expression + topology.

If the list of particles for receptor and ligand only has 1 index, `CustomBondForce` is used instead of `CustomCentroidBondForce`. This is because the latter supports only 64bit platforms. This way we can still run tests with a 32bit one. For example, on the Macbooks not equipped with an NVIDIA card, the OpenCL platform is 32bit.

### DIFF
--- a/Yank/restraints.py
+++ b/Yank/restraints.py
@@ -425,23 +425,36 @@ class RadiallySymmetricRestraint(ReceptorLigandRestraint):
     """
     Base class for radially-symmetric restraints between ligand and protein.
 
-    This class is intended to be inherited by Restraints using a `CustomBondForce`.
     The restraint strength is controlled by a global context parameter called
     'lambda_restraints'.
 
-    The class allows the restrained atoms to be temporarily, but in this case
-    `determine_missing_parameters()` must be called before using the restraint.
+    The class allows the restrained atoms to be temporarily undefined, but in
+    this case ``determine_missing_parameters()`` must be called before using
+    the restraint.
 
     Parameters
     ----------
-    restrained_receptor_atom : int, optional
-        The index of the receptor atom to restrain. This can temporarily be left
-        undefined, but in this case `determine_missing_parameters()` must be called
-        before using the Restraint object (default is None).
-    restrained_ligand_atom : int, optional
-        The index of the ligand atom to restrain. This can temporarily be left
-        undefined, but in this case `determine_missing_parameters()` must be called
-        before using the Restraint object (default is None).
+    restrained_receptor_atoms : list of int or str, optional
+        The indices of the receptor atoms to restrain, or an MDTraj DSL expression.
+        This can temporarily be left undefined, but ``determine_missing_parameters()``
+        must be called before using the Restraint object. The same if a DSL
+        expression is provided without passing the topology (default is None).
+    restrained_ligand_atoms : list of int or str, optional
+        The indices of the ligand atoms to restrain, or an MDTraj DSL expression.
+        This can temporarily be left undefined, but ``determine_missing_parameters()``
+        must be called before using the Restraint object. The same if a DSL
+        expression is provided without passing the topology (default is None).
+    topology : simtk.openmm.app.Topology or mdtraj.Topology, optional
+        The topology used to resolve the DSL string. If not provided, the atom
+        selection expressions will be resolved by ``determine_missing_parameters()``.
+
+    Attributes
+    ----------
+    restrained_receptor_atoms : list of int or None
+        The indices of the receptor atoms to restrain.
+    restrained_ligand_atoms : list of int or None
+        The indieces of the ligand atoms to restrain.
+    topology : mdtraj.Topology
 
     Notes
     -----
@@ -456,9 +469,62 @@ class RadiallySymmetricRestraint(ReceptorLigandRestraint):
         function to automatically determine these parameters from the atoms positions.
 
     """
-    def __init__(self, restrained_receptor_atom=None, restrained_ligand_atom=None):
-        self.restrained_receptor_atom = restrained_receptor_atom
-        self.restrained_ligand_atom = restrained_ligand_atom
+    def __init__(self, restrained_receptor_atoms=None, restrained_ligand_atoms=None, topology=None):
+        self.restrained_receptor_atoms = restrained_receptor_atoms
+        self.restrained_ligand_atoms = restrained_ligand_atoms
+        self.topology = topology
+
+    # -------------------------------------------------------------------------
+    # Public properties.
+    # -------------------------------------------------------------------------
+
+    class _RestrainedAtomsProperty(object):
+        """
+        Descriptor of restrained atoms.
+
+        Evaluate DSL atoms selections at runtime to allow partially specified restraints.
+
+        """
+        def __init__(self, atoms_type):
+            self._attribute_name = '_restrained_' + atoms_type + '_atoms'
+
+        def __get__(self, instance, owner_class=None):
+            restrained_atoms = getattr(instance, self._attribute_name)
+
+            # On-demand resolution of restrained receptor atoms.
+            if isinstance(restrained_atoms, str):
+                if instance.topology is None:
+                    raise RestraintParameterError('Cannot resolve receptor DSL expression "{}" '
+                                                  'without topology'.format(restrained_atoms))
+                restrained_atoms = instance.topology.select(restrained_atoms).tolist()
+                setattr(instance, self._attribute_name, restrained_atoms)
+
+            # Return None if an empty list.
+            if restrained_atoms is None or len(restrained_atoms) == 0:
+                return None
+            return restrained_atoms
+
+        def __set__(self, instance, new_restrained_atoms):
+            setattr(instance, self._attribute_name, new_restrained_atoms)
+
+    restrained_receptor_atoms = _RestrainedAtomsProperty('receptor')
+    restrained_ligand_atoms = _RestrainedAtomsProperty('ligand')
+
+    @property
+    def topology(self):
+        """The topology used to solve DSL atom selection"""
+        # On-demand conversion of OpenMM topology to mdtraj topology.
+        if isinstance(self._topology, openmm.app.Topology):
+            self._topology = md.Topology.from_openmm(self._topology)
+        return self._topology
+
+    @topology.setter
+    def topology(self, new_topology):
+        self._topology = new_topology
+
+    # -------------------------------------------------------------------------
+    # Public methods.
+    # -------------------------------------------------------------------------
 
     def restrain_state(self, thermodynamic_state):
         """Add the restraining Force(s) to the thermodynamic state's system.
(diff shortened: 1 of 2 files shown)


Yank/restraints.py
```diff
@@ -714,22 +787,24 @@ def _create_restraint_force(self, particle1, particle2):
...
-        force = openmm.CustomBondForce('lambda_restraints * ' + self._energy_function)
-        force.addGlobalParameter('lambda_restraints', 1.0)
+        if len(particles1) == 1 and len(particles2) == 1:
+            # CustomCentroidBondForce works only on 64bit platforms. When the
+            # restrained groups only have 1 particle, we can use the standard
+            # CustomBondForce so that we can support 32bit platforms too.
+            energy_function = self._energy_function.replace('distance(g1,g2)', 'r')
+            force = openmm.CustomBondForce('lambda_restraints * ' + energy_function)
+            force.addBond(particles1[0], particles2[0], parameter_values)
+        else:
```
Lnaden: Should we add an exception for 32-bit architecture, or at bare minimum a full documentation that you have to have 64bit if you want to use this feature for multiple atoms?

Yank/restraints.py
```diff
@@ -714,22 +787,24 @@ def _create_restraint_force(self, particle1, particle2):
         parameter_names, parameter_values = zip(*self._bond_parameters.items())
 
         # Create bond force and lambda_restraints parameter to control it.
-        force = openmm.CustomBondForce('lambda_restraints * ' + self._energy_function)
-        force.addGlobalParameter('lambda_restraints', 1.0)
+        if len(particles1) == 1 and len(particles2) == 1:
+            # CustomCentroidBondForce works only on 64bit platforms. When the
```
Lnaden: Where in the docs does it say it only supports 64-bit platforms? And what is the exact error it throws if you use the wrong platform?

Lnaden:
Just a couple comments, mostly just documentation changes requested. Can you also add in the auto-function linking to the docs once you merge the conflicting restraints file in? It should just be documentation clashes, not actual functionality ones.

Yank/restraints.py
andrrizzi: I didn't find it in the docs, but it's in the OpenCL platform code ([here](https://github.com/pandegroup/openmm/blob/master/platforms/opencl/src/OpenCLKernels.cpp#L5054)).

Yank/restraints.py
andrrizzi: OpenMM already raises a very clear exception if that's the case on `Context` creation.

andrrizzi:
> add this support to the YAML syntax

I didn't touch the YAML syntax docs because it redirects to the API documentation to see the options taken by each type of restraint.

> Can you also add in the auto-function linking to the docs 

Will do!

andrrizzi:
I'm also going to separate the `restrained_atoms` argument in the `Boresch` constructor into two `restrained_receptor_atoms` and `restrained_ligand_atoms` because I've just realized that it'll be problematic to specify SMIRKS regions in the future with a single parameter.

Yank/restraints.py
andrrizzi: I'll add the 64bit thing to the docstrings!

jchodera:
Thanks everyone --- this is great!

If we wanted to add a test to our validation suite, we could also simulate a host-guest system where we keep all lambdas at zero except for turning on the restraint lambdas over the course of the simulation. If we keep the box size fixed at the standard state volume, the standard state correction should exactly cancel the restraining free energy and we should get a zero free energy for that leg. (alternatively, if the box size is arbitrary, we would get the log of the ratio of box volumes)

andrrizzi:
@Lnaden sorry, the last two commits need review again.

>  to add a test to our validation suite

Sounds good! I'll try to make that work when I'll have the host-guest systems ready.

Yank/restraints.py
```diff
@@ -1476,7 +1516,9 @@ def determine_missing_parameters(self, thermodynamic_state, sampler_state, topog
                              'restraint parameters...'.format(attempt, MAX_ATTEMPTS))
 
                 # Randomly pick non-collinear atoms.
-                self.restrained_atoms = self._pick_restrained_atoms(sampler_state, topography)
+                restrained_atoms = self._pick_restrained_atoms(sampler_state, topography)
+                self.restrained_receptor_atoms = restrained_atoms[:3]
```
Lnaden: Do we want this to auto-repick both the receptor AND ligand atoms? What if the user specifies a series of receptor atoms they always want to use and dont care about the ligand atoms? This method will re-pick the receptor ones as well.

Yank/tests/test_restraints.py
```diff
@@ -221,24 +221,40 @@ def test_partial_parametrization():
         ('Harmonic', dict(spring_constant=2.0*unit.kilojoule_per_mole/unit.nanometer**2,
                           restrained_receptor_atoms=[5])),
         ('FlatBottom', dict(well_radius=1.0*unit.angstrom, restrained_ligand_atoms=[130])),
-        ('Boresch', dict(restrained_atoms=[106, 123, 117, 130, 131, 136],
+        ('Boresch', dict(restrained_ligand_atoms=[130, 131, 136],
```
Lnaden: Doesn't this test also need the "restrainted_receptor_atoms"? And if you are wanting to just randomly choose the receptor atoms as per how it is coded right now, it will also randomly rechoose the ligand atoms.

Yank/restraints.py
andrrizzi: The `_pick_restrained_atoms` won't re-pick atoms that have been explicitly given, it basically just give return them as they are.

Yank/tests/test_restraints.py
andrrizzi: I did it on purpose to test that only receptor atoms are picked at random and that ligand atoms are left as they are.

Yank/restraints.py
```diff
@@ -1564,46 +1737,73 @@ def _pick_restrained_atoms(self, sampler_state, topography):
...
-        heavy_ligand_atoms = set(topography.ligand_atoms) & heavy_atoms
-        heavy_receptor_atoms = set(topography.receptor_atoms) & heavy_atoms
         if len(heavy_receptor_atoms) < 3 or len(heavy_ligand_atoms) < 3:
             raise ValueError('There must be at least three heavy atoms in receptor_atoms '
                              '(# heavy {}) and ligand_atoms (# heavy {}).'.format(
                                      len(heavy_receptor_atoms), len(heavy_ligand_atoms)))
 
-        # Find valid pairs of ligand/receptor atoms within a cutoff
+        # If r3 or l1 atoms are given. We have to pick those.
+        if isinstance(heavy_receptor_atoms, list):
```
Lnaden: This will fail if they are a tuple or a set (which it will be if auto-chosen), which should be equally valid

Yank/restraints.py
```diff
@@ -1564,46 +1737,73 @@ def _pick_restrained_atoms(self, sampler_state, topography):
...
+            return self.restrained_receptor_atoms + self.restrained_ligand_atoms
+
+        # If receptor and ligand atoms are explicitly provided, use those.
+        heavy_ligand_atoms = self.restrained_ligand_atoms
+        heavy_receptor_atoms = self.restrained_receptor_atoms
+
+        # Otherwise we restrain only heavy atoms.
+        heavy_atoms = set(topography.topology.select('not element H').tolist())
+        # Intersect heavy atoms with receptor/ligand atoms (s1&s2 is intersect).
+        if heavy_ligand_atoms is None:
```
Lnaden: Just noticed, if user chosen, hydrogens can get selected as the heavy atoms to use here, may be a problem.

Yank/restraints.py
Lnaden: More exactly, this will always be False under many conditions unexpectedly, which may confuse our users

Yank/restraints.py
andrrizzi: Are you suggesting to raise a warning? I wouldn't forbid the user to pick hydrogens if he wanted to (with mass repartitioning for example).

Yank/restraints.py
andrrizzi: Shouldn't be a problem. `heavy_receptor_atoms` and `heavy_ligand_atoms` will be `list`s only if the user has specified them, otherwise they'll be `set`s.

Yank/restraints.py
Lnaden: Hmm, good point, that is fine

Yank/restraints.py
Lnaden: What is stopping the user from passing in a tuple or set of heavy atoms for either of those themselves?

andrrizzi: Ah, I see what you're saying now. The user can specify tuples, but the `restrained_receptor/ligand_atoms` property converts them to lists.

Lnaden: Ah, I see now. I've made a 3rd pass through and retract my concerns. lgtm

andrrizzi:
Awesome, thanks!

davidlmobley:
@elkhoury - looks like centroid restraints are in Yank now.

Thanks, all! @andrrizzi - would love sample inputs as soon as you have them (as I think Levi mentioned we're after seeing how to activate the different types of restraints in Yank in the context of a binding calculation, such as binding of a host to a guest -- flat bottom restraints, Boresch-style, and no restraints at all. Sample yaml files would be awesome).

We're going to be doing some benchmarking, initially on some specific host-guest system from the benchmark sets paper (probably overlapping with what you're looking at for the Yank paper) to examine the efficiency of different forms of restraints. Léa El Khoury, who I tagged above, is a new postdoc in the group just starting to figure out how to use Yank, so if you're able to share some examples as you set up these systems that will help her out. We're coordinating with Mehtap on this as well as the thought is that ultimately the work we're doing will connect up with her studies on multiple-site binding in HSA.

Thanks!

jchodera:
> We're going to be doing some benchmarking, initially on some specific host-guest system from the benchmark sets paper (probably overlapping with what you're looking at for the Yank paper) to examine the efficiency of different forms of restraints

@andrrizzi is in the process of setting those exact systems up right now for the accuracy benchmark in the YANK paper.

davidlmobley:
Yup. We'll use just one of them. I just wanted to point out the overlap since if he just lets us spectate on his setup then we'll be able to learn a lot. :)

andrrizzi:
> so if you're able to share some examples as you set up these systems that will help her out.

Sure! I hope I'll be able to create examples in the next couple of days. I'll upload the files in a repo and ping you guys as soon as they are ready.

PR #739, opened 2017-08-01
Fix MPI parallelization without processes_per_experiment

andrrizzi:
Thanks @Lnaden for finding out! I'll let the tests pass before merging.

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -718,9 +718,14 @@ class ExperimentBuilder(object):
             # until all of them are done.
             completed = [False]  # This is just to run the first iteration.
             while not all(completed):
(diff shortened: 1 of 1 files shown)


Yank/experiment.py
```diff
@@ -716,11 +716,16 @@ def run_experiments(self):
 
             # Cycle between experiments every switch_experiment_interval iterations
             # until all of them are done.
-            completed = [False]  # This is just to run the first iteration.
+            completed = [False for _ in range(len(all_experiments))]
```
Lnaden: You could have just said `[False] * len(all_experiments)` here, but thats just list compression.

PR #740, opened 2017-08-02
Analyze api cleanup

Lnaden:
* Moved analyze static class functions to module functions
* Renamed some functions and classes to be more clear
* Added ability to pre-render jupyter notebooks as static HTML or PDF objects
    * PDF requires `xelatex` binary, provided by most LaTeX engines
    * Personal note: This was much harder than it should have been

### DIFF
--- a/Yank/commands/analyze.py
+++ b/Yank/commands/analyze.py
@@ -15,6 +15,7 @@ Analyze YANK output file.
 
 from .. import utils, analyze
 import re
+import os
 import pkg_resources
 
 # =============================================================================================
@@ -37,7 +38,10 @@ Free Energy Required Arguments:
   -s=STORE, --store=STORE       Storage directory for NetCDF data files.
 
 YANK Health Report Arguments:
-  -o=REPORT, --output=REPORT    Name of the health report Jupyter notebook, can use a path + name as well
+  -o=REPORT, --output=REPORT    Name of the health report Jupyter notebook or static file, can use a path + name as well
+                                If the filename ends in .pdf or .html, the notebook is auto run and converted to a
+                                static PDF or HTML file respectively.
+                                Static generation may be slow!
 
 Extract Trajectory Required Arguments:
   --netcdf=FILEPATH             Path to the NetCDF file.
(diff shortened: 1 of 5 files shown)


Yank/commands/analyze.py
```diff
@@ -114,18 +119,67 @@ def dispatch_extract_trajectory(args):
...
     except ImportError:
-        print("Rendering this notebook requires the following packages:\n"
-              " - matplotlib\n"
-              "These are not required to generate the notebook however")
-    store = args['--store']
-    output = args['--output']
+        error_msg = "Rendering this notebook requires the following packages:\n"
+        " - matplotlib\n"
+        " - jupyter\n"
+        "These are not required to generate the notebook however"
```
andrrizzi: I don't think this works if you don't have parentheses around the strings.

Lnaden: I think that did work while I was in py 2 and 3, I'll double check it though.

andrrizzi: I think it doesn't raise an error because a line with a string is a valid Python expression, but the strings are not concatenated and `error_msg` will be assigned only the string in the first line.

Lnaden: Ah! yes, you are right, I'll cat it correctly when I get back to this

Lnaden:
I made the changes I feel comfortable with, but I have a few things I disagree with.

> `YankPhaseAnalyzer._prepare_mbar_input_data`

I am intentionally keeping that because the `_create_mbar_from_scratch` function should call that function.  I have documented this. I am trying to future proof a bit the class as other classes may need to do other things to prepare the mbar data which is unique, just like in prepping the mbar object, which may require additional work beyond just formatting the data.

Lnaden:
> What I meant is that _create_mbar_from_scratch is already a subclass-specific function

But it isn't? Both `_create_mbar_from_scratch` and `_prepare_mbar_input_data` are abstract methods of `YankPhaseAnalyzer`. I agree in the general sense that abstract classes should not enforce down to concrete classes, but this abstract class *needs* to do some very common things, no mater how the sampler changes. The problem is, the samplers would do this in such different ways at implementation that I could not come up with implementation beyond just the blanket subject of what they are doing, hence the two functions.

I should also note that the `_from_scratch` is method is really more meant for the case where the user does not know what they are doing, so its an auto-technique that should be hidden from the user working it on the back end.

Lnaden:
@andrrizzi and I have determined that we will merge this as is for now, because we  don't know how the SAMS sampler will work at the moment, but as that gets more crystalized, come back and retouch the abstract classes. Because they are internal, it should not affect the API.

PR #741, opened 2017-08-02
Small changes to repex API

andrrizzi:
Just small changes to the API that we discussed before releasing 1.0 . The 3 commits messages are pretty self-explanatory.

@Lnaden should I also rename the `checkpoint_storage_file` argument in `Reporter.__init__()` to `checkpoint_storage` to be consistent with the analysis one that is called `storage` (or do the opposite)?


Lnaden:
I think we should make it `checkpoint_storage` since we plan to eventually allow the reporter to take a `storage` object, your thoughts? Other than that there were some small things docs related I patched up but looks good.

andrrizzi:
> I think we should make it `checkpoint_storage`

Will do! And thanks for fixing the docs!

PR #743, opened 2017-08-08
Bugfix for multi-molecule mol2 files and allow to specify parameters in solvents

andrrizzi:
- When reading a multi-structure mol2 file with openeye, we assumed it contained different conformation of the same molecule. I've removed this assumption to handle different molecules in the same file.
- When there are multiple structures/residues in the mol2 file, `parmed` returns a `ResidueTemplateContainer` instead of a `ResidueTemplate` object. I've fixed `Mol2File.net_charge` and `Mol2File.resname(s)` to handle this case.
- I've added an optional `leap.parameters` key to the YAML `solvents` section. This is super useful to run different explicit solvent models that require different parameters files for both water molecules and ions.

I still have to update the YAML docs before merging. I opened a PR to check if tests pass.

### DIFF
--- a/Yank/pipeline.py
+++ b/Yank/pipeline.py
@@ -1328,6 +1328,7 @@ class SetupDatabase:
             molecule_parameters = self.molecules[mol_id]['leap']['parameters']
             tleap.load_parameters(*molecule_parameters)
 
+        tleap.load_parameters(*solvent['leap']['parameters'])
         tleap.load_parameters(*system_parameters)
 
         # Load molecules and create complexes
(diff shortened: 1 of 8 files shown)


Yank/pipeline.py
```diff
@@ -1347,7 +1348,7 @@ def _setup_system(self, system_file_path, pack, alchemical_charge,
                 for i, mol_id in enumerate(molecule_ids):
                     if mol_id not in self._pos_cache:
                         self._pos_cache[mol_id] = utils.get_oe_mol_positions(
-                                utils.read_oe_molecule(self.molecules[mol_id]['filepath']))
+                                utils.load_oe_molecules(self.molecules[mol_id]['filepath'], molecule_idx=0))
```
Lnaden: Do you mean to load the first index on every iteration? If I read this right, its loading multiple files, where we assume each has a single molecule in it (or at least is the first molecule). But if I'm wrong, then this behavior I don't think is right.

Brings the other question: how do we handle when the user has a mol2 file with multiple molecules they want to use?

andrrizzi: Before this step, we separate multi-molecules files into several files containing a single molecule according to the `select` options (that defaults to `all`). So when we get to this point, there is only 1 molecule in the file we're reading in both cases.

I'll work on the docs update tonight on the flight!

andrrizzi:
This should be good to go if tests pass. Besides updating the website docs, I've also fixed the OpenEye flavors we use to load mol2 files and I've pushed the fix for `extract_trajectory()` (and add a test for it).

PR #747, opened 2017-08-18
Add Langevin Splitting Integrator

Lnaden:
Uses the Splitting Integrator MCMC Move from OpenMMTools 0.13.0 as default integrator. Accepts None on splitting to go back to Native OpenMM Integrator


jchodera:
Awesome! Thanks!

Issue #771, opened 2017-09-11
Crash with automatic convergence detection


andrrizzi:
When `number_of_iterations` is `null` and a target statistical error is specified, YANK crashes at [this line](https://github.com/choderalab/yank/blob/master/Yank/experiment.py#L398) because it tries to subtract an integer from `None`.

I'll push a fix today.

Lnaden:
Ah, I missed this! Sorry. I have the check [here in repex](https://github.com/choderalab/yank/blob/0f7fc1ee268274f383cedb4783f8d68796bb148e/Yank/repex.py#L1538-L1539), but missed the check in experiment.

andrrizzi:
This was fixed in #773 .

Issue #769, opened 2017-09-05
Determine why Trailblaze protocol does not work in Explicit solvent


Lnaden:
The trailblaze protocol does not converge in explicit solvent. It times out no longer how long you give it. 

@lnaden needs to get the input files to @andrrizzi on our cluster to reproduce

andrrizzi:
This should have been solved in #773 . Feel free to reopen if it still happens.

Issue #766, opened 2017-08-31
Error when using automatic alchemical path determination


elkhoury:
As discussed in issue #749, we tried to compute the thermodynamic state spacing using the **auto** keyword, on two different complexes:
1- Cyclodextrin host/guest system. 
2- HSP90 protein/ligand system.

In both cases, YANK version 0.17.0 stops after the equilibration step.

We have attached input files to reproduce these cases.

Thank you,
Léa

(@davidlmobley, @renm1)
[harmonic_auto.tar.gz](https://github.com/choderalab/yank/files/1268391/harmonic_auto.tar.gz)

jchodera:
> In both cases, YANK version 0.17.0 stops after the equilibration step.

Can you elaborate on what you mean by "stops"? Does it crash? Does it hang?

elkhoury:
Please find attached the error files.

Thanks.
[auto_errors.tar.gz](https://github.com/choderalab/yank/files/1270969/auto_errors.tar.gz)

davidlmobley:
Looks like a different assertion error in each case -- so a type of crash, though it's very opaque as to what the underlying reason is without digging into the code.

jchodera:
For `hsp_90_80_harmonic_auto.e9577845`:
```
2017-08-28 18:27:21,111: Equilibration iteration 999/1000
2017-08-28 18:27:21,111: Propagating all replicas...
2017-08-28 18:27:21,111: Running _propagate_replica serially.
2017-08-28 18:27:23,558: Running _get_replica_move_statistics serially.
2017-08-28 18:27:23,558: Propagating all replicas took    2.446s
2017-08-28 18:27:23,558: Single node: executing <bound method ExperimentBuilder._generate_experiment_protocol.<locals>.DummyReporter.nothing of <yank.experiment.ExperimentBuilder._generate_experiment_protocol.<locals>.DummyReporter object at 0x2b634a0fb128>>
2017-08-28 18:27:39,547: Determining alchemical path for parameter lambda_electrostatics
Traceback (most recent call last):
  File "/home/elkhoury/anaconda3/bin/yank", line 11, in <module>
    load_entry_point('yank==0.17.0', 'console_scripts', 'yank')()
  File "/home/elkhoury/anaconda3/lib/python3.6/site-packages/yank/cli.py", line 71, in main
    dispatched = getattr(commands, command).dispatch(command_args)
[...]

PR #773, opened 2017-09-12
Various fixes for trailblaze scheme and loading local leap parameter dat file

### LINKED ISSUE #762: Auto-path function does not run in parallel
The auto-path function does not run in parallel. It throws an error about incorrect number of states between the alchemical path. Running a single process through either serial or through MPI does not throw the error.

The problem appears to be that the protocol generation procedure is not a blocking action, or its not being distributed correctly. For now users can run a 0 iteration serial to setup the process, then a production run with more iterations.

@andrrizzi I think the line [is here](https://github.com/choderalab/yank/blob/dfc99fb725c489be08e01cd785b1ff54b2395619/Yank/experiment.py#L1798) and I think it needs a `sync_nodes=True` flag. Is that right or do I need to do something else?

### LINKED ISSUE #763: Auto-path Output Cleanup
Running the `auto` alchemical path yields some interesting outputs in the final YAML file:

```yaml
 [1.0, !!python/object/apply:numpy.core.multiarray.scalar [
                        &id001 !!python/object/apply:numpy.dtype {args: [f8, 0, 1],
                            state: !!python/tuple [3, <, null, null, null, -1, -1,
                                0]}, !!binary "vqCl9usV6z8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "TnK3uCE65j8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "l5hgb0Yr3j8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "5zvxgmMR1j8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "2Wo5CkiTyT8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "RCke5h8+mj8=\n"], 0.0, 0.0, 0.0, 0.0, 0.0,
                    0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0]
                lambda_sterics: [1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "svANhh1b7T8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "siLm4VqD6j8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "Fp2ZkE3Z5z8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "XVMDJkDB5T8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "N7S4sgNA4z8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "5thvDVCG4D8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "hy2R1XQT3D8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "eo7QENm51z8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "1qpm3Fk40z8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "A4BDcsgYzz8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "nBndC2KyyD8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "NrN2pftLwj8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "bs1AF9TuuT8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "I1e5J8JrsT8=\n"], !!python/object/apply:numpy.core.multiarray.scalar [
                        *id001, !!binary "Xt0E9xLsoz8=\n"], 0.0]
```

While fully readable inside python, maybe not the most helpful thing for users. Probably just need to cast this to a list type output. 

You can read the file data with some python imports
```python
import yaml
from yank.experiment import YankLoader
import numpy as np

with open(YAMLFILE, 'r') as f:
   y = yaml.load(f, Loader=YankLoader)
y['protocols']['complex']['alchemical_path']
```

### LINKED ISSUE #772: Use infinity instead of None to specify unlimited number of iterations
Standard YAML supports infinity indicated as
```yaml
options:
  number_of_iterations: .inf
  online_analysis_target_error: 0.5
```
which get translated into a standard Python object `float('inf')`. What do you think of using that instead of `null`? It will make the internal handling in `Experiment` and `ReplicaExchange` easier too. I noticed a couple of places that may crash in few situations for reasons similar to #771.

andrrizzi:
Requires a new `openmmtools` release for tests to pass.

- Fix loading of leap parameters from a local `.dat` files (allow us to use local versions of gaff parameters for validation).
- Fix #772: Use infinity instead of `None` to specify unlimited number of iterations. Should fix also the crash described in #771 (testing this now).
- Fix #762: Trailblaze protocol crashes with MPI.
- Generalized restrained atoms selection during trailblaze scheme to include non-protein receptors (see also choderalab/openmmtools#290).
- Fixed bug when computing reduced potentials of simulated energies. This should solve the problem described in #766.
- Fix #763: Automatic path is saved in YAML as a mix of python and numpy floats.

This may solve also the problem in #769 even though I couldn't reproduce it so I'm not certain.


Yank/repex.py
```diff
@@ -1859,7 +1859,8 @@ def run(self, n_iterations=None):
             if (self._online_analysis_interval is not None and
                     self._iteration > self._online_analysis_minimum_iterations and
                     self._iteration % self._online_analysis_interval == 0):
-                self._run_online_analysis()
+                # Executed only by node 0 and broadcasted to be used in _is_completed().
+                self._last_err_free_energy = self._run_online_analysis()
```
Lnaden: Doesn't this assignment redundantly set the `self._last_err_free_energy`? or is it something to do with how MPI handles information?

andrrizzi: > or is it something to do with how MPI handles information?

Yes, exactly. Only node 0 calculates `_last_err_free_energy`, and then it broadcasts its value to the other processes so that they can check independently for completion.

andrrizzi:
Tagging @daveminh who may be interested in this.

andrrizzi:
Will do, thanks!

There's actually one last modification to the net charge part I want to do as Jason Swails answered me on the `parmed` issue tracker.

andrrizzi:
Done! Will merge when tests pass.

Lnaden:
Awesome! Thanks for that laundry list cleanup! Will this auto close all the issue it needs to (the couple in my list above)?

andrrizzi:
I'll close them manually and leave a comment, no worries.

Lnaden:
Great! Thanks again for your hard work on this!

andrrizzi:
No problem!

Issue #764, opened 2017-08-28
Site Documentation Cleanup


Lnaden:
Some random site cleanup things I am noticing going through the details. Most of these are just for clarity, critically incorrect information I will open a separate issue for so they can be resolved faster.

* [x] [Harmonic Restraints](http://getyank.org/0.17.0/algorithms.html#harmonic-restraints-harmonic) have conflicting information about whether the restraint should be fully coupled or decoupled at start (#778)
* [x] [Alcehm. Intermediates](http://getyank.org/0.17.0/algorithms.html#alchemical-intermediates) have some to-do's for filling in extra details. (#778)
* [x] [Ligand rot./displacement](http://getyank.org/0.17.0/algorithms.html#metropolis-monte-carlo-displacement-and-rotation-moves) should be inform users its not on for Boresch restraints.
* [x] "Identifying equilibration and production regions.." header in Algorithms RST header indicator line (`-------` or `====` under the text) is messed up. (#778)
* [x] Auto Protocol has 2 sections which should be combined in [here](http://getyank.org/0.17.0/algorithms.html#automatic-alchemical-protocol-selection) and [here](http://getyank.org/0.17.0/algorithms.html#autotuning-the-alchemical-protocol) (#778)
* [x] Add decision flow chart for what restraint to use (#782)
* [x] Resolve #765: additional restraint option documentation, see issue for discussion (#779)

This will be an ongoing and updated issue as I (or others) find new things

davidlmobley:
Also, is the new option for restraining to the center of a binding site in the docs yet? 

It would also be great if the really excellent stuff in the "health report" style Jupyter notebooks had either sample images of "good" and "bad" cases or links to where these could be found. (Maybe this is a separate issue). We were just using these for the first time and it wasn't totally clear to us initially what the mixing diagrams should ideally look like. After thinking about it for a while I figured it out, but I would think you'd want to make it easy for people to know whether these are "good" or "bad" or somewhere in between by providing concrete examples.

jchodera:
> Also, is the new option for restraining to the center of a binding site in the docs yet?

I don't think so, though the feature looks like it has been added. Can you add an issue suggesting we document this capability?

[...]

PR #778, opened 2017-09-18
Website docs laundry list

Lnaden:
This updates the website with missing information on the laundry list.

* Adds full description of all implicit solvent models
* Fixes a number of issues raised by #764

Only partial resolution on #764 as it does not add restraint flow chart, additional restraint documentation, or anything which would follow

@andrrizzi or @jchodera could either of you proof these changes? Specifically the solvent model science. If not, I can ask one of the other lab members to go over the sections in Slack for proofing.

### DIFF
--- a/docs/algorithms.rst
+++ b/docs/algorithms.rst
@@ -24,8 +24,8 @@ Implicit solvent
 
 YANK provides the somewhat unique ability to perform alchemical free energy calculations in implicit solvent.
 
-Solvent model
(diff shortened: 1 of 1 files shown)


andrrizzi:
In general, this looks good to me! I'll take a second pass at the math after this is merged and the LaTex has been rendered.

Lnaden:
@pgrinaway has reviewed this as well and said it looks good. Merging

PR #779, opened 2017-09-19
Add links to the individual Restraint API

### LINKED ISSUE #765: Document how to select flat bottom restraints size
Is there a mechanism to control the size of the flat bottom of the flat bottom restraining potential? We would very much like to be able to do this to assist with our benchmarking to compare the relative efficiency of flat bottom restraints, no restraints, and Boresch-style harmonic restraints. We've not been able to find anything in the docs on this.

It would also be useful to know how the size of the flat bottom potential is determined. 

cc @elkhoury

Lnaden:
The render looks like this where each item links to the API for the individual restraint (from screenshot)

restraints api

@davidlmobley is this what you were expecting for this page, or should there be more information to be clearer?

Fixes #765
Progress to #764

### DIFF
--- a/docs/yamlpages/experiments.rst
+++ b/docs/yamlpages/experiments.rst
@@ -1,3 +1,5 @@
+.. py:currentmodule:: yank.restraints
+
 .. _yaml_experiments_head:
 
 Experiments Header in YAML Files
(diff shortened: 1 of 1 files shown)


davidlmobley:
Looks better to me, yes. Thanks so much. I also pointed it out to @elkhoury and she'll get back to you if she sees any issues.

PR #780, opened 2017-09-19
Allow multiple experiments to keep running if one NaN's

### LINKED ISSUE #328: YANK termination for NaNs
It might be a good idea to have YANK terminate more gracefully and not print out pages of NaNs in the out put!

### LINKED ISSUE #616: Have multi-system simulations proceed to next system if one encounters a NaN
Currently, encountering a NaN in a single simulation will take out the entirety of a YANK run, even if we want to grind through dozens of different complexes. There may [still be some issues with the OpenMM 7.1 beta and NaNs](https://github.com/pandegroup/openmm/issues/1695), so let's add some logic to catch this exception and keep running through the rest of the phases and systems when this occurs.

Lnaden:
Enables multiple and/or combinatorial experiments to run if one of them goes to NaN.
Introduces a special error which can be trapped and sent up the stack from a sampler through the phase and experiment. Should also print out to the end of the logger a warning to the user if any NaN.

~This has NOT been tested yet and should NOT be merged, but I want to get the implementation discussion going while I figure out the correct way to test this. I'll strike this line and remove the [DNM] flag once its formally tested.~

*edited:* Added the test so this can be merged when everything passes now.

Fixes #328, Fixes #616


Yank/repex.py
```diff
@@ -2101,7 +2101,7 @@ def _propagate_replica(self, replica_id):
             prefix = 'nan-error-logs/iteration{}-replica{}-state{}'.format(
                 self._iteration, replica_id, thermodynamic_state_id)
             e.serialize_error(prefix)
-            raise
+            raise utils.SimulationNaNError
```
andrrizzi: I think you're returning the class here without instantiating.

Yank/experiment.py
```diff
@@ -2278,7 +2307,11 @@ def _run_experiment(self, experiment):
             switch_experiment_interval = self._options['switch_experiment_interval']
 
         built_experiment = self._build_experiment(experiment_path, experiment)
-        built_experiment.run(n_iterations=switch_experiment_interval)
+        # Trap a NaN'd simulation by capturing only the error we can handle, let all others raise normally
+        try:
+            built_experiment.run(n_iterations=switch_experiment_interval)
+        except utils.SimulationNaNError as e:
+            return e
```
andrrizzi: This may be redundant. The function should propagate the exception even without an explicit `try...except` clause.

Yank/experiment.py
```diff
@@ -400,7 +400,12 @@ def run(self, n_iterations=None):
 
                 # Run simulation for iterations_left or until we have to switch phase.
                 iterations_to_run = min(iterations_left[phase_id], switch_phase_interval)
-                alchemical_phase.run(n_iterations=iterations_to_run)
+                try:
+                    alchemical_phase.run(n_iterations=iterations_to_run)
+                except utils.SimulationNaNError:
+                    # Simulation has NaN'd, this experiment is done, flag phases as done and send error up stack
+                    self._are_phases_completed = [True] * len(self._are_phases_completed)
```
andrrizzi: I'm not sure this will work. The `Experiment` instance is rebuilt every time you call `ExperimentBuilder._run_experiment()`, so at the next iteration this variable will still try to run.

This is not a problem if `switch_experiment_interval` is not set anyway, and it's a performance issue rather than a functional one.

Yank/experiment.py
Lnaden: No, because here I `return e` not `raise e`, the former allows `run_experiments` to accept the error as a manipulatable object where as the later would just raise the error where it would have to be trapped up above too. I had thought about the `try..except` clause in `run_experiments`, but there was no way for me to trap the MPI distributed instance of it since each one could raise its own error. 

This block also ensures that any non `SimulationNaNError`s are raised as normal

Yank/experiment.py
Lnaden: Ah, you're right, I'll need to think of another way to block the experiment. I don't want to spend time on experiments which have failed, so I'll think about this some more

Yank/repex.py
Lnaden: You're right, the `raise` statement actually does instance the `Exception` class technically, but is not the PEP8/Pyhtonic way to do it, so I'll instance it.

Yank/experiment.py
andrrizzi: Ah, I understand now, thanks! That gives me an idea though, what do you think of making `mpi.distribute(catch_exceptions=True)` catching and returning the exception instead of the task result if something goes wrong? I don't think it'll be too hard to add.

Lnaden: I'm not sure I follow, if any of the MPI jobs fail, wont the error propagate up the stack instead of the task returning a proper task result? What would a caught exception do?

Lnaden:
Suggestion from @jchodera:
> Should we write unrecoverable errors to a special log file?
I'm just worried the normal output log will just have these unrecoverable errors getting buried in a sea of useless output
We also probably want to write out key diagnostic information (like a uniquely labeled archive of serialized XML files) to help us debug what went wrong

Thats not a bad idea actually. Right now I have the multi-experiment NaN writing the critical information at the very end of the log file, but I think having it has its own log may better. This would be the issue to fix it on as well so I'll look into that.

Lnaden:
Also from @jchodera 
> BTW, hoping to write anything to the "very end" of the log file is going to fail when your job is killed early by the scheduler hitting your wall clock time or a power failure
That's not a good solution in general as a result

Not something I had thought of, should be mitigated by adding a critical logger which writes out in real-time

Lnaden:
- Implement critical logger which will create a new file for logs to go to as needed.
- Remove experiments from the set of `all_experiments` as they complete (will need functional review, @andrrizzi, to make sure I did not break something else)
- Handle NaN writing at experiment and repex level as they are caught, might be a bit overkill, but I'd rather have the case where the modules are run standalone (api) than miss it.

Still need to figure out tests, still not ready to merge. Besides that: thoughts on the new implementation?

Yank/experiment.py
Lnaden: I decided to leave this check in place in case something besides the experiment builder runs the experiment (e.g. through the API) because we do expose Experiment as a public class. The `run_experiments` though now pops finished experiments out so they should not be re-run or even re-built.

Yank/experiment.py
```diff
@@ -700,16 +705,24 @@ def run_experiments(self):
...
                 else:
-                    completed, experiment_indices = mpi.distribute(self._run_experiment,
-                                                                   distributed_args=all_experiments,
-                                                                   group_nodes=processes_per_experiment)
+                    completed = mpi.distribute(self._run_experiment,
+                                               distributed_args=all_experiments,
+                                               group_nodes=processes_per_experiment,
+                                               send_results_to='all')
+                # Remove any completed experiments, releasing possible parallel resources to be reused
+                for exp_index in range(len(all_experiments)):
```
andrrizzi: You may have to perform the loop in reverse order or you'll delete the wrong experiments when multiple are completed.

Yank/experiment.py
```diff
@@ -700,16 +705,24 @@ def run_experiments(self):
 
             # Cycle between experiments every switch_experiment_interval iterations
             # until all of them are done.
-            completed = [False for _ in range(len(all_experiments))]
-            while not all(completed):
+            nand = [False for _ in range(len(all_experiments))]
```
andrrizzi: Unused?

Lnaden:
I still want to make tests for this feature before I merge it. Even if its a "slow" test we can only check manually and not on travis

Lnaden:
@andrrizzi I have added a couple tests for this now, could you do a quick once over of them? Its about 10 lines in the `test_experiment` file.

Yank/tests/test_experiment.py
```diff
@@ -200,6 +205,88 @@ def get_template_script(output_dir='.'):
...
+      system: premade_nan
+      protocol: absolute-binding
+      restraint:
+        type: FlatBottom
+
+    experiments: [{repeating}]
+
+    """
+    repeating_string = ', '.join(['the_exp'] * experiment_repeats)
+    repeating_nan_string = ', '.join(['the_nan_exp'] * number_nan_repeats)
```
andrrizzi: I believe this has no effect. The experiments are collected into a dictionary inside `ExperimentBuilder`, which means that there will be only 1 `the_exp` and 1 `the_nan_exp` regardless of how many times you specify them in the `experiments` list.

Lnaden:
Ah, well. Still works for the tests I wrote, once I get the tests working.

Lnaden:
Well, I had to add a trap for a special edge case which I invoke with the test that was not properly trapped. Happens under the following conditions:

* Iteration 0
* Linux, not OSX
* A starting coordinate is NaN

Lnaden:
Sweet merciful passing tests, this was much tricker than it should have been. But now we can have simulations NaN with grace and not halt everything, as well as put out more helpful messages.

jchodera:
Awesome, thanks!

PR #783, opened 2017-09-22
Enable Gaff2 support in antechamber

### LINKED ISSUE #610: Add support for gaff2
see http://archive.ambermd.org/201605/0281.html

Lnaden:
Should fix #610 since we already support Gaff2 if its in the parameter list

### DIFF
--- a/Yank/pipeline.py
+++ b/Yank/pipeline.py
@@ -1253,10 +1253,13 @@ class SetupDatabase:
                 # Generate parameters
                 charge_method = mol_descr['antechamber']['charge_method']
                 input_mol_path = os.path.abspath(mol_descr['filepath'])
+                # Use Gaff in parameters, otherwise default to gaff2
+                gaff = 'gaff' if 'leaprc.gaff' in mol_descr['leap']['parameters'] else 'gaff2'
                 with moltools.utils.temporary_cd(mol_dir):
                     moltools.amber.run_antechamber(mol_id, input_mol_path,
                                                    charge_method=charge_method,
-                                                   net_charge=net_charge)
+                                                   net_charge=net_charge,
+                                                   gaff_version=gaff)
 
                 # Save new parameters paths
                 mol_descr['filepath'] = os.path.join(mol_dir, mol_id + '.gaff.mol2')
--- a/Yank/pipeline.py
+++ b/Yank/pipeline.py
@@ -1254,7 +1254,7 @@ class SetupDatabase:
                 charge_method = mol_descr['antechamber']['charge_method']
                 input_mol_path = os.path.abspath(mol_descr['filepath'])
                 # Use Gaff in parameters, otherwise default to gaff2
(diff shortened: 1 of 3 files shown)


Lnaden:
This should also make Gaff2 the default parameterization

jchodera:
Don't forget to go back to the validation set and explicitly specify the original GAFF parameter set in all of those validation calculations that are trying to match GAFF 1.x results!

Actually, should we even add the ability to specify a *specific* version of GAFF?

Lnaden:
So long as there exists a 
```
leap:
    parameters: [leaprc.gaff]
```
entry, the gaff 1.x version will be used. All the validation .yaml file call this.

> should we even add the ability to specify a specific version of GAFF

This would be very hard to do I think since it would require us to fetch the specific GAFF version from somewhere online through some arcane specification. We have supported gaff2 for a while (just add it to the `parameters` entry), but it was not passed down to antechamber before this PR.

jchodera:
> This would be very hard to do I think since it would require us to fetch the specific GAFF version from somewhere online through some arcane specification. We have supported gaff2 for a while (just add it to the parameters entry), but it was not passed down to antechamber before this PR.

We'd store the gaff versions in the Python installation, and simply stage the appropriate file as needed. This may not provide much more flexibility than the current filename-based scheme, but if you think about whether it will cause a future support nightmare explaining how to find and use specific versions of gaff.xml for comparisons, it may be simpler.

Yank/pipeline.py
```diff
@@ -1253,10 +1253,13 @@ def _setup_molecules(self, *args):
                 # Generate parameters
                 charge_method = mol_descr['antechamber']['charge_method']
                 input_mol_path = os.path.abspath(mol_descr['filepath'])
+                # Use Gaff in parameters, otherwise default to gaff2
+                gaff = 'gaff' if 'leaprc.gaff' in mol_descr['leap']['parameters'] else 'gaff2'
```
andrrizzi: We had a discussion with David Mobley about this, and he feels strongly about preferring gaff against gaff2, so I'm not sure which one we should use as default.

Lnaden:
> We'd store the gaff versions in the Python installation, and simply stage the appropriate file as needed. 

This sounds like something we can address in a separate issue

>  David Mobley about this, and he feels strongly about preferring gaff against gaff2, so I'm not sure which one we should use as default.

The default is easy enough to fix, it just reverses the logic on the operator I added. I would defer to David on that if he has a particularly strong stance. Either way, we will still *have* gaff2 support which is the issue this PR was suppose to fix. I'll reverse it now

andrrizzi:
Actually, I've just realized that that default doesn't change much because we're not adding the `leaprc.gaff` or `leaprc.gaff2` parameters by default, so either way it's fine, sorry.

Lnaden:
Yes, but if someone pulls small molecule parameters from somewhere else, but wants to get missing things from antechamber, the original way would have used `gaff2`. I realize those conditions yield a corner case, but still one to consider. "Default" is a very dangerous word in programming, so I want to make sure we default is not a conditional action

Issue #777, opened 2017-09-16
Storage of data in complex.nc vs complex_checkpoint.nc


jchodera:
I think we somehow got confused about which data goes into which file.

Here is the intended division of data:
* Analysis file: This should contain the data we need to estimate free energies (energies, states, statistics), diagnose convergence problems, assess performance (timestamps), and visualize the protein:ligand binding ensemble (solute-only trajectories). It should be small and easy to share to help users debug their problems.
* Everything else: This file contains literally everything else: Metadata, serialized systems, full-system checkpoints, etc.

Instead, we have:
* The `complex_checkpoint.nc` file contains:
```
variables:
	int64 last_iteration(scalar) ;
	float positions(iteration, replica, atom, spatial) ;
	float box_vectors(iteration, replica, spatial, spatial) ;
	double volumes(iteration, replica) ;
	string timestamp(iteration) ;
```
* The `complex.nc` file contains:
```
variables:
	int64 last_iteration(scalar) ;
	string options(scalar) ;
	string metadata(scalar) ;
	int states(iteration, replica) ;
	double energies(iteration, replica, replica) ;
	double unsampled_energies(iteration, replica, unsampled) ;
	int accepted(iteration, replica, replica) ;
	int proposed(iteration, replica, replica) ;
	string timestamp(iteration) ;
```
I think the serialized states are stored somewhere in the `options` or `metadata` and probably need to be moved to what you call the `checkpoint` file, while the analysis file is missing the solute-only trajectory.

PR #784, opened 2017-09-25
Add solute-only trajectory storage

Lnaden:
This PR adds the ability for YANK to store only the solute trajectory as part of the analysis file. Minimal changes to the API calls in the form of new keywords; no new public methods nor changes to existing calls which would break backwards compatibility.

This is constructed automatically from the Topography if the appropriate flag is set in the YAML file, which defaults to True.

There are no new tests for this at the moment, which we may want to write before merging, but I at least want to get the conversation started and the code reviewed.

This has also changed the logic in the trajectory analysis code, so that will need to be verified (again, tests would fix)

Progress towards #777

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -1545,9 +1545,19 @@ def extract_trajectory(nc_path, nc_checkpoint_file=None, state_index=None, repli
         logger.info('Detected periodic boundary conditions: {}'.format(is_periodic))
 
         # Get dimensions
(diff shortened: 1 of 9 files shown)


Lnaden:
The functional part of the code is ready for review. Ideas for tests welcome

Lnaden:
Thanks @andrrizzi, I have one more test I thought of as a corner case I am adding, I'll merge this after I get that in and it passes.

PR #786, opened 2017-10-02
Cut Yank 0.18.0

Lnaden:
Due to official Python 2 support dropped, a new major revision is being cut.

Also taking advantage of the new feature in setuptools/pip for `python_requires` + the conda-recipe to force python 3.5 and up. The additional keyword should not cause an issue for pip < 9.0.1

Tests must pass before we can merge this to test the new requirements feature


Lnaden:
Hold off on merging this until we can track down the travis segfault

PR #788, opened 2017-10-03
Dummy PR for Travis Testing [DNM]

Lnaden:
This PR is only to trigger travis and chase down a stochastic failure

**DO NOT MERGE THIS**


Lnaden:
I think I found the issue, it appears to be a problem where the combination of nose, travis, and netcdf is not properly closing the NetCDF file before trying to reopen it, causing a segfault... Because thats stable (I blame HDF and NetCDF).

I pushed the change to #786 and will be deleting the dummy branch here

PR #793, opened 2017-10-10
Add new Regions API to Topography

Lnaden:
Add the ability to add and manipulate arbitrary regions to the Topography object. These are more general than the alchemy based alchemical regions and are only subsets of atoms.

Generalizes string selection handling with a future PR to handle SMARTS strings (as marked by TODO's). This PR does not address those so its a smaller chunk to review.

Adds a handy multitmethod technique to reduce cyclomatic complexity we may be able to apply elsewhere.

### DIFF
--- a/Yank/tests/test_yank.py
+++ b/Yank/tests/test_yank.py
@@ -70,15 +70,38 @@ def test_topography():
     assert len(topography.ions_atoms) == 0
 
 
+def test_topography_regions():
+    """Test that topography regions are created and fetched"""
+    toluene_vacuum = testsystems.TolueneVacuum()
+    topography = Topography(toluene_vacuum.topology)
+    # Should do nothing and return nothing without error
+    assert topography.remove_region('Nothing') is None
+    topography.add_region('A hard list', [0, 1, 2])
+    assert 'A hard list' in topography
+    topography.add_region('A junk list', [5, 5, 5])
+    topography.remove_region('A junk list')
+    assert 'A junk list' not in topography
+    assert topography.get_region('A hard list') == [0, 1, 2]
+    # Confirm that string typing is handled
+    topography.add_region('carbon', 'name C')
+    assert len(topography.get_region('carbon')) > 0
+    with nose.tools.assert_raises(ValueError):
+        topography.add_region('failure', 'Bad selection string')
+    # Ensure region was not added
+    assert 'failure' not in topography
+
+
 def test_topography_serialization():
     """Correct serialization of Topography objects."""
     test_system = testsystems.AlanineDipeptideImplicit()
     topography = Topography(test_system.topology)
+    topography.add_region('atest', [0, 1, 2, 3])
     serialized_topography = mmtools.utils.serialize(topography)
     restored_topography = mmtools.utils.deserialize(serialized_topography)
     assert topography.topology == restored_topography.topology
     assert topography.ligand_atoms == restored_topography.ligand_atoms
     assert topography.solvent_atoms == restored_topography.solvent_atoms
+    assert topography.get_region('atest') == restored_topography.regions['atest']
 
 
 # ==============================================================================
--- a/Yank/tests/test_yank.py
+++ b/Yank/tests/test_yank.py
@@ -83,7 +83,7 @@ def test_topography_regions():
     assert 'A junk list' not in topography
     assert topography.get_region('A hard list') == [0, 1, 2]
     # Confirm that string typing is handled
-    topography.add_region('carbon', 'name C')
+    topography.add_region('carbon', 'element C')
     assert len(topography.get_region('carbon')) > 0
     with nose.tools.assert_raises(ValueError):
         topography.add_region('failure', 'Bad selection string')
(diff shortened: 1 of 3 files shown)


Yank/yank.py
```diff
@@ -88,6 +88,10 @@ def __init__(self, topology, ligand_atoms=None, solvent_atoms='auto'):
         self.solvent_atoms = solvent_atoms
         self.ligand_atoms = ligand_atoms
 
+        # Initialize regions
+        self._built_in_regions = ['ligand_atoms', 'receptor_atoms', 'solute_atoms', 'solvent_atoms', 'ion_atoms']
```
andrrizzi: It could make sense to define this as a private class-level constant since it's never changed.

Yank/yank.py
```diff
@@ -179,6 +183,66 @@ def ions_atoms(self):
...
+        Returns
+        -------
+
+        Raises
+        ------
+        KeyError
+            If region is not part of the Topography
+        """
+        if region_name not in self._combied_regions:
+            raise KeyError("Cannot find region \"{}\" in this Topography.".format(region_name))
```
andrrizzi: I think this `if...raise` should go after the check for built-in-regions below.

Yank/yank.py
```diff
@@ -202,17 +267,54 @@ def __setstate__(self, serialization):
...
-    def _resolve_atom_indices(self, atoms_description):
-        if isinstance(atoms_description, str):
+    def _check_existing_regions(self, region_string):
+        """Make sure regions don't overlap"""
+        if region_string in self:
+            raise KeyError("{} is already part of this Topology! "
+                           "Cannot overwrite built-in regions!".format(region_string))
+
+    @property
+    def _combied_regions(self):
```
andrrizzi: Typo `combied` -> `combined`?

Yank/utils.py
```diff
@@ -252,6 +252,66 @@ def _wrapper(*args, **kwargs):
 
 
 # =======================================================================================
+# Multimethod wrapping to better handle cyclomatic complexity with input types
+# See Adam Bard's post: https://adambard.com/blog/implementing-multimethods-in-python/
```
andrrizzi: This is pretty cool! But I don't understand the difference with the `functools.singledispatch()` that is in the standard library.

Yank/tests/test_yank.py
```diff
@@ -70,15 +70,38 @@ def test_topography():
     assert len(topography.ions_atoms) == 0
 
 
+def test_topography_regions():
+    """Test that topography regions are created and fetched"""
+    toluene_vacuum = testsystems.TolueneVacuum()
+    topography = Topography(toluene_vacuum.topology)
+    # Should do nothing and return nothing without error
+    assert topography.remove_region('Nothing') is None
```
andrrizzi: I'm not sure if this is what you wanted, but currently, the `remove_region` method doesn't return anything, so the test would always pass.

Yank/yank.py
Lnaden: Is it not already a private class-level constant since its `self._name`? Or do you mean at `class name(); X=1, Y=2`? level

Yank/yank.py
Lnaden: Why? The `_combined_regions` property is the union of the `built-in` and the custom `regions`? I could make it it try to get the built in, then try to get the custom and raising an error if it can't find it.

Yank/utils.py
Lnaden: While that is a viable solution in normal functions, the problem is that `singledispatch` does not work on class methods, since the first argument is always `self` and not the arguments being passed as the first arg to run type checking on. The multi method is a bit more general, albeit requires to be implemented as its own methods

Yank/tests/test_yank.py
Lnaden: It is what I wanted. I wanted the test that trying to remove a non-existing region does nothing and does not return an error. The `assert` ensures we in fact are getting the correct output and the call itself ensures this corner case is tested

Yank/yank.py
andrrizzi: I mean it should be a constant of the class, not of the object.

Yank/yank.py
andrrizzi: Oh, right! Sorry for the confusion.

Yank/utils.py
andrrizzi: Ah, I understand now, thanks! Very cool! What do you think about using a wrapper like this instead: https://stackoverflow.com/a/24602374 This way we could have a single wrapper and keep the usual `singledispatch` syntax.

Yank/yank.py
```diff
@@ -202,17 +266,55 @@ def __setstate__(self, serialization):
...
+        """
+        Return the combined set of regions
+        This is its own property despite its simplicity since several functions and methods call it
+        """
+        # Iterate over the dict_keys object since its a static view and should be thread-safe
+        return self._built_in_regions + [region for region in self._regions.keys()]
+
+    def __contains__(self, item):
+        """Check the in operator to see if region is in this class"""
+        return item in self._combined_regions
```
andrrizzi: This is fine since `_combined_regions` should be very small. A cleaner alternative here could be
```python
return item in self._regions or item in self._BUILT_IN_REGIONS
```
This way the search would be based on the hash instead of instantiating and searching through a list.

Yank/yank.py
andrrizzi: You could also use
```python
if region_name not in self:
```
since you implemented a `__contains__` function

Yank/utils.py
Lnaden: Hmmm. I saw that and wondered how general it is. I'll give it a try and see what comes of it.

Yank/yank.py
Lnaden: Okay, there was another place I had `_combined_regions` but I suppose I can put them into the same `in` comparison now that I have implementation done.

PR #794, opened 2017-10-11
Arbitrary Molecular Regions in YAML

Lnaden:
This PR adds arbitrary region identification to the YAML documentation in continuation of the abilities added by #793.

Regions can be defined for any `molecules` entry and as many as you would like can be defined. These regions are validated to accept ints, strings, or lists of ints, although only strings is non-ambiguous and even improvements could be made upon that as well, but the framework is here.

~Although functionally complete, this is a Work In Progress for the following reasons:~ *edit*: no longer a WIP after additions.
- [x] Error message when `check_regions_clash` raises a `YamlParseError` is absolute garbage and falls back to an incorrect error message, this is a larger problem with YamlParseError in general though 
- [x] Documentation on the `docs` pages have not been added yet. 

The one functional improvement we should make is to have each molecule's region's only apply to the subset of atoms that molecule makes up in the final Topography. I think this would require a small change to the `add_region` call of Topography + some logic in `_build_experiment`, but may not be too hard. Debate on whether we make that change here or in another PR after we merge this one since this feature set is suppose to be a series of smaller PR chunks, each adding new infrastructure an leading up to the full feature. Thoughts?

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -2181,15 +2181,15 @@ class ExperimentBuilder(object):
         try:  # binding free energy calculations
             solvent_ids = [db_system['solvent'],
                            db_system['solvent']]
-            ligand_regions = self._db.molecules.get(db_system.get('ligand', {}), {}).get('regions', {})
-            receptor_regions = self._db.molecules.get(db_system.get('receptor', {}), {}).get('regions', {})
+            ligand_regions = self._db.molecules.get(db_system.get('ligand'), {}).get('regions', {})
+            receptor_regions = self._db.molecules.get(db_system.get('receptor'), {}).get('regions', {})
             # Name clashes have been resolved in the yaml validation
             regions = {**ligand_regions, **receptor_regions}
         except KeyError:  # partition/solvation free energy calculations
             try:
                 solvent_ids = [db_system['solvent1'],
                                db_system['solvent2']]
-                regions = self._db.molecules.get(db_system.get('solute', {}), {}).get('regions', {})
+                regions = self._db.molecules.get(db_system.get('solute'), {}).get('regions', {})
             except KeyError:  # from xml/pdb system files
                 assert 'phase1_path' in db_system
                 solvent_ids = [None, None]
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -1092,6 +1092,12 @@ class ExperimentBuilder(object):
                 return True
             return False
 
+        def regions_match_format(region_dict):
+            """Regions are formatted correctly"""
+
+        regions_schema = {Optional(str): Or(list, str)}
+
+
         validated_molecules = molecules_description.copy()
 
         # Define molecules Schema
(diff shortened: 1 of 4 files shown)


Yank/experiment.py
```diff
@@ -1351,6 +1363,25 @@ def _system_files(files):
...
+            Check that the regions have non clashing names by looking at regions in each molecule for a given
+            system
+            """
+            mol_description1 = self._db.molecules[system_dict[mol_class1]]
+            mol_description2 = self._db.molecules[system_dict[mol_class2]] if mol_class2 is not None else {}
+            # Fetch the regions or an empty dict
+            regions_1_names = mol_description1.get('regions', {}).keys()
+            regions_2_names = mol_description2.get('regions', {}).keys()
+            for region_1_name in regions_1_names:
+                if region_1_name in regions_2_names:
```
andrrizzi: You could use the `set.intersection()` operation here to check for repeated regions names. I think just doing `set(mol_description1.get('regions', {}))` should do the trick.

Yank/experiment.py
```diff
@@ -1359,12 +1390,14 @@ def _system_files(files):
...
+            And({'receptor': is_known_molecule, 'ligand': is_known_molecule,
+                 'solvent': is_pipeline_solvent, Optional('pack', default=False): bool,
+                 Optional('leap'): self._LEAP_PARAMETERS_SCHEMA},
+                lambda d: check_regions_clash(d, 'ligand', 'receptor')),
 
-            {'solute': is_known_molecule, 'solvent1': is_pipeline_solvent,
-             'solvent2': is_pipeline_solvent, Optional('leap'): self._LEAP_PARAMETERS_SCHEMA},
+            And({'solute': is_known_molecule, 'solvent1': is_pipeline_solvent,
+                'solvent2': is_pipeline_solvent, Optional('leap'): self._LEAP_PARAMETERS_SCHEMA},
+                lambda d: check_regions_clash(d, 'solute')),
```
andrrizzi: I'm not sure I understand. If we have only one molecule, is it possible to have region name clashes at all?

Yank/experiment.py
```diff
@@ -2143,17 +2176,24 @@ def _build_experiment(self, experiment_path, experiment):
             solvent_dsl = 'auto'  # Topography uses common solvent resnames.
         logger.debug('DSL string for the solvent: "{}"'.format(solvent_dsl))
 
-        # Determine complex and solvent phase solvents
+        # Determine complex and solvent phase solvents while also getting regions
+        db_system = self._db.systems[system_id]
```
andrrizzi: Could you maybe call this `system_description`? I think that's the common variable naming convention in the class.

Yank/experiment.py
Lnaden: I could, but I want the error message to be helpful, although thats currently having problems since the error stack is garbage at this point.

Yank/experiment.py
Lnaden: No, but I imagine it would not be hard to extend this formalism to allow regions defined in the *system* as well, if so, we will want to keep this framework here. Its a pretty cheap action but I could remove it if you think it would be best

andrrizzi: Ah, I see. Yeah, just leave it there then.

Lnaden:
I think I know what is happening with the errors, but its not something to be fixed here, see #795 for discussion. Other than that, I think this is ready.

PR #796, opened 2017-10-13
Add in Region selected atoms to the Restraint class

Lnaden:
This PR adds a number of changes to the restraint functionality as continuation of #793 and #794

This PR adds the ability for Restraint classes to accept Topography Region-based logic as part of the selection string, as well MDTraj strings. This also extends the Topography to handle complex region selection through the `and` and `or` operators.

Main logical question: Should we *remove* the ability for DSL strings to be passed as direct atom selections in Restraints and force it to use Topography Regions? That would clean up the code, but then offload the DSL processing exclusively to Topology and remove the functionality from restraints as a standalone.

There is a break to the `RadiallySymetricRestraint` class API which went unused for the most part, but still a break so version updated to 0.19.0.

~This is a WIP for the following reasons, although the functionality changes should all be in now, allowing discussion and review~ No longer a WIP as all items have been checked off.
* [x] OpenMMTools 0.13.2 is required (not released)  due to extended `math_eval`, tests will not run until thats released
 * [x] I'm considering extending the Schema validation for restraints to ensure Regions are previously defined, but that will be hard considering that the current Schema accepts any input so arbitrary parameters can be passed. Not sure how to check
 * [x] The tests should be extended to test the Region selection code
 * [x] The logic should be reviewed to ensure I am not missing corner cases
 * [x] The Restraints doc-strings need updated
 * [x] The website docs need the "whats new" updated

 Next on the list is how do we parse SMILES strings

### DIFF
--- a/Yank/restraints.py
+++ b/Yank/restraints.py
@@ -25,6 +25,7 @@ import math
 import random
 import inspect
 import logging
+import functools
 import itertools
 
 import numpy as np
@@ -34,6 +35,7 @@ import openmmtools as mmtools
 from simtk import openmm, unit
 
 from . import pipeline
+from .utils import methoddispatch
 
 logger = logging.getLogger(__name__)
 
@@ -451,9 +453,6 @@ class RadiallySymmetricRestraint(ReceptorLigandRestraint):
         This can temporarily be left undefined, but :func:`determine_missing_parameters`
         must be called before using the Restraint object. The same if a DSL
         expression is provided without passing the topology (default is None).
-    topology : simtk.openmm.app.Topology or mdtraj.Topology, optional
-        The topology used to resolve the DSL string. If not provided, the atom
-        selection expressions will be resolved by :func:`determine_missing_parameters`.
 
     Attributes
     ----------
@@ -476,10 +475,9 @@ class RadiallySymmetricRestraint(ReceptorLigandRestraint):
         function to automatically determine these parameters from the atoms positions.
 
     """
-    def __init__(self, restrained_receptor_atoms=None, restrained_ligand_atoms=None, topology=None):
+    def __init__(self, restrained_receptor_atoms=None, restrained_ligand_atoms=None):
         self.restrained_receptor_atoms = restrained_receptor_atoms
         self.restrained_ligand_atoms = restrained_ligand_atoms
-        self.topology = topology
 
     # -------------------------------------------------------------------------
     # Public properties.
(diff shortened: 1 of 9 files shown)


Lnaden:
After much debate with @andrrizzi, this is going to get refactored again to handle all atom selection through `Topography`. More commits will be needed and for the time being I will close the PR, repoen it at a future point.

Lnaden:
The new `Topography.select` function is a workhorse function with its own internal extendable classes for handling selection strings while trying to minimize cyclomatic complextiy. The restraint selection logic has been relegated mostly to the Topography selection.

Because this is mostly a rework of the internal, this should be ready for review, @andrrizzi and/or @jchodera

Yank/yank.py
```diff
@@ -245,6 +251,237 @@ def get_region(self, region_name):
...
+        selection : str or
+            String defining the selection
+        as_set : bool, Default False
+            Determines output format. Returns a Set if True, otherwise output is a list.
+        sort_by : str or None, Default: 'auto'
+            Determine how to sort the output if ``as_set`` is False.
+
+            * 'auto': Let the selection string determine how to sort it out based on its priorities
+            * 'ordered': Atoms are sorted index, smallest to largest
+            * 'inferred_by_region': Atoms are sorted by which region in the provided ``selection_string`` occurs first.
```
andrrizzi: I would call these two something like `sorty_by='index'` and `sorty_by='region_order'`.

Yank/yank.py
```diff
@@ -245,6 +251,237 @@ def get_region(self, region_name):
...
+            no effect.
+
+        """
+
+        # Helper functions for handling the sorting
+        def sort_output_ordered(sortable):
+            # Dont do list.sort, its an in place action.
+            return sorted(list(sortable))
+
+        def sort_output_inferred_by_region(sortable):
```
andrrizzi: I had some troubles following the logic here. Why not something like
```python
def sort_output_inferred_by_region(sortable):
    # Only valid when selection is a string
    final_output = []
    region_order = [region_name for region_name in selection.split() if region_name in self]
    for region_name in region_order:
        region = self.get_region(region_name)
        for atom_index in region:
            if atom_index not in final_output:
                final_output.append(atom_index)
```

Yank/yank.py
```diff
@@ -245,6 +251,237 @@ def get_region(self, region_name):
...
+        registered_selectors = selector_picker(selection)
+
+        selector_outputs = []
+        selector_errors = []
+        selector_names = []  # For handling error messages
+        for selector in registered_selectors:
+            selector_names.append(selector.__name__)
+            selection_output, errors = selector.select(selection)
+            selector_outputs.append(selection_output)
+            selector_errors.append(errors)
```
andrrizzi: The list is populated but unused afterwards.

Yank/yank.py
```diff
@@ -245,6 +251,237 @@ def get_region(self, region_name):
...
+            pass
+
+        # Dispatcher to parse the selection type and return the valid selection classes
+        @functools.singledispatch
+        def selector_picker(selection_input):
+            try:
+                # Catch both the numpy and built in integers
+                assert all([isinstance(i, np.integer) or isinstance(i, int) for i in selection_input])
+            except AssertionError:
+                raise AssertionError("Selection {} is not iterable of ints or any other readable type such as string! "
```
andrrizzi: I think you should use either a simple
```python
if not all(condition):
    raise ValueError()
``` 
or at most
```python
assert all(condition), 'Selection {} is not iterable ...'
```
`assert`s can be deactivated when calling the python interpreter.

Yank/yank.py
Lnaden: Ah, good catch, this was suppose to be part of a helpful error message, I added that.

Yank/yank.py
Lnaden: Very helpful, my brain was stuck on the more complex solution because of something I had to do elsewhere

Yank/yank.py
Lnaden: I am trying to keep cyclomatic complexity down, and this lets us much more easily extend the function later. I had actually thought about splitting the selection/sorting functionality into its own class, but that did not make sense since its only used in this function.

Lnaden:
@andrrizzi Thanks for the review! Found lots of things I missed. I did have a few questions which I put in the individual comments. One of them got buried in the "outdated" categories about the `singledispatch` in the `select` method.

Yank/restraints.py
```diff
@@ -731,6 +742,15 @@ def _determine_bond_parameters(self, thermodynamic_state, sampler_state, topogra
     # Internal-usage
     # -------------------------------------------------------------------------
 
+    @property
+    def _are_restrained_atoms_defined(self):
+        """Check if the restrained atoms are defined well enough to make a restraint"""
+        for atoms in [self.restrained_receptor_atoms, self.restrained_ligand_atoms]:
+            # Atoms should be a list or None at this point due to the _RestrainedAtomsProperty class
+            if atoms is None or not (isinstance(atoms, list) and len(atoms) > 0):
```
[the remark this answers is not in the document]
andrrizzi: If I understand correctly, `atoms` is ensured to be a `list` by `_RestrainedAtomsProperty`. The standard way of using `assert` vs `raise` is to use the first for developers (or in other words for debug, for example to check that the internal state of the object is the one we expect) and the second one for the user. In this case, if `isinstance(atoms, list)` fails, it means there's a bug somewhere, so I'd use an assert for this particular condition. But this is just a formality anyway.

Yank/yank.py
andrrizzi: I see. I think we should favor readability over lower cyclomatic complexity though. Again, this works perfectly so feel free to merge it, but I'd change the whole block with a more compact
```python
if isinstance(selection, int):
    registered_selectors = SelectInt,
elif isinstance(selection, str):
    registered_selectors = SelectRegion, SelectDsl, SelectSmarts
elif not all([isinstance(i, np.integer) or isinstance(i, int) for i in selection_input]):
    raise ValueError('...')
else:
    registered_selectors = SelectIterable,
```
I think it's just easier to understand what's going on, and we can sacrifice the extensibility since this is all hidden inside the function anyway.

PR #797, opened 2017-10-18
New automatic parameterization scheme for FlatBottom and various small fixes

andrrizzi:
- Save NaN logs into experiment folder. This avoids multiple experiments running in parallel to overwrite files generated for previous failures.
- Setup all systems even if experiments are not specified. This way it's easier to use the pipeline just to setup up systems.
- Fix experiment log file name when running a list of experiments in the YAML script.
- Raise an error if restraint shell volume is bigger than system volume when computing the standard state correction.


andrrizzi:
One of the test cases using the `FlatBottom` restraint is now failing because of the new volume check. I'll have to implement in this PR the new automatic parameterization scheme we discussed before merging.

andrrizzi:
I had to push the fix for #803 here because I needed all in one place to continue my simulations. I'll implement today the new `FlatBottom` automatic parameterization so that I can merge this.

I've moved the option `integrator_splitting` from being an `AlchemicalPhaseFactory` option to be an `ExperimentBuilder` option since it's used in `ExperimentBuilder._build_experiment`. `AlchemicalPhaseFactory.initialize_alchemical_phase` now uses the first integrator move in the sampler for equilibration (with appropriate collision rate and number of steps).

andrrizzi:
Ok! This should be ready now.

Lnaden:
Should be fine! go for it

andrrizzi:
Thanks!

PR #799, opened 2017-10-18
Add subset selection to Topography

Lnaden:
Adds the ability to filter atoms in the `select` function on a first pass, making the `selection` a relative choice. This is a required step for selecting SMARTS strings so we are only selecting parts of the topography first.

This relies on recursion through the `select` function itself. Open to suggestions for other test cases to include.


andrrizzi:
The code looks fine! But I think I'm missing the big picture here. Isn't this something we can already get through the `and` operator?

Lnaden:
For `Region`s and only regions, you are absolutely right. if we want to do a SMARTS string, then no. 

The main use case for this is given a SMARTS string on a small ligand, how do you then apply that SMARTS selection to only the ligand inside the Topography, which is just the whole molecule? Since we don't allow compound searches through different selection types (e.g. mixing region and MDTraj strings), then I had to introduce a method to sub-search first. See the changes in the `experiment.py` file in this PR to see that in use.

I did just have a thought that I will need to update the YAML docs because region selection is now relative to the molecule you define the region in.

andrrizzi:
Ah, I see the problem now, thanks!

> we don't allow compound searches through different selection types (e.g. mixing region and MDTraj strings)

I wonder if (in the future) we should add a syntax to specify that.

Lnaden:
I think it would be a great thing to have in the future, but not in this PR.

Thanks for the review. I'll merge when tests pass

PR #809, opened 2017-11-02
MPI exception handling

### LINKED ISSUE #798: NaN handling fails with MPI
When running a calculation on multiple GPUs, if one process finds a NaN, the exception is not raised anymore but collected by `ExperimentBuilder`. As a consequence, the other processes don't know that something is wrong, and they get stuck at the `MPI_gather` after propagation.

andrrizzi:
Fix #798 : NaN handling fails with MPI.

This adds exception handling support in the `mpi.py` module. Briefly, if a node's task raises an exception, it gets propagated the other nodes according to the `propagate_exceptions_to` argument that can assume the values `'all'`, `'group'` or `None`.

I've also added the `_MpiProcessingUnit` context manager. This will probably be the first step toward a more object-oriented API that separates the computational resources from the task execution.

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -721,7 +721,7 @@ class ExperimentBuilder(object):
         if len(self._experiments) == 0:
             raise YamlParseError('No experiments specified!')
 
-        # Setup and run all experiments with paths relative to the script directory
+        # Setup and run all experiments with paths relative to the script directory.
         with moltools.utils.temporary_cd(self._script_dir):
             self._check_resume()
             self._setup_experiments()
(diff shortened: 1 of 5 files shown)


andrrizzi:
Thanks for the review! I made `_run_experiment()` return `True` in that case because it looked to me that the exception was used only as a bool at this point of the code
```python
for exp_index in range(len(all_experiments)-1, -1, -1):
    if completed[exp_index]:
        all_experiments.pop(exp_index)
```
so I thought about simplifying it. Is there any other use of the returned exception I missed?

Lnaden:
Hmm.. No, I think you're right. I think that is from when I was originally working ont hat feature and I was going to actually do something with the exception at that point (like send it to the logger there), but at this point its fine. Do we want to trap other errors as well, because any other error will halt all of the sims I think.

andrrizzi:
> Do we want to trap other errors as well

Good point. I'd be very conservative on the types of errors that don't stop the execution though. What if we implement them as needed in the future only when we notice it's a problem (like we did with the nans)?

Lnaden:
Sounds good to me. We really should not get other errors at that point anyways

andrrizzi:
Tests pass!

Issue #465, opened 2016-08-28
Yank allows purely numeric mol2 filenames as input, but tleap throws syntax error


jlerche:
When a mol2 file, for example '5343.mol2', is used as an input file name, yank proceeds just fine until the tleap step which then throws a syntax error. [Pull request 464](https://github.com/choderalab/yank/pull/464) offers a prototype check (most likely not in the right place though) to catch these kind of filenames and provides an error message.

found in yank 0.9.0

jchodera:
Thanks!

As noted in #464: I think the better way to handle this is to ensure that we write an intermediate filename that tleap can read, rather than not be able to deal with numeric names.

jlerche:
Agreed, that is probably a better way of handling it.

jlerche:
It also seems that tleap does not allow variable names where the first and last character are digits

534E56 throws error but 534E5T is fine.

andrrizzi:
Hmm, this is weird. The [wrapper of `TLeap`](https://github.com/choderalab/yank/blob/master/Yank/utils.py#L1247) I wrote actually works only on hardcoded filenames that are copied to a temporary folder exactly to avoid problems with file names/paths, so this should not come up unless you try to run the `tleap` script manually.

I'll try to dig into why this happens. And thanks for the PR! I'll review it after I understand better what's happening.

jlerche:
The pull request is outdated but it's a matter of checking fname[0].isdigit() and fname[-1].isdigit(), maybe. I didn't exhaustively test the possibilities.

re tleap: If the original mol2 filename is kept as the intermediate one (which I believe it does, at least judging by the tleap scripts that were created) then the issue will still persist. A possibility might be to append an alphabetical character or string to the filename depending on how many mol2s there are.

andrrizzi:
> at least judging by the tleap scripts that were created

The tleap scripts that are saved for logging contains the original file names, but the tleap script that is actually _executed_ by Yank substitute those names with `molecule1.mol2`, `molecule2.mol2` names so it should not have that problem.

Lnaden:
@andrrizzi do you know if this is still an issue and/or if it got fixed from other updates to YAML since then?

1.0 touch

andrrizzi:
Ah! Thanks for reminding me about this. It should be still there. I'll check it today.

andrrizzi:
I was able to reproduce this. The problem is not with file names which are hardcoded in the executed script, but with the name of the group which is set to the molecule id. For example, having
[...]

PR #819, opened 2017-11-13
Bugfix and add replica state index statistical inefficiency to analysis

### LINKED ISSUE #815: Find better way to generate residue names for ligands generated from SMILES
Currently, we just use the first three letters of the molecule's name in the YAML script, but this can generate collisions with common residue names used for amino acids and solvent. The solutions are see are two:
1. Call everything with a standard name (e.g. `MOL`). This has the advantage of being very predictable from the user perspective.
2. Alternatively, we keep this scheme but we need to check that the generated name is not in the set of MDTraj common resnames that we use for selection in Topography. We still need to fall back to a standard resname if this is the case.

cc: choderalab/yank-examples#76

### LINKED ISSUE #816: Setup logs are saved in the wrong folder
The logs for the setup phase are stored in the incorrect folder, which results in 2 setup folders created. This doesn't have any consequence for the simulation, but it can be confusing. I'll push a fix later today.

### LINKED ISSUE #817: Compute statistical inefficiency of replica state index in analysis
To add this capability to the analysis, I'm thinking about doing the following:
* Modify `ReplicaExchangeAnalyzer.generate_mixing_statistics()` to also return the result of `pymbar.timeseries.statisticalInefficiencyMultiple()` on the data (after discarding equilibration iterations).
* Modify `ReplicaExchangeAnalyzer.show_mixing_statistics()` and `notebook.HealthReportData.generate_mixing_plot()` to print this statistic right below the Perron eigenvalue and the state equilibration timescale.

Does this sound reasonable?

### LINKED ISSUE #818: Crash when setting FlatBottom/Harmonic restrained atom to a single int
When setting `restrained_ligand_atoms` or `restrained_receptor_atoms` to a single integer, I get the error
`TypeError: 'int' object is not iterable`.

This is caused by a small typo in this dispatch call in `RadiallySymmetricRestraint._determine_restrained_atoms`.
```python
@compute_atom_set.register(int)
def compute_atom_int(input_atom, topography_key, _):
    """Helper for int parsing, ensures the atom you restrain is actually part of the molecule you chose"""
    assert set(input_atom) & set(getattr(topography, topography_key))
    return [input_atom]
```
It should be `set([input_atom])`. I'll push a fix soon and add a test case to `test_restraints`.

andrrizzi:
In this PR:
* Fix #816: Setup logs are saved in the wrong folder.
* Fix #818: crash with single FlatBottom/Harmonic restrained atom.
  - I've slightly refactored the descriptor `_RestrainedAtomsProperty` to remove shared code between Boresch-style and radially symmetric restraints and to always casts integers to a list so that we don't need to worry about this extra case in the internal logic.
* Fix #815: Better way to generate residue name for SMILES and sdf molecules.
  - We now generate a residue name in the form `YXX` where `Y` is an uppercase letter and `XX` are two digits. This gives us a total of 2600 unique residue names. The algorithm fails to maintain unique residue names if the list of molecules is updated in the middle of the setup, but I couldn't find a better solution to keep supporting parallel setup. Suggestions are welcomed. We never use the `SetupPipeline` class this way anyway, so this should not be a problem.
* Fix #817: Compute statistical inefficiency of replica state index in analysis.
  - I've added this to both the text analysis and the notebook.
* I've also fixed the alignment of the final printed free energy. I broke it several months ago and it was driving me crazy.

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -785,8 +785,8 @@ class ReplicaExchangeAnalyzer(YankPhaseAnalyzer):
             )
 
         # Print information about replica state index statistical efficiency.
-        logger.info('Replica state index statistical efficiency is '
-                    '{:.3f}'.format(mixing_statistics.statistical_efficiency))
+        logger.info('Replica state index statistical inefficiency is '
+                    '{:.3f}'.format(mixing_statistics.statistical_inefficiency))
 
     def get_states_energies(self):
         """
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -1614,10 +1614,11 @@ def extract_trajectory(nc_path, nc_checkpoint_file=None, state_index=None, repli
                 n_equil_frames = n_equil_iterations
             frame_indices = frame_indices[n_equil_frames:-1]
 
-        # Extract state positions and box vectors
-        positions = np.zeros((len(frame_indices), n_atoms, 3))
+        # Extract state positions and box vectors.
+        # MDTraj Cython code expects float32 positions.
+        positions = np.zeros((len(frame_indices), n_atoms, 3), dtype=np.float32)
         if is_periodic:
-            box_vectors = np.zeros((len(frame_indices), 3, 3))
+            box_vectors = np.zeros((len(frame_indices), 3, 3), dtype=np.float32)
         if state_index is not None:
             logger.info('Extracting positions of state {}...'.format(state_index))
 
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -1408,20 +1408,20 @@ def analyze_directory(source_directory):
             calculation_type = ' of solvation'
 
     # Print energies
-    logger.info("")
-    logger.info("Free energy{}: {:16.3f} +- {:.3f} kT ({:16.3f} +- {:.3f} kcal/mol)".format(
+    logger.info('')
+    logger.info('Free energy{:<13}: {:9.3f} +- {:.3f} kT ({:.3f} +- {:.3f} kcal/mol)'.format(
         calculation_type, DeltaF, dDeltaF, DeltaF * kT / units.kilocalories_per_mole,
         dDeltaF * kT / units.kilocalories_per_mole))
-    logger.info("")
+    logger.info('')
 
     for phase in phase_names:
-        logger.info("DeltaG {:<25} : {:16.3f} +- {:.3f} kT".format(phase, data[phase]['DeltaF'],
-                                                                   data[phase]['dDeltaF']))
+        logger.info('DeltaG {:<17}: {:9.3f} +- {:.3f} kT'.format(phase, data[phase]['DeltaF'],
+                                                                 data[phase]['dDeltaF']))
         if data[phase]['DeltaF_standard_state_correction'] != 0.0:
-            logger.info("DeltaG {:<25} : {:25.3f} kT".format('restraint',
-                                                             data[phase]['DeltaF_standard_state_correction']))
-    logger.info("")
-    logger.info("Enthalpy{}: {:16.3f} +- {:.3f} kT ({:16.3f} +- {:.3f} kcal/mol)".format(
+            logger.info('DeltaG {:<17}: {:18.3f} kT'.format('restraint',
+                                                            data[phase]['DeltaF_standard_state_correction']))
+    logger.info('')
+    logger.info('Enthalpy{:<16}: {:9.3f} +- {:.3f} kT ({:.3f} +- {:.3f} kcal/mol)'.format(
         calculation_type, DeltaH, dDeltaH, DeltaH * kT / units.kilocalories_per_mole,
         dDeltaH * kT / units.kilocalories_per_mole))
 
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -21,10 +21,11 @@ Fully extensible to support new samplers and observables.
 # =============================================================================================
 
 import os
-import os.path
-
 import abc
 import copy
+import typing
+from typing import Union
+
 import yaml
 import numpy as np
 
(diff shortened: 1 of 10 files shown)


andrrizzi:
Also fixes #821: Extract trajectory fails to image molecules.

Yank/pipeline.py
```diff
@@ -1112,6 +1111,43 @@ def _generate_molecule(self, molecule_id):
...
+        -------
+        residue_name : str
+            A three-character residue name.
+        """
+        # We need to associate a unique number to this molecule, and do
+        # so in such a way that distributing molecule setups over multiple
+        # MPI process still ends up in unique residue names for each molecule.
+        molecule_ids = sorted(self.molecules.keys())
+        n_molecule = molecule_ids.index(molecule_id)
+        assert n_molecule < 2600
```
Lnaden: I don't think this will be a problem any time soon, but what happens in the combinatorial space, if someone, for instance, set up a combinatorial molecule which got to `>2600` molecules, would this fail?

Not something we need to fix now by any means, just thinking about the conditions we might see it fail for the first time that a user my not be expecting.

Yank/pipeline.py
```diff
@@ -1112,6 +1111,43 @@ def _generate_molecule(self, molecule_id):
 
         return molecule
 
+    def _generate_residue_name(self, molecule_id):
```
Lnaden: Can we leverage this to fix #465 while we are at it?

Yank/reports/notebook.py
```diff
@@ -345,43 +346,57 @@ def generate_mixing_plot(self, mixing_cutoff=0.05, mixing_warning_threshold=0.90
...
-            output_image = subplot.imshow(mixing_data, aspect='equal', cmap=cmap, vmin=0, vmax=1)
-            # Add colorbr
+            mixing_statistics = analyzer.generate_mixing_statistics(
+                number_equilibrated=self.nequils[phase_name])
+            transition_matrix, eigenvalues, statistical_inefficiency = mixing_statistics
+
+            # Without vmin/vmax, the image normalizes the values to mixing_data.max
+            # which screws up the warning colormap.
+            # Can also use norm=NoNorm(), but that makes the colorbar manipulation fail.
+            output_image = subplot.imshow(mixing_statistics.transition_matrix, aspect='equal',
```
Lnaden: Why not just use `transition_matrix` here since you split it off in the line above?

Yank/tests/test_analyze.py
```diff
@@ -182,7 +182,7 @@ def test_repex_phase_initilize(self):
     def test_repex_mixing_stats(self):
         """Test that the Repex Phase yields mixing stats that make sense"""
         phase = analyze.ReplicaExchangeAnalyzer(self.reporter, name=self.repex_name)
-        t, mu = phase.generate_mixing_statistics()
+        t, mu, g = phase.generate_mixing_statistics()
```
Lnaden: Not something you have to but remember if you don't use a variable you expand from, you can always cast it to `_` to get the IDE to stop complaining about unused variable

Yank/tests/test_experiment.py
```diff
@@ -772,8 +772,9 @@ def test_validation_correct_experiments():
         {'system': 'sys', 'protocol': 'absolute-binding', 'restraint': {
             'type': 'FlatBottom', 'well_radius': '5.2*nanometers', 'restrained_receptor_atom': 1644}},
         {'system': 'sys', 'protocol': 'absolute-binding', 'restraint': {
-            'type': 'Boresch', 'restrained_atoms': [1335, 1339, 1397, 2609, 2607, 2606],
```
Lnaden: How did this test pass before?

Yank/tests/test_restraints.py
```diff
@@ -183,7 +183,7 @@ def test_harmonic_standard_state():
     LJ_fluid = testsystems.LennardJonesFluid()
 
     # Create Harmonic restraint.
-    restraint = yank.restraints.create_restraint('Harmonic', restrained_receptor_atoms=[1])
+    restraint = yank.restraints.create_restraint('Harmonic', restrained_receptor_atoms=1)
```
Lnaden: Is there a test somewhere where a list of int (even if its len == 1) is checked? If not, it would be good to make one to ensure functionality.

Yank/analyze.py
```diff
@@ -1408,20 +1433,20 @@ def analyze_directory(source_directory):
             calculation_type = ' of solvation'
 
     # Print energies
-    logger.info("")
-    logger.info("Free energy{}: {:16.3f} +- {:.3f} kT ({:16.3f} +- {:.3f} kcal/mol)".format(
+    logger.info('')
```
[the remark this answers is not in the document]
andrrizzi: Nothing major, I try to keep all string consistent because it makes it easier to search particular strings (otherwise I have to search both with `'` and `"` characters), and we usually want to have `"` inside a string so I picked the single quote.

Yank/analyze.py
```diff
@@ -1589,10 +1614,11 @@ def extract_trajectory(nc_path, nc_checkpoint_file=None, state_index=None, repli
                 n_equil_frames = n_equil_iterations
             frame_indices = frame_indices[n_equil_frames:-1]
 
-        # Extract state positions and box vectors
-        positions = np.zeros((len(frame_indices), n_atoms, 3))
+        # Extract state positions and box vectors.
+        # MDTraj Cython code expects float32 positions.
+        positions = np.zeros((len(frame_indices), n_atoms, 3), dtype=np.float32)
```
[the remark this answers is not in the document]
andrrizzi: > Does it expect python float or will it take a numpynp.float32?

It expects numpy float32 (directly from the netcdf file).
> and what is the default here normally? `np.float64`?

I'm not sure actually, but based on the error that mdtraj was giving us, I think that's the case.

Yank/pipeline.py
andrrizzi: That's actually a different problem that should be handled by `utils.TLeap`, but I can definitely make the change and add it to this PR!

Yank/pipeline.py
andrrizzi: It'll keep going through the ascii characters. There are a handful of special symbols that worry me in between uppercase and lowercase letters though, that's why I added the `assert` statement for now. In the future we can skip the next characters and go to lowercase letters to get other 2600 possible resnames.

Yank/reports/notebook.py
andrrizzi: Thanks! I didn't notice it. Will do.

Yank/tests/test_experiment.py
andrrizzi: The schema did not check for restraints arguments at parsing time. The error, in this case, would have been thrown only at runtime. Might be something that we need to add.

Yank/tests/test_restraints.py
andrrizzi: Yep! There's one in `test_restraints.test_partial_parametrization`.

Yank/pipeline.py
Lnaden: I actually think the assert is a good thing since the current algorithm will generate unique `[A-Z]\d\d` patterns for 2600 molecules, knowing the limit so when/if it is ever hit (probably not any time soon) we know its a problem instead of errors which are not failures (which is way worse)

andrrizzi:
Added the fix for #465 . @Lnaden could you take a look at the last commit before I merge?

Lnaden:
That looks good to me! Did you test it?

andrrizzi:
Yup! I adapted the two `TLeap` tests to use numeric unit names (both failed before the fix).

Thanks for the review!

Lnaden:
Awesome. I think we're making some good changes with these recent PRs from a end-user standpoint. Thanks!

PR #825, opened 2017-11-16
Fix a couple site document issues

Lnaden:


andrrizzi:
Looks good, but I'm not entirely sure we want to set `ph_tolerance: 7.0` as a standard example. If I remember correctly, the few times I cranked up the tolerance, Epik took forever to finish. Are protonation states with probability 10^-7 important to include in a YANK calculations? Otherwise, we could set it to some value in between.

Lnaden:
I agree, I think it might be a bit too large. @jchodera, @steven-albanese, any feedback as to a reasonable ph_tolerance, or reason why you think +-7 ph is a good tolerance for ph? Or am I mis-interpreting the function of that parameter?

Lnaden:
I added a warning about Epik calculation times in the `ph_tolerance` option of `epik`. Thanks to @steven-albanese for some input on this!

jchodera:
10^7 is a penalty of 16 kT, which is quite large. If we want to permit a more modest (but realistic) 6 kT penalty, that would be ~3 pK units.

PR #826, opened 2017-11-16
Add check for solvent_model vs parameter mismatch

### LINKED ISSUE #776: YANK tries to solvate a system in LEaP with TIP4P-Ew even if TIP3P is used
cc: https://github.com/choderalab/yank-examples/issues/61

We currently use `solvent_model: tip4pew` by default, but don't bother to check if the appropriate leaprc file is used or vice-versa. This leads to quite non-intuitive error messages when extra sites are added or missing when they are not expected.

Lnaden:
This adds a check for the solvent_model parameter vs what files were loaded to TLEaP. Its a bit crude, but provides better handling of the error and tries to inspect the LEaP log for the error it knows.

Its a 2-fold trap:
First it warns the user if they have not loaded the expected leaprc file, warning looks like:
```
WARNING - yank.pipeline - WARNING: The solvent_model tip4pew may not work for loaded leaprc.water.X files.
 We expected leaprc.water.tip4pew to make your solvent model work, but did not find it.
 This is okay for tip4pew leaprc file with tip3p solvent_model, but not the other way around.
 ```

 Second, if the system fails to build, the logfile is searched for a common match error an throws an error with the following message:
 ```
 RuntimeError: Solvent {SOLVENT_ID}: It looks like the water used has virtual sites, but missing parameters.
Make sure your leap parameters use the correct water model as specified by solvent_model.
Please see the log file for LEaP for more info as there may be more wrong:
{ORIGINAL_MESSAGE}
 ```
 Where ORIGINAL_MESSAGE is filled in with the actual error thrown by the `telap.run()` function in `utils`, and SOLVENT_ID is the name of the solvent as it appears in YAML

 I have also updated the FAQ and the example in the yaml pages to use the default. cc @jchodera since this has come up a few times

 Thoughts?

 Fixes #776
 cc #759 as its a related bug, since this sets up the framework for catching other errors like the one there.

### DIFF
--- a/Yank/pipeline.py
+++ b/Yank/pipeline.py
@@ -18,6 +18,7 @@ Utility functions to help setting up Yank configurations.
 
 import os
 import re
+import sys
 import copy
 import inspect
 import logging
@@ -529,11 +530,6 @@ _OPENMM_LEAP_SOLVENT_FILES_MAP = {
     'tip4pew': 'leaprc.water.tip4pew',
     'tip5p': 'leaprc.water.tip4pew',  # Enables the EP atom type
     'spce': 'leaprc.water.spce',
-}
-
-# Reverse map which helps in some cases
-_OPENMM_LEAP_SOLVENT_FILES_MAP_R = {
-    v: k for k, v in _OPENMM_LEAP_SOLVENT_FILES_MAP.items()
 }
 
 
@@ -1402,7 +1398,7 @@ class SetupDatabase:
 
         # Start error tracking variables
         # Water
-        known_solvent_files = [file for file in _OPENMM_LEAP_SOLVENT_FILES_MAP_R.keys()]
+        known_solvent_files = [file for file in _OPENMM_LEAP_SOLVENT_FILES_MAP.values()]
         loaded_water_files = []  # Detected loaded water files
 
         def extend_list_of_waters(leap_parameters):
@@ -1488,7 +1484,6 @@ class SetupDatabase:
             solvent_model = solvent['solvent_model']
             # Check that solvent model has loaded the appropriate leap parameters.
             # This does guarantee a failure, but is a good sign of it.
-            solvent_warning = None
             if _OPENMM_LEAP_SOLVENT_FILES_MAP[solvent_model] not in loaded_water_files:
                 solvent_warning = ("WARNING: The solvent_model {} may not work for loaded "
                                    "leaprc.water.X files.\n We expected {} to make your "
@@ -1583,27 +1578,9 @@ class SetupDatabase:
         try:
             warnings = tleap.run()
         except RuntimeError as e:
-            # Try to find the log file
-            m = re.search('Check log file (.*)', str(e))
-            if m is not None:
-                log_file_string = m.group(1)
-                try:
-                    with open(log_file_string, 'r') as log_file:
-                        log_file_data = log_file.read()
-                    # Analyze log file for water mismatch
-                    if "EP - OW" in log_file_data:
-                        # Found mismatch water and missing parameters
-                        error = ("Solvent {}: "
-                                 "It looks like the water used has virtual sites, but "
-                                 "missing parameters.\nMake sure your leap parameters "
-                                 "use the correct water model as specified by "
-                                 "solvent_model.\nPlease see the log file for LEaP "
-                                 "for more info as there may be more wrong:\n{}")
-                        raise RuntimeError(error.format(solvent_id, e))
-                except IOError:
-                    pass
-            # If no other condition was found, raise the original error
-            raise e
+            error = RuntimeError('Solvent {}: {}'.format(solvent_id, str(e)))
+            error.with_traceback(sys.exc_info()[2])
+            raise error
 
         for warning in warnings:
             logger.warning('TLeap: ' + warning)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: Yank/pipeline.py [hunk 1]
```diff
@@ -1381,6 +1400,17 @@ def _setup_system(self, system_file_path, pack, alchemical_charge,
...
+        # Start error tracking variables
+        # Water
+        known_solvent_files = [file for file in _OPENMM_LEAP_SOLVENT_FILES_MAP_R.keys()]
+        loaded_water_files = []  # Detected loaded water files
+
+        def extend_list_of_waters(leap_parameters):
+            """Extend the loaded_water_files list given the leap_parameters"""
+            loaded_water_files.extend([water for water
+                                       in leap_parameters
+                                       if water in known_solvent_files])
```
andrrizzi: If you decide to keep `_OPENMM_LEAP_SOLVENT_FILES_MAP_R ` above, you could remove `known_solvent_files` and use directly `_OPENMM_LEAP_SOLVENT_FILES_MAP_R` to search by hash instead of going through the list.

Yank/pipeline.py
```diff
@@ -1451,6 +1486,19 @@ def _setup_system(self, system_file_path, pack, alchemical_charge,
 
             # Solvate unit. Solvent models different than tip3p need parameter modifications.
             solvent_model = solvent['solvent_model']
+            # Check that solvent model has loaded the appropriate leap parameters.
+            # This does guarantee a failure, but is a good sign of it.
+            solvent_warning = None
```
andrrizzi: This seems to be unused outside the `if` so you should be able to remove this line.

Yank/pipeline.py
```diff
@@ -518,6 +518,24 @@ def read_system_files(positions_file_path, parameters_file_path, system_options,
...
+_OPENMM_LEAP_SOLVENT_FILES_MAP = {
+    'tip3p': 'leaprc.water.tip3p',
+    'tip3pfb': 'leaprc.water.tip3p',
+    'tip4pew': 'leaprc.water.tip4pew',
+    'tip5p': 'leaprc.water.tip4pew',  # Enables the EP atom type
+    'spce': 'leaprc.water.spce',
+}
+
+# Reverse map which helps in some cases
+_OPENMM_LEAP_SOLVENT_FILES_MAP_R = {
```
andrrizzi: Only the keys of this dictionary are used in the code. What about generating this at runtime in `_setup_system()` as `known_solvent_files  = set(_OPENMM_LEAP_SOLVENT_FILES_MAP.values())`?

Yank/pipeline.py
```diff
@@ -1451,6 +1486,19 @@ def _setup_system(self, system_file_path, pack, alchemical_charge,
...
+            # Check that solvent model has loaded the appropriate leap parameters.
+            # This does guarantee a failure, but is a good sign of it.
+            solvent_warning = None
+            if _OPENMM_LEAP_SOLVENT_FILES_MAP[solvent_model] not in loaded_water_files:
+                solvent_warning = ("WARNING: The solvent_model {} may not work for loaded "
+                                   "leaprc.water.X files.\n We expected {} to make your "
+                                   "solvent model work, but did not find it.\n "
+                                   "This is okay for tip4pew leaprc file with tip3p solvent_model, "
+                                   "but not the other way around.\nThis does "
+                                   "not mean the system will not build, but it may throw "
```
andrrizzi: Hmm. I'm worried that we risk clogging the output with warnings and errors so detailed since they get printed even if `verbose` is `False`.

This would totally be a separate issue/PR, but to keep our messages both compact and exhaustive, what do you think about having short error/warning messages that convey the point associated to numerical codes and have a page on the docs with more details on the errors in case the user is not already familiar with them?

Yank/pipeline.py
```diff
@@ -1531,7 +1579,32 @@ def _setup_system(self, system_file_path, pack, alchemical_charge,
...
+                    # Analyze log file for water mismatch
+                    if "EP - OW" in log_file_data:
+                        # Found mismatch water and missing parameters
+                        error = ("Solvent {}: "
+                                 "It looks like the water used has virtual sites, but "
+                                 "missing parameters.\nMake sure your leap parameters "
+                                 "use the correct water model as specified by "
+                                 "solvent_model.\nPlease see the log file for LEaP "
+                                 "for more info as there may be more wrong:\n{}")
+                        raise RuntimeError(error.format(solvent_id, e))
```
andrrizzi: Should we put this code in `tleap.run()` rather than here? The problem is fairly general. To keep the `solvent_id` specification in the error message and preserve the traceback we could do something like
```python
try:
    warnings = tleap.run()
except RuntimeError as e:
    error = RuntimeError('{}: {}'.format(solvent_id, str(e)))
    error.with_traceback(sys.exc_info()[2])
    raise error
```

Lnaden: Will that work as-is?

Yank/pipeline.py
Lnaden: I like this idea, but I think this should be handled in a future PR, and when we go through and make better error handling more robust. Like I said, this is a bit of a crude fix, but it sets up the framework. I'll make an enhancement issue or this.

Yank/pipeline.py
andrrizzi: Yeah, I think so!

PR #827, opened 2017-11-20
Start end and skip are ignored in extract trajectory

### LINKED ISSUE #558: --end argument to 'yank analyze extract-trajectory' doesn't appear to work
I just noticed that the `--end=framenum` argument to `yank analyze extract-trajectory` doesn't appear to work.

andrrizzi:
Fix #558.


andrrizzi:
Thanks!

PR #830, opened 2017-11-21
Cut the 0.19.0 release

Lnaden:
* Requires minimum version of OpenMMTools 0.13.3 for critical fix in SamplerState
* Changed from Schema to Cerberus for YAML data validation
* Adds more general region selection schema for Topography and atom selection
    * Establishes the future framework for SMARTS selection

@andrrizzi could you double check this has all the features we want?

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -20,7 +20,7 @@ requirements:
     - netcdf4
     - openmm >=7.1
     - mdtraj >=1.7.2
-    - openmmtools >=0.13.2
+    - openmmtools >=0.13.3
     - pymbar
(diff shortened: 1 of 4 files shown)


Lnaden:
So this release will not have @andrrizzi's local context re-initilization scheme in, the current fix is to use OpenCL, and it also only came up in rare conditions. I will set up Merck sims with this release and manually insert the hacky, slow work-around while we figure out the actual problems.

andrrizzi:
Looks good to me! The fix will have to be implemented in `openmmtools`, not in YANK. I haven't had time to re-run the bromodomain simulations with the `SamplerState` fix yet, so I'm not sure that's still a problem there.

PR #835, opened 2017-11-30
Make Analysis only particles include ions

### LINKED ISSUE #559: --nosolvent argument to 'yank analyze extract-trajectory` should include ions
We should probably retain the counterions when `--nosolvent` is used.
We can provide a separate `--noions` argument to drop the counterions.

Lnaden:
Fixes #559

cc @andrrizzi

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -280,7 +280,9 @@ class AlchemicalPhaseFactory(object):
             checkpoint_interval = self.options['checkpoint_interval']
             # Get the solute atoms
             if self.options['store_solute_trajectory']:
-                solute_atoms = self.topography.solute_atoms
+                # "Solute" is basically just not water. Includes all non-water atoms and ions
+                # Use set union (|) cast back to list to get output
+                solute_atoms = list(frozenset(self.topography.solute_atoms) | frozenset(self.topography.ions_atoms))
                 if checkpoint_interval == 1:
                     logger.warning("WARNING! You have specified both a solute-only trajectory AND a checkpoint "
                                    "interval of 1! You are about write the trajectory of the solute twice!\n"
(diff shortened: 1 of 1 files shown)


Yank/experiment.py
```diff
@@ -280,7 +280,9 @@ def create_alchemical_phase(self):
             checkpoint_interval = self.options['checkpoint_interval']
             # Get the solute atoms
             if self.options['store_solute_trajectory']:
-                solute_atoms = self.topography.solute_atoms
+                # "Solute" is basically just not water. Includes all non-water atoms and ions
+                # Use set union (|) cast back to list to get output
+                solute_atoms = list(frozenset(self.topography.solute_atoms) | frozenset(self.topography.ions_atoms))
```
andrrizzi: `Topography` already guarantees that the intersection of `solute_atoms` and `ions_atoms` is null. I think we should do this with something like
```python
solute_atoms  = self.topography.solute_atoms + self.topography.ions_atoms
```
because going through a `set` can scramble the order of the atom indices, which feels more error-prone. The order should already be correct as ions atoms are always added after the solute.

Lnaden: > intersection of solute_atoms and ions_atoms is null.

Ah, I missed that, 

>  going through a set can scramble the order

Which is why I used frozen set. But since they already don't intersect, I can fix that, but this seemed more future proof to me.

Lnaden:
Got it anyways, I was making some changes to other PRs and branches I am working on, easy fix

Yank/experiment.py
andrrizzi: I'm not sure that `frozenset` differs from the normal `set` in this. I believe there's no guarantee about the order of the elements in both cases, especially after the intersection operation.

Lnaden:
Great, If the tests pass, I'll merge

PR #840, opened 2017-12-01
Add support for amber rst7 files and make jobid 1-based

### LINKED ISSUE #838: Support for amber rst7 system files
I've realized that to run the SAMPLing calculations I'll have to add support for `rst7` files as an alternative to `inpcrd`. This is already supported, but it needs to be allowed in the YAML syntax. I have a local fix ready. I'll push it tomorrow.

Question: do we need to enforce `phase1_path` and `phase2_path` to have the same type of input files (e.g., both phases have to use amber files)? The two phases are completely independent calculations so I'm tempted to remove this restriction. Any objection?

### LINKED ISSUE #839: Make --jobid a 1-based id
To run YANK with a job array, we have implemented the command line options `--jobid` and `--njobs`. Only schedulers usually use 1-based ids while YANK uses 0-based. We should probably use 1-based job ids for YANK too to avoid doing arithmetic in bash in the LSF script.

andrrizzi:
- Fix #838: Support for amber rst7 system files.
- Fix #839: Make jobid 1-based.

After this is merged, I'd make a 0.19.3 release that we can refer to in the SAMPL paper, if there are no objections.

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -1657,16 +1657,23 @@ class ExperimentBuilder(object):
             schema:
                 type: string
                 validator: file_exists
-            dependencies: phase2_path
-            coerce: sort_alphabetically_by_extension
+            validator: supported_system_files
+            coerce: sort_system_files_by_type
         phase2_path:
             required: no
             type: list
             schema:
                 type: string
                 validator: file_exists
-            validator: is_system_files_matching_phase1
-            coerce: sort_alphabetically_by_extension
+            validator: supported_system_files
+            coerce: sort_system_files_by_type
+        gromacs_include_dir:
+            required: no
+            type: string
+            dependencies: [phase1_path, phase2_path]
+            validator: directory_exists
+
+        # Solvents
         solvent:
             required: no
             type: string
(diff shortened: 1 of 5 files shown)


Yank/experiment.py
```diff
@@ -1657,16 +1659,21 @@ def _is_pipeline_solvent_with_receptor(field, solvent_id, error):
             schema:
                 type: string
                 validator: file_exists
-            dependencies: phase2_path
```
Lnaden: Why did you remove the dependency on `phase2_path` here?

Yank/experiment.py
```diff
@@ -502,7 +502,7 @@ class ExperimentBuilder(object):
         can load it later by using :func:`parse` (default is None).
     job_id : None or int
         If you want to split the experiments among different executions,
-        you can set this to an integer 0 <= job_id <= n_jobs-1, and this
+        you can set this to an integer 1 <= job_id <= n_jobs, and this
```
Lnaden: I don't see it in the yaml docs, but just to make sure: `job_id` is not a parameter which can be YAML specified, correct?

andrrizzi: Correct! It's just a CLI option to make it work with array jobs.

Yank/experiment.py
andrrizzi: For the records, we talked about this, and we decided to postpone trying to rework the schema to use `required: yes` and `excludes: [receptor, ligand, solute]` in a future PR.

I'll re-add the `dependencies` solution here before merging!

Lnaden:
I can't approve the codeclimate issue, which seems odd. Can you?

andrrizzi:
Odd. I can't find anything to click to approve either.

PR #845, opened 2017-12-05
[WIP] Add option to rigidify ligand to Boresch restraints

jchodera:
In inspecting @Lnaden's Abl:imatinib calculations, it looks like the ligand flexibility in the noninteracting state may be a major cause of replica poor mixing. This experimental PR adds an option to Boresch restraints that also adds torsional restraints to rotatable bonds to rigidify the ligand when the restraints are imposed.

If this works well in the `complex` phase, we would have to figure out how to also impose the torsion restraints in the `solvent` phase. But for now, I'm just testing this branch on the complex phase.

### DIFF
--- a/Yank/restraints.py
+++ b/Yank/restraints.py
@@ -1381,6 +1381,9 @@ class Boresch(ReceptorLigandRestraint):
     standard_state_correction_method : 'analytical' or 'numeric', optional
         The method to use to estimate the standard state correction (default
         is 'analytical').
+    rigidify : simtk.unit.Quantity, optional, default=None
+        If specified, will rigidify the ligand as restraints are turned on by
+        adding internal torsion barriers to lock initial configuration.
 
     Attributes
     ----------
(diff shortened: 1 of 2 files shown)


Lnaden:
What evidence is there its the non-rigidity of the ligand which causes problems? Based on the replica mixing plots, the decouple state is one of the better mixed states. If we do add this, we'll need to make sure the standard state correction is fixed as well, or include the rigidity in the solvent phase as you mentioned, otherwise the free energy difference will be quite off.

jchodera:
I examined the trajectories for Abl:imatinib replicas from your simulations, and it appeared that the entropic bottleneck of pulling the disordered ligand into the binding pose may be limiting efficiency. It would be better to offer a way to restrain the ligand RMSD, but we still don't have an RMSDForce in OpenMM (though I've wanted one for years) so adding torsion restraints was the best compromise between simplicity and minimizing the additional thermodynamic length. 

The effect will cancel if we apply the same rigidification restraints in the solvent phase. If we end up using this, we would need some refactoring to allow this.

For now, this is just an experiment. I'm running with "auto" to see if this minimizes the number of thermodynamic states needed and enhances mixing. 

You may be right that the electrostatics phase is the most problematic here. We still need to explore some simple protocol variants, but we can start those experiments today.

PR #850, opened 2017-12-08
Fix #808: Parallel MPI nodes try to setup the same molecule.

### LINKED ISSUE #808: Parallel molecule setup with Combinatorial tries to make directories over each other
Self explanatory title.

When the `!combinatorial` is set for a single molecule, the new parallel setup code causes `os.mkdir` to be called on all directories, resulting in the problem observed it #807 and choderalab/yank-examples#70

Use the `abl-imatinib` example to reproduce

andrrizzi:
The bug was caused by the non-deterministic order of `dict`s when assigning the parallel nodes to molecules/systems to set up.

### DIFF
--- a/Yank/pipeline.py
+++ b/Yank/pipeline.py
@@ -1081,9 +1081,10 @@ class SetupDatabase:
         """
         # Find all molecules that need to be set up.
         molecules_to_setup = []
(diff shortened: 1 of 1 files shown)


Lnaden:
Tested locally, confirmed works. Thank you!

PR #853, opened 2017-12-11
Fix bug experiment option and new status command

### LINKED ISSUE #852: Better handling of large combinatorial experiments
Babysitting very large combinatorial simulations can be painful right now. If only some of the calculations fail for some reason (e.g., cluster problems), we can only do one of the following:
1. Restart every combinatorial experiment. This is currently slow because we need to load the full `ReplicaExchange` object from disk to determine whether a specific phase is completed or not.
2. For job arrays, we have to manually dig in the logs to find the jobid of the failed experiments and run only those. This is obviously not ideal.

I'd like to make two changes
1. Add a class method in `ReplicaExchange` (and relative instance method in `AlchemicalPhase`) to check the status of the calculations without loading the full `ReplicaExchange` object. Something like:
```python
class ReplicaExchangeStatus(typing.NamedTuple):
    """Status of the replica-exchange calculation"""
    number_of_iterations: int
    is_completed: bool

@classmethod
def get_status(cls, storage):
    """Read information about the status of the calculation from the storage file.

    Parameters
    ----------
    storage : str or Reporter
        The path to the storage file or the reporter object.

    Returns
    -------
    status : ReplicaExchangeStatus
        The status of the calculation.
    """
```
2. Resuscitate the `yank status` CLI argument (currently broken) to optionally take a YAML script as an input instead of the directory of a single experiment, and print status information for all experiments. The output would look like
```
yank status --yaml=yank.yaml
Total number of experiments: 155 (25 completed, 75 running, 55 pending)
```
or with verbose output
```
yank status --yaml=yank.yaml -v
experiment1:
    phase1: number_of_iterations=10000, status=completed
    phase2: number_of_iterations=10000, status=completed
...
experiment55:
    phase1: number_of_iterations=10000, status=completed
    phase2: number_of_iterations=2458, status=running
...
experiment124:
    phase1: status=pending
    phase2: status=pending
...
Total number of experiments: 155 (25 completed, 75 running, 55 pending)
```
I'd also add the argument `--status` to print only info about completed/running/pending experiments, and the argument `--njobs` to associate a `jobid` to the experiments, which can be helpful for array jobs.

I think this could save us a lot of troubles in the next couple of months. Objections? Thoughts?

andrrizzi:
@Lnaden I ended up implementing the new `status` command too for the release because it would be incredibly useful to have it for the benchmark calculations.

- Fixed bug where options in experiment section were not coerced.
- It is possible to save dictionaries on the netcdf file in nested format now. This makes it possible to retrieve dictionary keywords from disk without loading the entire dictionary. We could use this in the future to retrieve the `temperature` very quickly from the thermodynamic state dictionary and speed up the analysis!
- Fix #852: faster restart and fix  `status` command to print information about all combinatorial experiments.

### DIFF
--- a/Yank/commands/status.py
+++ b/Yank/commands/status.py
@@ -36,11 +36,11 @@ Required Arguments:
 
 General Options:
   --status=STRING               Print only the jobs in a particular status. Accepted values
-                                are completed, running, and pending. This works only if
-                                --verbose is set.
+                                are "completed", "ongoing", and "pending". This works only
+                                if verbose is set.
   --njobs=INTEGER               Print the job id associated to each experiment assuming
                                 njobs to be the one specified here. This works only if
-                                --verbose is set.
+                                verbose is set.
   -v, --verbose                 Print status of each experiment individually. If this is
                                 not set, only a summary of the status of all experiments
                                 is printed.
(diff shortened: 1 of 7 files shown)


andrrizzi:
Shoot, I just realized we don't have tests activated for PRs to branches other than `master`. Feel free to review here, then I'll fix whatever there is to fix after the first merge.

Yank/commands/status.py
```diff
@@ -23,26 +27,104 @@
...
+    for k, g in itertools.groupby(enumerate(job_ids), lambda x:x[0]-x[1]):
+        group = list(map(operator.itemgetter(1), g))
+        if len(group) == 1:
+            contiguous_job_ids.append(str(group[0]))
+        else:
+            contiguous_job_ids.append('{}-{}'.format(group[0], group[-1]))
+    return ','.join(contiguous_job_ids)
+
 
 def dispatch(args):
```
Lnaden: Can this function ever return a `False` status?

Yank/repex.py
```diff
@@ -1575,6 +1659,59 @@ def from_storage(cls, storage):
...
 
+    # TODO use Python 3.6 namedtuple syntax when we drop Python 3.5 support.
+    Status = typing.NamedTuple('Status', [
+        ('iteration', int),
+        ('target_error', float),
+        ('is_completed', bool)
+    ])
+
+    @classmethod
+    def read_status(cls, storage):
```
Lnaden: Question, what happens if you open a netcdf file for read while its also open for write from a running simulation, and then what happens when the read-opened file is closed?

Yank/commands/status.py
andrrizzi: Nope! This should always get dispatched since there are no subcommands.

Yank/repex.py
andrrizzi: I think it's fine. I tried it on my running simulations on the cluster and it worked correctly.

Lnaden:
Okay, I'll merge this into the 0.19.4 pr and we can see if the tests pass

PR #859, opened 2017-12-21
Workaround for travis failures

### LINKED ISSUE #857: Travis build errors
It seems that yank travis nightly builds have been failing for some time, but I never received a build failure warning:
image
@lnaden : Can you look into this when able?

jchodera:
Implements the `--quiet` workaround for travis to fix #857

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,30 +20,30 @@ install:
   - source devtools/travis-ci/install.sh
   - export PYTHONUNBUFFERED=true
   # Add org channel
-  - conda config --add channels ${ORGNAME}
-  # Add conda-forge channel
-  - conda config --add channels conda-forge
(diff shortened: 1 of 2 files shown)


Lnaden:
Even if this works around the problem, I am not sure I like this solution. I think having conda be quiet is a great way to miss build errors/issues.

jchodera:
`--quiet` just suppresses the progress bars:
```
  -q, --quiet           Do not display progress bar.
```
It doesn't suppress the output about which packages will be installed or changed.

Lnaden:
Ah. The only other concern is for massive packages which may take more than 10 min to download, but I don't think those exist or we just don't use them. We'll see if this fixes it. If it does, I'll go touch every other repo I help maintain with this fix

Lnaden:
Looks like this worked. Still odd!

PR #861, opened 2017-12-21
Make the analysis storage chunk equal to checkpoint interval

### LINKED ISSUE #860: Optimize chunksizes in Reporter
For the non-checkpoint file, `Reporter` still uses chunksizes of 1 iteration, which is going to slow down writing unnecessarily. We can easily make these 10 without worry, but the ideal number would be to set them equal to the checkpoint write interval.

Lnaden:
Self-explanatory. Some variables are not chunked to the checkpoint_interval because they are not accessed in bulk usually (online free energy, timestamp).

 This is an optimization for read/write.

 Fixes #860

### DIFF
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -175,6 +175,11 @@ class Reporter(object):
         return {'checkpoint': self._storage_checkpoint, 'analysis': self._storage_analysis}
 
     @property
+    def _storage_chunks(self):
+        """Known NetCDF storage chunk sizes"""
+        return {'checkpoint': 1, 'analysis': self._checkpoint_interval}
+
+    @property
     def analysis_particle_indices(self):
         """Return the tuple of indices of the particles which additional information is stored on for analysis"""
         return self._analysis_particle_indices
@@ -615,7 +620,8 @@ class Reporter(object):
 
             # Create variables and attach units and description.
             ncvar_states = self._storage_analysis.createVariable('states', 'i4', ('iteration', 'replica'),
-                                                                 zlib=False, chunksizes=(1, n_states))
+                                                                 zlib=False,
+                                                                 chunksizes=(self._checkpoint_interval, n_states))
             setattr(ncvar_states, 'units', 'none')
             setattr(ncvar_states, "long_name", ("states[iteration][replica] is the thermodynamic state index "
                                                 "(0..nstates-1) of replica 'replica' of iteration 'iteration'."))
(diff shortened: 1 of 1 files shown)


jchodera:
> Some variables are not chunked to the checkpoint_interval because they are not accessed in bulk usually (online free energy, timestamp).

I'd chunk anything that is *written* each iteration to the same interval. Otherwise, we won't get the speed benefit from not having to enlarge the file each iteration.

Lnaden:
> I'd chunk anything that is written each iteration to the same interval

Fair. The only one I would need to change then is the `timestamp` chunk since its written each interval. Online analysis is on its own interval, so I'll need to think about that a bit more.

Lnaden:
This is currently throwing errors due to the following problem:

https://github.com/Unidata/netcdf4-python/issues/566

Given that this has been open now for a year and a half, I'm not sure if it will ever be fixed.

jchodera:
The code is [here](https://github.com/Unidata/netcdf-c/blob/98dd736a40a9ffae67e2ba9d13ae8bd1b38ce3bd/libsrc4/nc4var.c#L1068-L1089).

I guess we're stuck with chunksizes of 1 for now, but it would be worth contributing what must be the trivial change to fix this back to Unidata.

jchodera:
It looks like you just need to modify [line 1081](https://github.com/Unidata/netcdf-c/blob/98dd736a40a9ffae67e2ba9d13ae8bd1b38ce3bd/libsrc4/nc4var.c#L1081)
from
```C++
            if(var->dim[d]->len > 0 && chunksizes[d] > var->dim[d]->len)
```
to
```C++
            if(! var->dim[d]->unlimited && var->dim[d]->len > 0 && (chunksizes[d] > var->dim[d]->len))
```
Maybe we should try to fix it for them?

Lnaden:
@jchodera I had just written that up  on the parent issue https://github.com/Unidata/netcdf-c/issues/299 then came back here and saw the post. I'll see what they say about any pitfalls in chunking 2 variables on the same dimension differently with existing data.

jchodera:
@lnaden has proposed a PR fix in https://github.com/Unidata/netcdf-c/pull/760

jchodera:
@Lnaden : Can you revive this PR for the `sams` branch when able? Your updated NetCDF version is now live on `conda`.

Lnaden:
Superseded by #912, closing

PR #862, opened 2017-12-21
Decouple number of replicas and states for Reporter

jchodera:
This PR makes minor changes to `Reporter` so that an arbitrary number of replicas can be used.

This is the first step in implementing the SAMS sampler, but it useful to review and merge on its own.

This PR also contains some minor refactoring to reduce duplicated code, leading to overall -13 lines!

### DIFF
--- a/Yank/repex.py
+++ b/Yank/repex.py
@@ -222,7 +222,7 @@ class Reporter(object):
                 open_check_list.append(storage.isopen())
         return np.all(open_check_list)
 
-    def open(self, mode='r', convention='ReplicaExchange'):
+    def open(self, mode='r', convention='ReplicaExchange', netcdf_format='NETCDF4'):
         """
         Open the storage file for reading/writing.
 
@@ -234,8 +234,10 @@ class Reporter(object):
         ----------
         mode : str, Optional, Default: 'r'
             The mode of the file between 'r', 'w', and 'a' (or equivalently 'r+').
-        converntion : str, Optional, Default: 'ReplicaExchange'
+        convention : str, Optional, Default: 'ReplicaExchange'
             NetCDF convention to write
+        netcdf_format : str, Optional, Default: 'NETCDF4'
+            The NetCDF file format to use
 
         """
         # Ensure we don't have already another file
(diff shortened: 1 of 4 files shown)


Yank/repex.py
```diff
@@ -222,7 +222,22 @@ def _are_subfiles_open(self):
                 open_check_list.append(storage.isopen())
         return np.all(open_check_list)
 
-    def open(self, mode='r', convention='ReplicaExchange'):
+    def _create_dimension_if_needed(self, dimname, dimsize):
```
Lnaden: Should this also check that the size of the dim is the size we want to shape it? I.e. raise an error or a warning if the `dimsize` argument does not match the `dimsize` of the exiting dimension?

Yank/repex.py
```diff
@@ -754,11 +769,11 @@ def read_mixing_statistics(self, iteration=slice(None)):
 
         Returns
         -------
-        n_accepted_matrix : kxk numpy.ndarray
+        n_accepted_matrix : numpy.ndarray with shape (n_states, n_states)
```
Lnaden: Is this going to be left for Repex only (`n_states x n_states` swapping)? That is, will the `SAMS` reporter have a different Mixing statistics output, or just a different way its calculated?

Yank/repex.py
jchodera: Great idea! I'll add this.

jchodera: @Lnaden : What kind of `Exception` should we raise if there is a mismatch?

Lnaden: Looking at it some more, it may not be needed since this is an internal function (i.e. the user should never call it so we should know what we are doing as devs when we invoke it). That said, I would make it a `ValueError` by default, but if you wanted to make this a formal assert, it could be an `AssertionError`

Yank/repex.py
jchodera: No, we can still build the empirical transition matrix on `n_states x n_states` for SAMS.

Yank/repex.py
jchodera: I'll use `ValueError`.

jchodera: Addressed!

jchodera:
@Lnaden : The points you raised have been addressed. Please re-review!

Yank/repex.py
```diff
@@ -222,7 +222,34 @@ def _are_subfiles_open(self):
...
+
+        """
+        if dimname not in self._storage_analysis.dimensions:
+            self._storage_analysis.createDimension(dimname, dimsize)
+        else:
+            # Check dimension matches expected size
+            dimension = self._storage_analysis.dimensions[dimname]
+            if dimsize == 0:
+                if not dimension.isunlimited():
+                    raise ValueError("NetCDF dimension '%s' already exists: was previously unlimited, but tried to declare it with size %d" % (dimension.name, dimsize))
```
Lnaden: My only request is to use the Python `.format()` syntax instead of `%` syntax for string formatting. Really minor thing but otherwise is fine!

jchodera: I hate `.format()`.

jchodera:
Formatting requests addressed. Please re-review @Lnaden

jchodera:
@Lnaden : Can you merge when ready? I'm in transit.

PR #864, opened 2017-12-22
Switch analysis from u_kln to u_kn representation

Lnaden:
Simpler than I thought, but this is distinct from the pymbar assumption since the `k` is per sampler, and the `l` is thermodynamic state

All methods except the _prepare_mar_input_data use the u_kln since they only take the input data, truncate/decorrelate, then return it.

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -24,7 +24,7 @@ import os
 import abc
 import copy
 import typing
-from typing import Union
+from typing import Union, Optional
 
 import yaml
 import numpy as np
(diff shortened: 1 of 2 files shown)


Lnaden:
@jchodera after our discussion, I think this is the only function which needed changed, which has to be re-written for any new analyzer anyways with the way the API and class is structured.

Lnaden:
There are also some small syntax errors I fixed for multi-line strings

Yank/analyze.py
```diff
@@ -565,6 +566,50 @@ def get_timeseries(passed_timeseries):
 
         raise NotImplementedError("This class has not implemented this function")
 
+    @staticmethod
+    def sampler_state_sample_to_state_sample(u_kln: np.ndarray, n_k: Optional[np.ndarray]=None) -> np.ndarray:
```
jchodera: This is an extremely confusing name. Can we simplify it?

Issue #69, opened 2014-07-10
Add ability to specify protein point mutations


luirink:
It would be useful to be able to specify protein point mutations rather than feeding a PDB file with the mutation already present.

Lnaden:
Leave this for post 1.0,

Lnaden:
Fixed by #865

PR #865, opened 2017-12-25
Add option to pipeline to use pdbfixer to make protein mutations

jchodera:
Addresses #69 

I'm having a bit of trouble with the tests, but this should otherwise be mostly complete.

For single mutations, simply use the familiar one-letter-code mutation:
```yaml
   molecules:
     Abl:
       make_mutations:
         chain: A
         mutations: T315I
```
For double mutations, use the '/' separator:
```yaml
   molecules:
     EGFR:
       make_mutations:
         mutations: L858R/T790M
```
when no `chain` is specified, it should default to `None`.

If needed, we could add the capability to support mutations in chains with different chain ids. Currently, this is not supported.

I believe you should be able to automatically use `!$Combinatorial` syntax with this as well to provide a list of mutations, which is our intended use case. (Let us know if that isn't the case!)

I don't like the idea of ingesting the one-letter codes and their maps, but they can later be refactored into something else (or, ideally, the functionality to support more flexible syntax put into `pdbfixer` before it is released).

Reviews/feedback welcomed when able!

cc: @steven-albanese


jchodera:
For some reason, the output PDB file is not being generated, but no failure information is given as to why:
```
FAIL: Test that mutations are made correctly.
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/yank-0.19.5-py3.5-linux-x86_64.egg/yank/tests/test_experiment.py", line 1037, in test_make_mutations
    assert os.path.exists(output_path)
AssertionError
```

Lnaden:
I spoke with @steven-albanese about what the implementation for this should look like. I think we should add the option to support a CSV file which support multiple mutations which would be inferred to be run `!Combinatorially` like the multiple molecules in a CSV file are now. This would be in addition to the single mutations as implemented now.

I can take a stab at putting this in after I'm done looking over the SAMS PR if you want

Lnaden:
I made #867 to fix some of the initial problems with this PR.

As discussed in slack, we may want to generalize this function to `pdbfixer` and send keyword arguments beyond just the mutations to it as well. The question is do we want to do that now, or get this merged in first, then make a second pass on the generalization?

Lnaden:
If we decide to put in the generalization now, then we should add the general CSV file for mutation support as well.

jchodera:
Tests pass! Ready to review and merge @Lnaden.

jchodera:
I've significantly generalized this so that essentially the entire PDBFixer pipeline is now integrated. 

The current format for programming mutations is:
```yaml
   molecules:
     Abl:
       filepath: input/abl.pdb
       pdbfixer:
         apply_mutations:
           mutations: T315I
           chain_id: A
```
I believe `!Combinatorial` should work too, so we can try
```yaml
   molecules:
     Abl:
       filepath: input/abl.pdb
       pdbfixer:
         apply_mutations:
           mutations: !Combinatorial [T315I, M244V, F317R]
           chain_id: A
```
We can also use other features of PDBFixer if desired, such as modeling of missing atoms and loops, though this should be used with caution given how simple the loop modeling scheme is:
```yaml
   molecules:
     Abl:
       filepath: input/abl.pdb
       pdbfixer:
         add_missing_residues: yes
         add_missing_atoms: heavy
```
All options are documented in the YAML docs.

The CSV-driven input is a good idea, but due to the difficulty of its implementation, it will be easiest for @steven-albanese to write a short Python script that reads his CSV file of kinase(s), mutants, and inhibitors to generate a single YAML file. For example, I think the following will work for applying many mutations to Abl:
```yaml
molecules:
  Abl:
    filepath: input/abl.pdb
    strip_protons: yes     
    pdbfixer:
      apply_mutations:
        mutations: !Combinatorial [T315I, M244V, F317R]
        chain_id: A
  inhibitors:
    filepath: input/c-Met_bmcl_neutral_docked.sdf
    select: all
    openeye:
      quacpac: am1-bcc
    antechamber:
      charge_method: null```

PR #867, opened 2017-12-27
Fix the initial mutations PR

Lnaden:
I made many changes as a secondary author to this feature. I want to make sure this still does what you intended @jchodera


jchodera:
Thanks!

Issue #14, opened 2013-11-08
Switch to external repositories for Repex and TestSystems (when they are ready)


kyleabeauchamp:
Also, this means that we shouldn't invest too much time cleaning up those files in the current Yank repo.

PS: I think TestSystems is pretty much ready to integrate, so that should happen first.

Lnaden:
Touching for YANK 1.0 release, this will be something handled after 1.0

Lnaden:
Considering how much has changed since this issue has opened. I think we can close this.
OpenMMTools now handles many of the external functions + we have vastly overhauled the sampler code with the SAMS branch

PR #877, opened 2018-01-08
Massive refactor of the `repex.py` file, which no longer exists

Lnaden:
`Repex.py` is dead, long live `sampling`

For review purposes: most of the "changes" are restructuring exiting code into its own subpackage and module.

* Split out the `repex.py` into its own `sampling` subpackage
    * `MultiStateSampler`
    * `MultiStateReporter`
    * `ReplicaExchangeSampler`
    * `ParallelTemperingSampler`
* Divided the classes of `repex.py` into their own modules
* Decoupled most of `sampling` from yank-specific things for eventual migration to OpenMMTools
    * Still remaining: Hooks into MPI module
    * Still remaining: Hooks into a logger config from utils
    * Moved the `SimulationNaNError` to `samplingutils`
    * Custom YAML Loader/Dumpers contained in `sampling`
    * Moved `mixing` to submodule of `sampling`
 * Tried to fix all the tests (they pass locally)
 * Most references to classes in `sampling` should be `sampling.{class}`
 * Some other PEP8 cleanup as I found it

 Things which still need done:
 * Update and verify all Doc pages, but all the code changes should be in

Progress towards #14


jchodera:
`multistate/` might be fine if `multistatesamplers/` is too long.

Lnaden:
> Let's move anything that is generically applicable to a single multistate sampler class into multistatesamplers/

Can you provide an example of what this would be? I thought I moved everything unique to the `MultiStateSampler` and derivatives into the `sampling` directory already, except for some larger, but already decoupled hooks like the `mpi` module

jchodera:
> Can you provide an example of what this would be? I thought I moved everything unique to the MultiStateSampler and derivatives into the sampling directory already, except for some larger, but already decoupled hooks like the mpi module

Sure. From `Yank/analyze.py`, we have:
* [`MultiStateSamplerAnalyzer.generate_mixing_statistics()`](https://github.com/choderalab/yank/blob/b29b0d9ae51db23bb0074cb2ca07aace98e8a454/Yank/analyze.py#L720-L783), which analyzes a single `Reporter` to extract mixing statistics. This is fully general for all `MultiStateSamplers`, and not at all YANK-specific.
* [`MultiStateAnalyzer.show_mixing_statistics()`](https://github.com/choderalab/yank/blob/b29b0d9ae51db23bb0074cb2ca07aace98e8a454/Yank/analyze.py#L785-L833): same
* [`.get_states_energies()`](https://github.com/choderalab/yank/blob/b29b0d9ae51db23bb0074cb2ca07aace98e8a454/Yank/analyze.py#L835-L871): same
* [`.get_timeseries()`](https://github.com/choderalab/yank/blob/b29b0d9ae51db23bb0074cb2ca07aace98e8a454/Yank/analyze.py#L873-L901): same
* [`._prepare_mbar_input_data()`](https://github.com/choderalab/yank/blob/b29b0d9ae51db23bb0074cb2ca07aace98e8a454/Yank/analyze.py#L903-L927): same
* [`._compute_free_energy()`](https://github.com/choderalab/yank/blob/b29b0d9ae51db23bb0074cb2ca07aace98e8a454/Yank/analyze.py#L929-L969): same

and so on. In fact, can't the entirety of [`MultiStateAnalyzer`](https://github.com/choderalab/yank/blob/b29b0d9ae51db23bb0074cb2ca07aace98e8a454/Yank/analyze.py#L701-L1127) and subclasses be migrated, since all of this is generic and depends on a single phase?

Lnaden:
@jchodera Okay, massive migration of the `analyze.py` module has been done. It is also dead, long live `multistate.analyzers`.

Summary of changes:
* Renamed `sampling` to `multistate`
* Moved all of `analyze.py` into `yank.multistate.analyzers` sub module.
    * yank.analyze no longer exists
    * analyzers submodule does not depend on anything outside of `multistate` parent other than logging, which can be cut down later
* Made the ObservablesRegistry an instanced based class with API to register new observables on the fly.
* Fixed all doc links I could find, I think I got them all
* Updated API documentation with all the changes I have made.

Should be ready for another review

Yank/multistate/analyzers/multistateanalyzer.py
```diff
@@ -8,40 +8,48 @@
...
+import logging
 import numpy as np
-
+import simtk.unit as units
 import openmmtools as mmtools
-from .repex import Reporter
 
-from pymbar import MBAR  # multi-state Bennett acceptance ratio
-from pymbar import timeseries  # for statistical inefficiency analysis
+from . import analyzerutils as autils
```
andrrizzi: I'd call these extra utility functions module just `utils` so that people can refer to them as `analyzers.utils` and `multistate.utils` instead of `analyzers.analyzerutils` and `multistate.samplingutils`. I understand this creates files with the same name, but I'm not sure if there are downsides for this.

Yank/yank.py
```diff
@@ -672,7 +671,9 @@ def sampler_states(self):
 
     @abc.abstractmethod
     def create(self, thermodynamic_state, sampler_states, storage,
-               unsampled_thermodynamic_states, metadata):
+               unsampled_thermodynamic_states=None,
+               initial_thermodynamic_states=None,
+               metadata=None):
```
andrrizzi: I wouldn't add default arguments to abstract methods. `AlchemicalPhase` either use them or not so eventual default arguments don't have effect. Also Python doesn't enforce them anyway.

Also, if `AlchemicalPhase` doesn't use `initial_thermodynamic_states` (I haven't found it after a brief search), I would take it away from the contract. This abstract class is only here to document what the user must absolutely implement to have a sampler compatible with `AlchemicalPhase`. Everything else should be up to him.

Lnaden:
Okay, One final take.

I have re-created the analyze.py file which then sub-classes the analyzers in Multistate to add the YANK specific things like `analyze_phase` and `get_standard_state_correction`. All tests have been updated and are passing locally.

I moved the `ObservablesRegistry`  into the main `multistateanalyzers` module and they now hook in directly to any `PhaseAnalyzer` for direct reference. This is also extended in the `analyze.py` file and I added a small test for it as well. 

I think this is ready for one final once over, @jchodera, @andrrizzi 

Thinking about the future for moving `multistate` to OpenMMTools, one hurdle will be the Py2 compatibility. Not something to address now, but I wanted to document it while I'm thinking about it.

Lnaden:
@jchodera I made one final restructure pass based on our discussion. 

I decided to keep `mutlistatesampler.py`, `multistatereporter.py` and `multistateanalyzer.py` as three files as combining them would result in a 4000+ line monolith of a module which would be hard to navigate and edit. I did make `replicaexchange.py` and `paralleltempering.py` single files which build both the sampler and the analyzer. SAMS can follow the same structure.

The `__all__` structure in that module is to make the docs build correctly, which has some mixed results, but when we eventually port it, we can clean that up then.

jchodera:
This looks great! Thanks!

PR #880, opened 2018-01-15
AbsoluteAlchemicalFactory options and CUDA deterministic forces

### LINKED ISSUE #849: Set CUDA platform to use deterministic forces
I just realized we might not be setting CUDA to use deterministic forces/energies. 

We may want to do this at the `openmmtools` level.

The platform property we need to set is `DeterministicForces`:
https://github.com/pandegroup/openmm/blob/master/platforms/cuda/include/CudaPlatform.h#L122-L126

Anytime we are using reweighting methods like MBAR with CUDA, we should be using deterministic energies.

andrrizzi:
I still have to update the `whatsnew` file and the docs, but the implementation is ready for review.
- Add `AbsoluteAlchemicalFactory` options to YAML
- Fix #849: Set CUDA platform to use deterministic forces
- Fixes a bug that we missed when transitioning to Python3 where parallel MPI processes may pick a different counterion to alchemically modify due to the non-deterministic order of dictionaries.

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -2260,6 +2260,10 @@ class ExperimentBuilder(object):
         else:
             platform = openmm.Platform.getPlatformByName(platform_name)
 
+        # Set CUDA DeterministicForces (necessary for MBAR).
+        if platform_name == 'CUDA':
+            platform.setPropertyDefaultValue('DeterministicForces', 'true')
+
         # Use only a single CPU thread if we are using the CPU platform.
         # TODO: Since there is an environment variable that can control this,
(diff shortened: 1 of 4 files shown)


jchodera:
Thanks! This looks good to me.

And thanks for finally getting that deterministic PME in there. That was certainly degrading our PME endpoint overlap a bit, though I'm not sure if it was responsible for the huge effects you are seeing with charged systems.

andrrizzi:
Thanks for the review!

I need to make more changes for the upcoming choderalab/openmmtools#320. I'll merge this and add them to a separate PR (together with the updated docs) to simplify the review.

Issue #678, opened 2017-05-14
Possible optimization to _compute_replica_energies


jchodera:
In explicitly solvated simulations, the time spent in [`_compute_replica_energies()`](https://github.com/choderalab/yank/blob/master/Yank/repex.py#L1920-L1944) can still be significant.

When cycling over different thermodynamic states to compute energies where the only the alchemical parameters are changing but the positions are fixed, it should be possible to achieve a significant speedup by assigning the alchemically-modified `Force` objects to a different force group and only computing the energies for the alchemical force group. This could be added to the static component (independent of alchemical parameters) to compute the total reduced potential.

More generally, this same optimization can be used frequently, especially in multidimensional umbrella sampling with SAMS where only a fast-to-compute umbrella potential is changing during a lot of energy evaluations. I wonder if we could incorporate some sort of automatic caching mechanism into [`ThermodynamicState.reduced_potential()`](https://github.com/choderalab/openmmtools/blob/master/openmmtools/states.py#L530-L622) that handles this under the hood, maybe with some hints that the alchemical parameters all appear in one force group.

andrrizzi:
This was implemented in choderalab/openmmtools#321 and #881 .

PR #881, opened 2018-01-16
Fast computation of the energy matrix

andrrizzi:
This incorporates the changes in choderalab/openmmtools#320 to address #678 .

The Travis tests can only pass once the new version of `openmmtools` is released.

Before merging, I still have to do the following:
- [x] Bump minimum required version of `openmmtools`.
- [x] Update the `whatsnew` file and the YAML docs with these changes and the changes in #880 .

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -1234,13 +1234,24 @@ class ExperimentBuilder(object):
                                'number_of_iterations': integer_or_infinity,
                                'anisotropic_dispersion_cutoff': check_anisotropic_cutoff}
 
-        # Validate parameters
+        # Validate parameters.
         try:
             validated_options = utils.validate_parameters(options, template_options, check_unknown=True,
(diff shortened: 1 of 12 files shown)


andrrizzi:
I've also added a YAML option `max_n_contexts` that allows setting the maximum number of contexts we want to keep in memory at the same time. With these changes, we now have full control over which state should be simulated in its own `Context`.

jchodera:
This is great!

There are two more long-term solutions we might consider to further speed this up or reduce our maintenance burden in future:
1. We could push the tracking of which forces are valid/invalid to the OpenMM layer. Right now, it just invalidates the whole context if an alchemical parameter is changed; if it instead only invalidated affected forces, that would do this for us.
2. Our `CustomNonbondedForces` that use interaction groups currently compute *all* interactions, even those beyond the cutoff. We could request an OpenMM feature where cutoffs are still used to speed up this calculation.

PR #885, opened 2018-01-29
 Bring in changes from master for 0.20.1

Lnaden:
Also fixed the function keyword arg from inherited methods.

@jchodera once this is merged in, you can branch from the `sams` branch to put in the SAMS sampler.

@andrrizzi [here is the inspection](https://github.com/choderalab/yank/blob/21c180550896b42534ff29fbba7a3074292b87a4/Yank/utils.py#L1039-L1101) changes I made to get `kwargs`'s defaults from inherited functions. I'm not sure we can do better than this, but I think this is what we need.

### DIFF
--- a/Yank/utils.py
+++ b/Yank/utils.py
@@ -1059,9 +1059,7 @@ def get_keyword_args(function_to_inspect, try_mro_from_class=None):
 
     """
 
-    master_mro_stack = []
-
-    def stack_up_argspec(input_argspec):
+    def extract_kwargs(input_argspec):
         defaults = input_argspec.defaults
         if defaults is None:
             defaults = []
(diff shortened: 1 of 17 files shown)


jchodera:
There's a lot here, but it looks good to me.

Lnaden:
Far less here than it appears, its the merge of what is currently in `master` to the `sams` branch to make sure this branch here (`sams`) is still up to date with the changes we make in in `master`

Yank/utils.py
```diff
@@ -1051,9 +1058,46 @@ def get_keyword_args(function):
...
+        n_args = len(input_argspec.args)
+        # Cycle through the kwargs only
+        cycle_kwargs = input_argspec.args[n_args - n_defaults:]
+        cycle_kwargs = {arg: value for arg, value in zip(cycle_kwargs, defaults)}
+        # Handle the kwonlyargs for calls with `def F(a,b *args, x=True, **kwargs)
+        if input_argspec.kwonlydefaults is not None:
+            cycle_kwargs = {**cycle_kwargs, **input_argspec.kwonlydefaults}
+        return cycle_kwargs
+
+    def build_argspec(input_function, input_cls):
```
andrrizzi: I'm not sure there's the need for recursion and to keep track of the explored classes as `getmro()` should already transform the recursive part of the problem into loop form.

Maybe there's something I'm missing but, what happens with something like this:
```python
argspec = inspect.getfullargspec(function)
kwargs = stack_up_argspec(argspec)

if mro_cls is not None and argspec.varkw is not None:
    try:
        mro = inspect.getmro(mro_cls)
    except AttributeError:
        # No MRO, input_function is not part of a class
        mro = [mro_cls]

    for cls in mro[1:]:
        try:
            parent_function = getattr(cls, function.__name__)
        except AttributeError:
            # class of the MRO does not have the method
            pass
        else:
            argspec = inspect.getfullargspec(parent_function)
            kwargs = {**stack_up_argspec(argspec), **kwargs}
return kwargs
```

Lnaden: Sure, each step of the recursion internally will recursively do the right thing, but its redundant. consider an MRO stack of classes `(X, Y, Z)`.
`X` will queue `Y` and `Z` to search, but `Y` will *also* search `Z`, so `Z` gets searched twice, yields the same result, and the `X` search through `Z` will replace the `Y` search through `Z`. So I made a master MRO list to avoid the extra work.

I see what you mean about the non-recursive search. I'll fix that

Lnaden: I remembered why I wanted the recursion, its so that the `argspec.varkw` could be checked for each MRO and terminated as needed. But I also realize that the way I had it was insufficient since they would all be checked anyways, unless all but one class did not have the `varkw` (`**kwargs)`. It might be fixable with a bunch of cross-mro checks inspecting diffs of the mro lists, but probably not worth putting in at this time.

Lnaden:
Tests pass locally, I'm going to merge since the tests already passed before the minor change I made.

PR #887, opened 2018-01-30
[DO NOT MERGE] Temporarily add script for unbiasing the restraint

andrrizzi:
In this script, I've extended the old `ReplicaExchangeAnalyzer` to unbias a radially symmetric restraint. I'll think about the final design and push the changes to the main analysis code once it is stable again and we have experimented with this a little. For now, this can be used similarly to the current `yank analyze` command
```
python unbiased_analyzer.py -s "path/to/experiment/directory/" -v --energy-cutoff=3
```
the `--energy-cutoff` argument is in kT and optional.

A couple of things I'd like to run by someone to verify that I'm doing the correct thing.
1. I'm unbiasing both the `lambda=1` and the `lambda=0` state by adding two extra unsampled states to the energy matrix.
2. I'm "unbiasing" also the standard state correction, which basically means that the restraint is reduced to a squared well, where the hard wall is placed according to the given cutoff. If no cutoff is provided, I'm computing the standard deviation as `-log(box_volume/standard_volume)`.

One thing I realized working on this is that our current design doesn't make it easy to read information from and manipulate our custom forces. It currently requires a lot of poorly extensible guesswork. I think we should encapsulate some of the logic currently inside our factories in actual force objects. I'll open an issue on `openmmtools` to discuss this separately.

### DIFF
--- a/scripts/unbiased_analyzer.py
+++ b/scripts/unbiased_analyzer.py
@@ -1,0 +1,609 @@
+#!/usr/local/bin/env python
+
+import os
+import copy
+import scipy
+import logging
+
+import yaml
+import numpy as np
+
+import mdtraj
+import pymbar
+from simtk import openmm, unit as unit
+import openmmtools as mmtools
+import yank.restraints
+from yank import analyze, repex, utils
+
+logger = logging.getLogger(__name__)
+
+
+def get_restraint_force(system):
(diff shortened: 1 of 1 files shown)


andrrizzi:
This can be closed now that this analysis has been integrated into YANK (see #915).

PR #888, opened 2018-02-02
Add neighborhood support to MultiStateSampler and ReplicaExchange

jchodera:
As discussed, I'm working on first adding support for restricted neighborhoods to the existing samplers before implementing SAMS.

Presuming we keep the current order of mix-propagate-energy, my first hurdle is to determine how to store the neighborhood that was used to compute the replica energies at a subset of thermodynamic states.

@andrrizzi @Lnaden : I'd appreciate your advice here! Should I:
1. Extend [`read_energies` and `write_energies`](https://github.com/choderalab/yank/blob/sams-neighborhood/Yank/multistate/multistatereporter.py#L690-L778) to also read and write an `[n_iterations, n_replicas, n_states]` integer array that is 1 if the energy was computed for that state and 0 if not?
2. Modify the current energy read/write code to use [masked arrays](http://unidata.github.io/netcdf4-python/#netCDF4.Variable.set_auto_mask) and simply mask the values of the existing [`energies`](https://github.com/choderalab/yank/blob/sams-neighborhood/Yank/multistate/multistatereporter.py#L742-L747) matrix that are not computed?

I'm thinking option (2) might be simpler, though I've never used masked arrays before. Has anyone had experience with these?

### Progress

- [x] Add `_neighborhood(state_index)` method to compute local neighborhood
- [x] Restrict energy computation to neighborhood
- [x] Store energies only within neighborhood, alongside neighborhood information
- [x] Add test to ensure that energy array is properly stored when locality is restricted
- [x] Test restricted neighborhood `MultiStateSampler`
- [x] Implement fast online analysis
- [x] Implement locality support in replica exchange and parallel tempering
- [ ] Implement locally weighted histogram analysis for offline analysis: [paper1](http://www.stat.rutgers.edu/home/ztan/Publication/SAMS_preprint.pdf),  [paper2](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4729400/pdf/JCPSA6-000144-034107_1.pdf) - bumping this to a subsequent PR

### DIFF
--- a/Yank/multistate/multistatereporter.py
+++ b/Yank/multistate/multistatereporter.py
@@ -54,6 +54,8 @@ logger = logging.getLogger(__name__)
 # MULTISTATE SAMPLER REPORTER
 # ==============================================================================
 
+MISSING_VALUE = np.NaN # value for populating missing/masked values
+
 class MultiStateReporter(object):
     """Handle storage write/read operations and different format conventions.
 
@@ -707,7 +709,6 @@ class MultiStateReporter(object):
         """
         iteration = self._calculate_last_iteration(iteration)
         energy_thermodynamic_states = self._storage_analysis.variables['energies'][iteration, :, :]
-        energy_thermodynamic_states = self._storage_analysis.variables['energies'][iteration, :, :]
         try:
             energy_unsampled_states = self._storage_analysis.variables['unsampled_energies'][iteration, :, :]
         except KeyError:
@@ -745,6 +746,8 @@ class MultiStateReporter(object):
                                                                    ('iteration', 'replica', 'state'),
                                                                    zlib=False,
                                                                    chunksizes=(1, n_replicas, n_states))
+            ncvar_energies.missing_value = MISSING_VALUE
+            ncvar_energies.set_auto_mask(True)
             ncvar_energies.units = 'kT'
             ncvar_energies.long_name = ("energies[iteration][replica][state] is the reduced (unitless) "
                                         "energy of replica 'replica' from iteration 'iteration' evaluated "
--- a/Yank/multistate/multistatereporter.py
+++ b/Yank/multistate/multistatereporter.py
@@ -54,7 +54,7 @@ logger = logging.getLogger(__name__)
 # MULTISTATE SAMPLER REPORTER
 # ==============================================================================
 
-MISSING_VALUE = np.NaN # value for populating missing/masked values
+MISSING_VALUE = np.inf # value for populating missing values
 
 class MultiStateReporter(object):
     """Handle storage write/read operations and different format conventions.
@@ -749,7 +749,7 @@ class MultiStateReporter(object):
                                                                    'f8',
                                                                    ('iteration', 'replica', 'state'),
                                                                    zlib=False,
-                                                                   fill_value=0.0,
+                                                                   fill_value=MISSING_VALUE,
                                                                    chunksizes=(1, n_replicas, n_states))
             ncvar_energies.units = 'kT'
             ncvar_energies.long_name = ("energies[iteration][replica][state] is the reduced (unitless) "
@@ -1116,9 +1116,9 @@ class MultiStateReporter(object):
             online_group = analysis_nc.createGroup('online_analysis')
             # larger chunks, faster operations, small matrix anyways
             online_group.createVariable('f_k', float, dimensions=('iteration', 'f_k_length'),
-                                        zlib=True, chunksizes=(1, len(f_k)), fill_value=np.inf)
+                                        zlib=True, chunksizes=(1, len(f_k)), fill_value=MISSING_VALUE)
             online_group.createVariable('free_energy', float, dimensions=('iteration', 'Df'),
-                                        zlib=True, chunksizes=(1, 2), fill_value=np.inf)
+                                        zlib=True, chunksizes=(1, 2), fill_value=MISSING_VALUE)
         online_group = analysis_nc.groups['online_analysis']
         online_group.variables['f_k'][iteration] = f_k
         online_group.variables['free_energy'][iteration, :] = free_energy
--- a/Yank/multistate/multistatereporter.py
+++ b/Yank/multistate/multistatereporter.py
@@ -710,9 +710,16 @@ class MultiStateReporter(object):
             ``sampler_states[iteration, i]`` and ThermodynamicState ``unsampled_thermodynamic_states[iteration, j]``.
 
         """
+        # Determine last consistent iteration
         iteration = self._calculate_last_iteration(iteration)
+        # Retrieve energies at all thermodynamic states
         energy_thermodynamic_states = np.array(self._storage_analysis.variables['energies'][iteration, :, :], np.float64)
-        energy_neighborhoods = np.array(self._storage_analysis.variables['neighborhoods'][iteration, :, :], 'i1')
+        # Retrieve neighborhoods, assuming global neighborhoods if reading a pre-neighborhoods file
+        try:
+            energy_neighborhoods = np.array(self._storage_analysis.variables['neighborhoods'][iteration, :, :], 'i1')
+        except KeyError:
+            energy_neighborhoods = np.ones(energy_thermodynamic_states.shape, 'i1')
+        # Read energies at unsampled states, if present
         try:
             energy_unsampled_states = np.array(self._storage_analysis.variables['unsampled_energies'][iteration, :, :], np.float64)
         except KeyError:
--- a/Yank/multistate/multistatereporter.py
+++ b/Yank/multistate/multistatereporter.py
@@ -707,6 +707,7 @@ class MultiStateReporter(object):
         """
         iteration = self._calculate_last_iteration(iteration)
         energy_thermodynamic_states = self._storage_analysis.variables['energies'][iteration, :, :]
+        energy_thermodynamic_states = self._storage_analysis.variables['energies'][iteration, :, :]
         try:
             energy_unsampled_states = self._storage_analysis.variables['unsampled_energies'][iteration, :, :]
         except KeyError:
--- a/Yank/multistate/multistatereporter.py
+++ b/Yank/multistate/multistatereporter.py
@@ -54,7 +54,7 @@ logger = logging.getLogger(__name__)
 # MULTISTATE SAMPLER REPORTER
 # ==============================================================================
 
-MISSING_VALUE = np.inf # value for populating missing values
+MISSING_VALUE = np.NaN # value for populating missing values
 
 class MultiStateReporter(object):
     """Handle storage write/read operations and different format conventions.
--- a/Yank/multistate/multistatereporter.py
+++ b/Yank/multistate/multistatereporter.py
@@ -702,22 +702,26 @@ class MultiStateReporter(object):
         energy_thermodynamic_states : n_replicas x n_states numpy.ndarray
             ``energy_thermodynamic_states[iteration, i, j]`` is the reduced potential computed at
             SamplerState ``sampler_states[iteration, i]`` and ThermodynamicState ``thermodynamic_states[iteration, j]``.
+        energy_neighborhoods : n_replicas x n_states numpy.ndarray
+            energy_neighborhoods[replica_index, state_index] is 1 if the energy was computed for this state,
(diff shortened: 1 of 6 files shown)


Lnaden:
I would prefer the `masked` arrays myself since they are doing exactly what you want to do with (1); I have used them somewhat in the past. However, I have also found NetCDF tends to mangle the non-masked entries and you don't always get out the original matrix. I've not spent very long digging into that though.

andrrizzi:
Solution 2 sounds better to me, but I'm not sure if netcdf can handle saving also the `array.mask` attribute automatically.

Lnaden:
NetCDF Supports a `fill`, and it knows how to interpret numpy masked arrays, but I have seen it mangle values where are not supposed to be masked, filling them in when they should not be. So the round-trip np->nc->np comes out wrong

jchodera:
I should be able to safely fill with NaN, since we catch any NaNs that occur in energy computation separately.

Lnaden:
That is actually a good idea. You would need to tell the reporter to convert that info back on fetching, but thats kind of the point of the reporter. I think so long as its as Python Numpy Array, it should be a masked one (`numpy.ma.masked_equal` is a great function for this)

jchodera:
I've abandoned the masked array idea because I couldn't figure out how to get it to work. Instead, I'm explicitly storing the neighborhood mask in the NetCDF file.

jchodera:
@Lnaden : I think I want to make a clearer distinction between online and offline estimators.
I believe we can use (1) to implement an inexpensive online estimator that will work for `MultiStateSimulation`, `ReplicaExchange`, `ParallelTempering`, and `SAMS`, and (2) (and MBAR) to implement a general offline estimator that works for everything else.

I'm taking a stab at (1) over the next day.

1. Zhiqiang Tan. Optimally Adjusted Mixture Sampling and Locally Weighted Histogram Analysis. Journal of Computational and Graphical Statistics 26:54, 2017. [DOI](https://doi.org/10.1080/10618600.2015.1113975)
2. Tan, Z., Xia, J., Zhang, B.W., and Levy, R.M. Locally weighted histogram analysis and stochastic solution for large-scale multistate free energy estimation. Journal of Chemical Physics 144:034107. [DOI](https://doi.org/10.1063/1.4939768)

jchodera:
I haven't managed to get all the bugs out yet, but this shows some of the restructuring I was thinking about. Let's chat about this in person @Lnaden @andrrizzi

Yank/multistate/multistatesampler.py
```diff
@@ -1140,14 +1188,18 @@ def _compute_energies(self):
     def _compute_replica_energies(self, replica_id):
         """Compute the energy for the replica in every ThermodynamicState."""
         # Initialize replica energies for each thermodynamic state.
-        energy_thermodynamic_states = np.zeros(self.n_states)
+        energy_thermodynamic_states = np.ma.zeros(self.n_states)
```
Lnaden: You said you got rid of all masks, but there is one here. Since it was causing some undefined behavior in your tests according to you, then I don't think you intended for this to still be here. Probably best to remove it so it does not cause other unforeseen issues

Yank/multistate/multistatesampler.py
```diff
@@ -1140,14 +1188,18 @@ def _compute_energies(self):
...
         # Retrieve sampler state associated to this replica.
         sampler_state = self._sampler_states[replica_id]
 
+        # Determine neighborhood
+        state_index = self._replica_thermodynamic_states[replica_id]
+        neighborhood = self._neighborhood(state_index)
+
         # Compute energy for all thermodynamic states.
-        for energies, states in [(energy_thermodynamic_states, self._thermodynamic_states),
+        for energies, states in [(energy_thermodynamic_states, self._thermodynamic_states[neighborhood]),
```
Lnaden: Don't you also need to slice the `energy_thermodynamic_states` here? Otherwise wont the line `for energy_idx, state_idx in enumerate(state_indices)` ([absolute reference to line](https://github.com/choderalab/yank/blob/18929cc0c66d498a131d3190e20f1f252e98cfcb/Yank/multistate/multistatesampler.py?utf8=%E2%9C%93#L1219)) fail, or at least assign energies to the wrong place?

Yank/multistate/paralleltempering.py
```diff
@@ -226,7 +226,7 @@ class ParallelTemperingAnalyzer(ReplicaExchangeAnalyzer):
     See Also
     --------
     ReplicaExchangeAnalyzer
-    PhaseAnalyzer
+    MultiStateSamplerAnalyzer
```
Lnaden: Did you mean to remove `PhaseAnalyzer` as well? I don't think the docs fill in all the methods inherited from the parent class(es), so I wanted to leave `PhaseAnalyzer` in place to so people could jump to it.

If you dont like the idea of documented ABC's, we can try to think of something better for the docs.

Yank/multistate/replicaexchange.py
```diff
@@ -209,6 +209,9 @@ def _repex_mixing_scheme_validator(instance, replica_mixing_scheme):
             if replica_mixing_scheme not in supported_schemes:
                 raise ValueError("Unknown replica mixing scheme '{}'. Supported values "
                                  "are {}.".format(replica_mixing_scheme, supported_schemes))
+            if instance.locality != None:
```
Lnaden: Small pep8 thing, you should be able to do `if ... and ....` since its back to back `if`s. Assuming you dont plan to do anything else between the two `if`s here

Yank/multistate/replicaexchange.py
```diff
@@ -352,6 +352,8 @@ def _mix_neighboring_replicas(self):
         """Attempt exchanges between neighboring replicas only."""
         logger.debug("Will attempt to swap only neighboring replicas.")
 
+        # TODO: Extend this to allow more remote swaps or more thorough mixing if locality > 1.
```
Lnaden: I really think we can easily adapt the existing Cython code since you have to feed it all the data manually, so we can just slice out the `states` we can swap to, and run it through the code

Yank/multistate/multistatereporter.py
```diff
@@ -54,7 +54,7 @@
 # MULTISTATE SAMPLER REPORTER
 # ==============================================================================
 
-MISSING_VALUE = np.NaN # value for populating missing/masked values
+MISSING_VALUE = np.inf # value for populating missing values
```
Lnaden: I'm not sure `inf` is a good idea here since we can have real states with `inf` energy (think decoupled atom on top of normal atom evaluated in fully coupled state)

Yank/multistate/replicaexchange.py
jchodera: Extending the `all-swaps` cython code is certainly nontrivial. At the very least, it would require explicitly *rejecting* proposed moves that propose an exchange where one of the replicas has no energy evaluated at a new state, which means we have to also feed in the neighborhoods matrix.

We can also extend the `neighbor-swap` scheme to allow more remote proposals if `locality > 1`, but the acceptance probability falls off more rapidly.

I think these are great things to explore, but probably in a separate PR.

Yank/multistate/replicaexchange.py
jchodera: I don't understand.

Yank/multistate/paralleltempering.py
jchodera: Fixed!

Yank/multistate/multistatesampler.py
jchodera: Good catch!

Yank/multistate/multistatesampler.py
jchodera: Good catch!

Yank/multistate/multistatesampler.py
```diff
@@ -1021,6 +1039,30 @@ def _store_options(self):
...
+        -------
+        neighborhood : list of int
+            The states in the local neighborhood
+        """
+        if self.locality == None:
+            # Global neighborhood
+            return slice(0, self.n_states)
+        else:
+            # Local neighborhood specified by 'locality'
+            return slice(max(0, state_index - self.locality), min(self.n_states, state_index + self.locality + 1))
```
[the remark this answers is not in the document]
jchodera: Yes.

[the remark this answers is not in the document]
jchodera: That should really me 0...(n_states-1)

Yank/multistate/multistatereporter.py
```diff
@@ -771,8 +788,9 @@ def write_energies(self, energy_thermodynamic_states, energy_unsampled_states, i
                                                  "(unitless) energy of replica 'replica' from iteration 'iteration' "
                                                  "evaluated at unsampled thermodynamic state 'state'.")
 
-        # Store states energy.
-        self._storage_analysis.variables['energies'][iteration, :, :] = energy_thermodynamic_states[:, :]
+        # Store values
+        self._storage_analysis.variables['energies'][iteration,:,:] = energy_thermodynamic_states
+        self._storage_analysis.variables['neighborhoods'][iteration,:,:] = energy_neighborhoods
         if energy_unsampled_states.shape[1] > 0:
```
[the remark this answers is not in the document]
jchodera: Yes. That behavior is unchanged.

Yank/multistate/multistatereporter.py
```diff
@@ -700,62 +702,77 @@ def read_energies(self, iteration=slice(None)):
...
 
-            # Create dimensions if they weren't created by other functions.
-            self._ensure_dimension_exists('replica', n_replicas)
-            self._ensure_dimension_exists('state', n_states)
-
-            # Create variable for thermodynamic state energies with units and descriptions.
+        # Create dimensions and variables if they weren't created by other functions.
+        self._ensure_dimension_exists('replica', n_replicas)
+        self._ensure_dimension_exists('state', n_states)
+        if 'energies' not in self._storage_analysis.variables:
```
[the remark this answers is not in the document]
jchodera: The logic of lazy creation was changed so that things are created as they are needed.
One hope was that this would let us silently upgrade old-style netcdf files by adding dense neighborhoods if we try to resume with the post-SAMS code.

Yank/multistate/replicaexchange.py
Lnaden: `if instance.locality != None and replica_mixing_scheme not in ['swap-neighbors']` instead of `if` on one line, and a second `if` on another

Yank/multistate/replicaexchange.py
Lnaden: > I think these are great things to explore, but probably in a separate PR.

great idea!

jchodera:
I think that this is much as is sensible to do for this PR. We still need to update the analyzer to handle non-global locality, but that really needs an implementation of Local WHAM first.

Should we:
* Wait on this PR to add local WHAM to pymbar?
* Raise an exception if the analyzer is applied to a simulation with non-global locality, since this isn't supported yet, and merge to continue the SAMS implementation?

Also, do we yet have an API point for retrieving the latest *online* estimate? Or do we still have to dig through the file right now?

Lnaden:
I'm in favor of the 2nd option, get actual changes in and just raise for not implemented, reduces number of branches we have to track.

As for getting the latest online free energy estimates:
https://github.com/choderalab/yank/blob/sams-neighborhood/Yank/multistate/multistatereporter.py#L1056

Which can of course be improved substantially

andrrizzi:
+1 on number 2 option for now as well.

jchodera:
@Lnaden @andrrizzi : This should be ready to review. Anything else we want to include in this PR, or should we merge and continue with the L-WHAM addition to analysis and SAMS addition to samplers?

jchodera:
I've punted on that issue. I think we still want to only validate at create() time, but for now, I've just changed things in the "right" order.

jchodera:
Maybe we should open an issue to revisit the parameters validation stuff later?

Lnaden:
> Maybe we should open an issue to revisit the parameters validation stuff later?

Sounds good. I think this is ready otherwise!

jchodera:
Merging this now. We'll have to bump L-WHAM offline analysis to a subsequent PR.

I think @Lnaden was going to look into implementing that within `pymbar`.

PR #893, opened 2018-02-20
Implement SAMS sampler

jchodera:
This supercedes #886.

@Lnaden : I need your help adding writing/reading logZ to the storage layer in an appropriately modular way. I'd like to store it without having to implement a ton of extra code to subclass things. We have a couple of options:

Option 1: Reuse `log_f_k`. It could make sense to co-opt the online analysis framework for SAMS, since it provides its own online analysis, but it would also require overriding or restructuring the way we do online analysis right now, which is very MBAR-centric. I'd like to overhaul that, but maybe later. Note that `logZ[k] = - log_f_k[k]` in my convention, so we can easily store one or the other---no real need to store both.

Option 2: Make it easier to extend what kinds of data are stored to / resumed from the storage. Is there some way we can make it really easy for subclasses to simply specify they want to store more data without having to subclass a ton of things? Right now, it seems like this would require
* Creating a new `SAMSReporter` that has its own copy of `__init__`
* Implementing `write_X` and `read_X` in `SAMSReporter`
* Implementing `from_storage` and `_report_iteration` to first call `MultiStateSampler` versions of these methods and then `read_X` or `write_X`

There may still be other bugs to squish, but resuming/storing `logZ` is the biggest challenge so far.

Once we get this implemented, we still need a way to test, since there is no way to select the sampler via the YAML file yet.

### DIFF
--- a/Yank/multistate/multistatereporter.py
+++ b/Yank/multistate/multistatereporter.py
@@ -1060,6 +1060,49 @@ class MultiStateReporter(object):
         """
         self._storage_analysis.variables['last_iteration'][0] = iteration
 
+    def read_logZ(self, iteration):
+        """
+        Read logZ at a given iteration from file.
+
+        Parameters
+        ----------
+        iteration : int
+            iteration to read the free energies from
+
+            if the iteration was not written at a the given iteration, then the free_energies are all 0
+
+        Returns
+        -------
+        logZ : np.array with shape [n_states]
+            Dimensionless logZ
+        """
+        online_group = self._storage_analysis.groups['online_analysis']
+        logZ = online_group.variables['logZ'][iteration]
+        return logZ
+
+    def write_logZ(self, iteration: int, log_Z: np.ndarray):
+        """
+        Write logZ
+
+        Parameters
+        ----------
+        iteration : int,
+            Iteration at which to save the free energy.
+            Reads the current energy up to this value and stores it in the analysis reporter
+        logZ : np.array with shape [n_states]
+            Dimensionless logZ
+        """
+        analysis_nc = self._storage_analysis
+        if 'logZ' not in analysis_nc.dimensions:
+            analysis_nc.createDimension('logZ_length', len(logZ))
+            if 'online_analysis' not in analysis_nc.groups:
+                online_group = analysis_nc.createGroup('online_analysis')
+                # larger chunks, faster operations, small matrix anyways
+                online_group.createVariable('logZ', float, dimensions=('iteration', 'logZ_length'),
+                                            zlib=True, chunksizes=(1, len(logZ)), fill_value=MISSING_VALUE)
+        online_group = analysis_nc.groups['online_analysis']
+        online_group.variables['logZ'][iteration] = logZ
+
     def read_mbar_free_energies(self, iteration):
         """
         Read the MBAR dimensionless free energy at a given iteration from file.
--- a/Yank/multistate/multistatereporter.py
+++ b/Yank/multistate/multistatereporter.py
@@ -1275,7 +1275,7 @@ class MultiStateReporter(object):
             variable_parameters = self._determine_netcdf_variable_parameters(iteration, data, storage)
             storage.createVariable(variable, variable_parameters['dtype'],
                                    dimensions=variable_parameters['dims'],
-                                   cunksizes=variable_parameters['chunks'],
+                                   chunksizes=variable_parameters['chunks'],
                                    zlib=False)
         # Get the variable
         nc_var = storage[variable]
--- a/Yank/multistate/multistatereporter.py
+++ b/Yank/multistate/multistatereporter.py
@@ -9,7 +9,7 @@ Multistatereporter
 ==================
 
 Master multi-thermodynamic state reporter module. Handles all Disk I/O
-reporting operations for any MultiStateSampeler derived classes.
+reporting operations for any MultiStateSampler derived classes.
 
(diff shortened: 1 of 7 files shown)


Yank/multistate/multistatereporter.py
```diff
@@ -1060,6 +1060,49 @@ def write_last_iteration(self, iteration):
         """
         self._storage_analysis.variables['last_iteration'][0] = iteration
 
+    def read_logZ(self, iteration):
```
jchodera: We don't necessarily want to keep these `read_logZ/write_logZ`, or keep them here if we want to keep them.

Yank/multistate/sams.py
```diff
@@ -0,0 +1,593 @@
...
+                raise Exception('Initial log_Z_guess (dim {}) must have same number of states as n_states ({})'.format(len(log_Z_guess), self.n_states))
+            self._logZ = np.array(self.log_Z_guess, np.float64)
+
+        # Update log weights
+        self._update_log_weights()
+
+
+    # TODO: Get rid of from_storage() when we move read_logZ and write_logZ to multistatesampler.py
+    @classmethod
+    def from_storage(cls, storage):
```
jchodera: This can be eliminated if we can just use existing `log_f_k` storage.

Yank/multistate/sams.py
```diff
@@ -0,0 +1,593 @@
...
+        """
+        sampler = MultiStateSampler.from_storage(storage)
+
+        sampler._reporter.open(mode='a')
+        self._logZ = sampler._reporter.read_logZ()
+        sampler._reporter.close()
+
+        return sampler
+
+    def _report_iteration(self):
```
jchodera: This can be eliminated if we can just use existing `log_f_k` storage.

Yank/multistate/sams.py
```diff
@@ -0,0 +1,593 @@
...
+            if self._stage == 'initial':
+                beta_factor = 0.4
+                t = self._iteration + 1.0
+                gamma = min(pi_k[state_index], t**(-beta_factor)) # Eq. 15
+            elif self._stage == 'asymptotically-optimal':
+                gamma = 1.0 / float(self._iteration - self._t0 + 1./self.gamma0) # prefactor in Eq. 9 and 12 from [1]
+            else:
+                raise Exception('Programming error:unreachable code')
+
+            # TODO: Store gamma for each replica: self.ncfile.variables['gamma'][self.iteration] = gamma
```
jchodera: We also want to store `gamma` here.

jchodera:
Option 2 may be better, since there are some other bits of information we want to store each iteration, like `gamma`:
https://github.com/choderalab/yank/pull/893/files#diff-a3a2806a37b6877068dfc229ba0c1273R546

Lnaden:
@jchodera here is my pass at generalizing online data storage in both per-iteration, and non-per-iteration based approaches. The main hook that specific reporters or samplers can use is in the `multistatereporter.py` file in the `write/read_online_analysis_data` functions which generalizes the input to keywords and dynamically creates dims/variables as needed.

Per-iteration data is stored optionally if the iteration passed is `None` or an `int`, and variable names are chosen dynamically by the functions themselves so the user does not have to. I also added a helper function `write_online_data_dynamic_and_static` which calls the `write...` function twice, setting both per-iteration and non-per-iteration data of the same type. Slightly doubles up but prevents the need to search the the GBs of file with minimal duplication. Name could use work though.

I left the existing frameworks as comments for `online_analysis`, `logZ`, etc, but those could mostly be removed.

Your comment on the `report_iteration` and the `from_storage` functions:  I don't think they can be removed, especially the `from_storage` function since it initializes much more than just the last log_z's

> We also want to store gamma here. 

I'm not really sure why gamma needs stored since nothing restores from it and it seems to be a cheap calculation used only in `_update_logZ_estimates`, and only loosely related to `log_Gamma_j_L`, although never used in calculation of one another. I added a line to  [store the gammas here](https://github.com/choderalab/yank/pull/893/files#diff-a3a2806a37b6877068dfc229ba0c1273R570) but I don't see how they are used, or even if thats the correct place to drop the storage of it.

jchodera:
Thanks! Should I go ahead and clean up the rest of the test failures now?

jchodera:
@Lnaden : One issue I'm running into is that I use `self._replica_P_k` as a temporary variable that isn't meant to persist in storage, but this seems to cause problems in our `test_from_storage()` test:
```
ERROR: Test that from_storage completely restore the sampler.
  File "/Users/choderaj/miniconda/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/choderaj/github/choderalab/yank/yank/Yank/tests/test_sampling.py", line 1061, in test_from_storage
    restored_value = restored_dict.pop(attr)
KeyError: '_replica_log_P_k'
```
Any suggestions on the best way to handle this?

jchodera:
I've fixed some issues with the tests that caused them to erroneously fail with the `SAMSSampler` when only one replica was used, but there are a number of other failures we have to address yet.

jchodera:
Woohoo!

Lnaden:
If this passes, the SAMS sampler will be in, then its just YAML hooks and analyzer

Lnaden:
Hey! Tests pass!
=

jchodera:
I'd like to merge this and #899 (when tests pass) and then we can clean everything up in the `sams` branch before merging it into `master`.

PR #897, opened 2018-02-20
Augment --jobid to take the contents of an environment variable if desired

jchodera:
This adds the ability to specify `--jobid=ENVNAME`, where `ENVNAME` is the name of a zero-indexed environment variable.

How do we test `--jobid` in the units tests? I was looking to add a unit test, but couldn't find where the `--jobid` tests were.

This feature is needed for ORNL TITAN.

### DIFF
--- a/Yank/commands/script.py
+++ b/Yank/commands/script.py
@@ -105,11 +105,20 @@ def dispatch(args):
                 job_id = int(os.environ[job_id]) + 1
             else:
                 raise ValueError("--jobid '%s' unknown" % job_id)
-
     else:
         job_id = None
+
     if args['--njobs']:
-        n_jobs = int(args['--njobs'])
+        n_jobs = args['--njobs']
+        try:
+            # Try processing it as an integer
+            n_jobs = int(n_jobs)
+        except ValueError:
+            if n_jobs in os.environ:
+                # Take the value of the environment variable
+                n_jobs = int(os.environ[n_jobs])
+            else:
+                raise ValueError("--njobs '%s' unknown" % njobs)
     else:
         n_jobs = None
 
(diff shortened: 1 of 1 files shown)


jchodera:
@Lnaden @andrrizzi : This test failure is stochastic. Can we instead get this merged in before we merge in `sams` and cut a new release? Otherwise, I keep having to make these changes locally.

Yank/commands/script.py
```diff
@@ -38,7 +38,11 @@
                                 1 <= job_id <= n_jobs. In this case, njobs must be specified as well and YANK will
                                 run only 1/n_jobs of the experiments. This can be used to run several separate YANK
                                 executions in parallel starting from the same script.
+                                Alternatively, a zero-indexed environment variable name can be specified, in which
+                                case the jobid will be taken from this variable (0 <= environment variable < n_jobs).
```
andrrizzi: As I mentioned above, I don't think this will work. The index must be `1 <= jobid <= njobs` or `ExperimentBuilder` will complain.

jchodera: I changed [the code](https://github.com/choderalab/yank/pull/897/files#diff-a415ceb8df1d0ac62f46ca6dd0d5bb71R107) to add 1 after you pointed this out! I've confirmed it works on titan.

andrrizzi: Ah! Sorry, I didn't notice the `+1`. Feel free to merge then!

andrrizzi:
I think the `jobid`/`njobs` arguments are passed to the `ExperimentBuilder` constructor below those lines, but it is simply not shown in the diff.

andrrizzi:
All those new Travis errors are there because of the new `openmmtools` release, but I've already fixed them in the MCMC moves PR so don't worry about them.

Issue #898, opened 2018-02-27
Allow to specify the MultiStateSampler and MCMCMoves in the YAML script


andrrizzi:
Picking up from the discussion in #863, what do you think about this?
```yaml
mcmc_moves:
    my-sequence-move:
        type: SequenceMove
        # Similarly to restraints, here go the __init__ arguments of the MCMCMove.
        move_list:
            - type: MCDisplacementMove
              # __init__ arguments of the nested MCMCMove.
              displacement_sigma: 1.5*nanometer
            - type: MCRotationMove
            - type: LangevinSplittingDynamicsMove
              n_steps: 500
              splitting: VRORV
    single-move:
        type: LangevinSplittingDynamicsMove
        timestep: 1.0*femtosecond

samplers:
    independent:
        type: MultiStateSampler
        # Here we can specify all the __init__ options.
        mcmc_moves: my-sequence-move
        number_of_iterations: 2000
        locality: 3
    repex:
        type: ReplicaExchangeSampler
        mcmc_moves: [single-move, single-move, my-sequence-move]
        number_of_iterations: 1000
        replica_mixing_scheme: swap-all
    sams:
        type: SAMSSampler
        state_update_scheme: local-jump

experiments:
    sampler: !Combinatorial [independent, repex]
    system: MySystem
    protocol: MyProtocol
    restraint: FlatBottom
```
I'd add the `samplers` and the `mcmc_moves` in two separate PRs (starting from the samplers).

I plan to set the defaults to use `ReplicaExchangeSampler` and 5000 iterations. Does this sound ok? Or do we want this to be a mandatory option?

Lnaden:
I like it, but I think there was specifically an issue that if we *dont* set the mcmc_moves option, it defaults to a stock set

andrrizzi:
I think the issue that came up was about both `timestep` and `number_of_iterations`. The second one is probably the single most changed option in the YAML script. This is why I wanted to have your opinion about default value vs. mandatory.

IMHO the `timestep` is a little less critical for now. We may get by using the default in most cases. So far, the only instances I had to change a default was while running tests to debug NaN issues.

Lnaden:
I think there should be global options like `timestep` and `number_of_iterations` which should be able to overwrite in the global options.

andrrizzi:
[...]

PR #899, opened 2018-03-01
Add samplers hooks to YAML scripts

andrrizzi:
This adds the sampler hooks into the YAML syntax as discussed in #898 . I've also added the `default_number_of_iterations` global option.

@Lnaden, I still have to update the `whatsnew` and the YAML docs, but I don't think I'll have time to do it before Tue (I need to do another couple of things before the end of the day) so feel free to merge/modify this if you need it before next week.

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -1997,7 +1997,7 @@ class ExperimentBuilder(object):
         restraint:
             required: no
             type: dict
-            validator: ensure_type_is_key
+            validator: is_restraint_constructor
             keyschema:
                 type: string
         """
@@ -2008,8 +2008,6 @@ class ExperimentBuilder(object):
         experiment_schema['protocol']['allowed'] = [str(key) for key in self._protocols.keys()]
         # Options validator
         experiment_schema['options']['coerce'] = coerce_and_validate_options_here_against_existing
-        # Restraint requirements
-        experiment_schema['restraint']['validator'] = ensure_restraint_type_is_key
 
         experiment_validator = schema.YANKCerberusValidator(experiment_schema)
         # Schema validation
@@ -2641,6 +2639,19 @@ class ExperimentBuilder(object):
         if not os.path.isdir(directory):
             os.makedirs(directory)
 
+    def _create_experiment_restraint(self, experiment_description):
+        """Create a restraint object for the experiment."""
+        # Determine restraint description (None if not specified).
+        restraint_description = experiment_description.get('restraint', None)
+        if restraint_description is not None:
+            restraint_parameters = copy.deepcopy(restraint_description)
+            restraint_type = restraint_parameters.pop('type')
+            # Convert quantity parameters.
+            restraint_parameters = {par: convert_if_quantity(value)
+                                    for par, value in restraint_parameters.items()}
+            return restraints.create_restraint(restraint_type, **restraint_parameters)
+        return None
+
     def _build_experiment(self, experiment_path, experiment):
         """Prepare a single experiment.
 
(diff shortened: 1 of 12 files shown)


jchodera:
Awesome! Thanks!

Yank/schema/validator.py
```diff
@@ -102,10 +113,135 @@ def _validate_type_quantity(self, value):
...
+
 
-# ====================================================
-# UTILITY FUNCTIONS
-# ====================================================
+# ==============================================================================
+# OBJECT CONSTRUCTORS PARSING
+# ==============================================================================
+
+def call_constructor(parent_cls, constructor_description, special_conversions=None,
```
Lnaden: Ooohhhhh... I like this.... I'll need to spend more time wrapping my head exactly what it does, but I very much like this

Yank/experiment.py
```diff
@@ -1011,12 +990,11 @@ def _filter_options(reference_options):
 
         experiment_options = _filter_options(self.EXPERIMENT_DEFAULT_OPTIONS)
         phase_options = _filter_options(AlchemicalPhaseFactory.DEFAULT_OPTIONS)
-        sampler_options = _filter_options(multistate.ReplicaExchangeSampler.default_options())
         alchemical_region_options = _filter_options(mmtools.alchemy._ALCHEMICAL_REGION_ARGS)
         alchemical_factory_options = _filter_options(utils.get_keyword_args(
             mmtools.alchemy.AbsoluteAlchemicalFactory.__init__))
 
-        return (experiment_options, phase_options, sampler_options,
+        return (experiment_options, phase_options,
```
Lnaden: Note: Docstring on line 971 needs changed to make IDE happy

Yank/experiment.py
```diff
@@ -2662,8 +2734,7 @@ def _build_experiment(self, experiment_path, experiment):
 
         # Get and validate experiment sub-options and divide them by class.
         exp_opts = self._determine_experiment_options(experiment)
-        (exp_opts, phase_opts, sampler_opts,
-         alchemical_region_opts, alchemical_factory_opts) = exp_opts
+        (exp_opts, phase_opts, alchemical_region_opts, alchemical_factory_opts) = exp_opts
```
Lnaden: Since the sampler have been stripped out of here, do the `!Combinatorial` experiments still work, I'm assuming so, just checking.

Yank/experiment.py
```diff
@@ -2641,6 +2640,79 @@ def _safe_makedirs(self, directory):
         if not os.path.isdir(directory):
```
Lnaden: May as well clean up this todo while we are here . Note to self -LNN

Lnaden:
So I think these tests pass, the problem is that `test_charged_ligand` is painfully slow, like 530-800 seconds long

Yank/schema/validator.py
```diff
@@ -1,12 +1,17 @@
 import os
+import copy
+import inspect
 import logging
 import cerberus
 import simtk.unit as unit
 
 from datetime import date
 from openmmtools.utils import typename
+from openmoltools.utils import unwrap_py2
```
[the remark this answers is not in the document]
andrrizzi: Yeah, the problem is `openmoltools` is still Py2-compatible. Until we drop support, I think we'll need this unwrapper for functions that come from that library.

Yank/experiment.py
andrrizzi: I think they should work as usual! I haven't added the possibility to have combinatorial samplers yet, but I thought we can add that later in a second pass.

PR #904, opened 2018-03-06
[WIP] Implement RMSD restraints

jchodera:
This is a draft implementation of RMSD restraints to hold the ligand in a defined pose with respect to the receptor.

Until more features are added to OpenMM's `RMSDForce` to allow one atomset for alignment and another atomset for RMSD evaluation, all atoms selected for the receptor and protein are used together for both alignment and RMSD evaluation.

The intention is that if the ligand or protein atoms aren't already specified by the user, all ligand heavy atoms and all receptor alpha carbons (MDTraj DSL `name CA`) are used.

The energy penalty is
```
lambda_restraints * step(RMSD - RMSD0) * (K_RMSD / 2) * (RMSD - RMSD0)^2
```
where the user specifies `RMSD0` and `K_RMSD` if they wish. Sensible defaults are provided.

The intention is for the restraint to be left on throughout the complex phase to define the integration volume for that phase. However, we need to do two more things to make this work correctly:
1. Also use the RMSD restraints for the solvent phase, applied to just the ligand (with the same ligand reference coordinates), where the restraints are turned *on* as the ligand is decoupled
2. Compute the standard state correction by the ratio of the ligand simulation volume to the standard state volume. Anything else is going to be very expensive and not as effective.

For some reason, the tests are failing, but I'm not sure why. 

@andrrizzi @lnaden: If you can take a look, that would be awesome!

I'm not actually sure how to solve issues (1) or (2) with the current code structure, which has grown a little unwieldy (especially the `Boresch` restraint class) for me to be able to easily comprehend how to make those changes.


Lnaden:
Okay. So this is ready to be looked at, but not ready to be used scientifically.

RMSD restraint is in, I have changed the base to merge into `master` instead of `sams`. I finally have worked out all the kinks and tests pass locally. However, there are some re-engineering issues I want to go over with @andrrizzi to make sure we get this right.

* What is the correct "Standard State correction" for the RMSD restraint
* How to we apply this restraint to the ligand in the solvent/vacuum phase, right now restraints can only be applied in the Complex phase. 

For the later, we rely on the `yank` module to determine how and when restraints are applied. I think we would need each restraint class to define what phases its valid in, which may take some more engineering, but that seems like the least effort while still keeping the code clean.

Yank/restraints.py
```diff
@@ -475,6 +552,29 @@ def _add_force_in_separate_group(cls, system, restraint_force):
         restraint_force.setForceGroup(min(available_force_groups))
         system.addForce(restraint_force)
 
+    @property
+    def _excluded_init_parameters(self):
```
andrrizzi: I wonder if we should create something like an `OrientationalRestraint` base class, and push `_excluded_init_parameters` and `_parameters` into that. The reason being that these two attributes make assumptions about a constructor that `ReceptorLigandRestraint` doesn't have, and, for a general restraint, we want to keep the constructor as free as possible.

If you would like to avoid a base class for `Boresch` and `RMSD`, we could instead have a mix-in class that exposes these two properties so that we can keep the very base restraint class as general as possible.

Yank/restraints.py
```diff
@@ -1992,6 +2048,285 @@ def _assign_if_undefined(attr_name, attr_value):
...
+    ...                  K_RMSD=1.0*unit.kilocalories_per_mole/unit.angstrom**2,
+    ...                  RMSD0=1*unit.angstroms)
+
+    Get standard state correction.
+
+    >>> correction = restraint.get_standard_state_correction(thermodynamic_state)
+
+    """
+    def __init__(self, restrained_receptor_atoms=None, restrained_ligand_atoms=None,
+                 K_RMSD=None, RMSD0=None):
```
andrrizzi: Can you maybe add a `reference_positions=None` argument here? It makes sense if people use the API to be able to instantiate a restraint without instantiating a `SamplerState` and a `Topography` object first.

For curiosity, do you know why OpenMM requires passing the positions of all the particles to `RMSDForce` instead of only the restrained ones?

Lnaden: I thought about making it public, but wanted to ask about it. I can make it public

It needs a full reference coordinate set to know how the atoms are placed in the world relative to the system itself. Its much much easier to engineer a solution where you pass the same number of particles to the system as every other force and uses the same atomic indices (e.g. atom 512 is Force A is the same as 512 in Force B)

docs/algorithms.rst
```diff
@@ -363,6 +363,24 @@ Note that the analytical standard state correction described in Eq. 32 of :cite:
...
+
+If the RMSD restraints are used, the ligand and protein atoms are restrained via a flat-bottom harmonic restraint applied to the
+least-squares RMSD: ``lambda_restraints * step(RMSD - RMSD0) * (K_RMSD/2) * (RMSD - RMSD0)^2``
+
+Standard use of RMSD restraints to define the binding mode to evaluate is to leave these restraints on throughout the entire
+alchemical protocol.
+
+If the receptor and ligand atom sets are not defined, receptor CA atoms (``name CA`` selection for MDTraj) and ligand heavy atoms are used.
+
+.. warning:: The RMSD restraints must also be applied in the solvent phase of the calculation, but this is not implemented yet.
```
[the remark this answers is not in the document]
jchodera: We should probably run some tests to identify an optimal way to use this restraint before updating our guidance in the flowchart. More generally, we can add experimental options and denote them as such in the docs.

Yank/restraints.py
```diff
@@ -2117,18 +2123,23 @@ class RMSD(ReceptorLigandRestraint):
...
     def __init__(self, restrained_receptor_atoms=None, restrained_ligand_atoms=None,
-                 K_RMSD=None, RMSD0=None):
+                 K_RMSD=None, RMSD0=None,
+                 reference_state=None):
         self.restrained_receptor_atoms = restrained_receptor_atoms
         self.restrained_ligand_atoms = restrained_ligand_atoms
         self.K_RMSD = K_RMSD
         self.RMSD0 = RMSD0
-        self._reference_sampler_state = None
+        self.reference_state = reference_state
```
andrrizzi: Thanks!

I think `reference_state` is a little ambiguous though. I've had somebody trying to use the API complaining in an email that handling all these "states" can be confusing at times. I would either call this `reference_sampler_state` or just pass the positions and call it `reference_positions`.

Yank/restraints.py
```diff
@@ -1992,6 +2048,297 @@ def _assign_if_undefined(attr_name, attr_value):
...
+        _assign_if_undefined('RMSD0', 2.0 * unit.angstroms)
+
+        # Write restraint parameters
+        msg = 'restraint parameters:\n'
+        for parameter_name, parameter_value in self._parameters.items():
+            msg += '{0:24s} : {1}\n'.format(parameter_name, parameter_value)
+        logger.debug(msg)
+
+        # Assign the sampler state
+        _assign_if_undefined('reference_state', sampler_state)
```
Lnaden: @andrrizzi I'd like your opinion here.

This block reassigns the `reference_state` if it has not been defined. However, that assumes you do not want to overwrite the `reference_state` from `__init__`. Should that be the default behavior, or should `determine_missing_parameters` overwrite `__init__` parameters? This is the only time it has happened where the parameter has a 1-to-1 match in `__init__` and `determine_missing_parameters`.

My thought is that `__init__` is priority and `determine_missing_parameters` should only fill in what was not already provided.

Yank/restraints.py
Lnaden: I thought about allowing either a `SamplerState` or numpy array, but then I figured it would be better to minimize the number of objects users are allowed to and should pass around. Enforcing sampler state ensures a certain structure.

I'll change the name

Yank/restraints.py
andrrizzi: > My thought is that __init__ is priority and determine_missing_parameters should only fill in what was not already provided.

I agree!

jchodera:
On the question of standard state corrections, we have a few options:
* Ensure the RMSD restraint is turned off at `lambda=0` in the complex phase, where the ligand is released into the entire simulation box. The log ratio of (average) box volume to standard state volume is then used as a standard state correction. This may be inefficient because the ligand can drift far from the binding site. Efficiency could be aided by using a functional form for the restraint that is flat-harmonic-linear, so that the energy does not grow too rapidly at large distances.
* The above can be made more efficient if Harmonic or FlatBottom restraints can be turned on as the RMSD restraint is turned off. That would prevent the ligand from drifting too far and provide a simpler standard state correction. This would require we have the ability to use two restraints simultaneously (and correctly select the standard state correction), however.
* Once we have the ability to align on receptor atoms and restrain the ligand atoms, we can keep the restraint on at `lambda=0` in both complex and solvent phases and the restraints will cancel except for a localization correction that keeps the ligand in a defined orientation and volume. We would need to impose a separate restraint in solution because there is no receptor to align against---only the ligand atoms are present.

jchodera:
@Lnaden : This just needs to be cleaned up a bit to merge in files from `master` and checked to make sure the average box volume is used during analysis for the standard state correction, but is otherwise good to merge!

Lnaden:
I'm not convinced we should merge this yet since the OpenMM it depends on is not released. I still don't like the idea of making users add additional flags to install. We have gotten very strong push back when we add too many flags (e.g. when we added conda forge to Omnia required install, there used to be a 3rd flag we considered and that got pretty hard rejected).

I also don't want to merge this into the bugfix release we are trying for the end of this week on for the same reasons.

A middle ground I would be willing to come to is to merge this in and change the tests to ONLY run on the dev branch for this restraint for now and provide a disclaimer to users that it requires the dev OpenMM 7.3 or whatever it will be called.

jchodera:
> A middle ground I would be willing to come to is to merge this in and change the tests to ONLY run on the dev branch for this restraint for now and provide a disclaimer to users that it requires the dev OpenMM 7.3 or whatever it will be called.

This is the best approach. YANK should not die if users have OpenMM 7.2 installed, and we should add a check that prints a helpful error message ("this feature requires OpenMM 7.3 or higher") if they try to use the RMSD restraints and OpenMM 7.3+ isn't installed.

jchodera:
But we absolutely need this feature.

Lnaden:
I added a new ability to flag things as "Developmental" with a class generating function. @andrrizzi may want to look over what I did in case an easier solution presents itself.

Lnaden:
I cannot figure out why the test is failing. I cannot reproduce locally, even in the same docker image Travis is using, and its only in the one test. Re-running the test just to see if its strongly 1-off

Yank/utils.py
```diff
@@ -1731,6 +1732,75 @@ def _sanitize_unit_name(unit_name):
...
+            check_dict = {**check_dict, **{test_string: test}}  # Create check sub-dict
+
+    class DevelopmentFeature(object):
+        DEV_ERROR = base_err if not valid else None
+        dev_reasons = check_dict
+        dev_features = feature_dict
+        dev_validate = valid
+
+        def __init__(self, *args, **kwargs):
+            if not self.dev_validate:
```
andrrizzi: I think you'll have to call `super().__init__(*args, **kwargs)` here or this will likely have problems when used as a mix-in. I think this is working right now because `ReceptorLigandRestraint` doesn't implement a constructor.

Lnaden: It will, which is why its only a "psuedo"-mixin.

We can fix it in a later PR

PR #907, opened 2018-03-07
Implement automatic determination of processes_per_experiments

andrrizzi:
By default, `processes_per_experiment` was set to `None` (i.e. no MPI communicator splitting). This causes problems with SAMS simulations as we support only 1 replica through the YAML interface right now.

This adds the `processes_per_experiment: auto` (now the default) that tries to split the MPI communicator to run as many experiments in parallel as possible given the allocated resources.

I've also started to add in the `mpi` module code to simulate an MPI environment for tests.

### DIFF
--- a/docs/yank-yaml-cookbook/all-options.yaml
+++ b/docs/yank-yaml-cookbook/all-options.yaml
@@ -50,7 +50,7 @@ options:
   minimize_tolerance: 1.0 * kilojoules_per_mole / nanometers    # Set minimization tolerance.
   number_of_equilibration_iterations: 1                         # Number of equilibration iterations.
   equilibration_timestep: 1.0 * femtosecond                     # Timestep for use in equilibration.
-  number_of_iterations: 1                                       # Number of replica-exchange iterations to simulate.
+  default_number_of_iterations: 1                               # The default number of iterations to simulate.
   nsteps_per_iteration: 500                                     # Number of timesteps per iteration.
   timestep: 2.0 * femtosecond                                   # Timestep for Langevin dyanmics.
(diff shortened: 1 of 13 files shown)


docs/yank-yaml-cookbook/all-options.yaml
```diff
@@ -50,7 +50,7 @@ options:
   minimize_tolerance: 1.0 * kilojoules_per_mole / nanometers    # Set minimization tolerance.
   number_of_equilibration_iterations: 1                         # Number of equilibration iterations.
   equilibration_timestep: 1.0 * femtosecond                     # Timestep for use in equilibration.
-  number_of_iterations: 1                                       # Number of replica-exchange iterations to simulate.
+  default_number_of_iterations: 1                               # The default number of iterations to simulate.
   nsteps_per_iteration: 500                                     # Number of timesteps per iteration.
   timestep: 2.0 * femtosecond                                   # Timestep for Langevin dyanmics.
   replica_mixing_scheme: swap-all                               # Specify how to mix replicas. Possible values are
```
jchodera: Should this go in `samplers` now?

andrrizzi: `default_number_of_iterations` is a global option that is overwritten by the specific sampler's `number_of_iterations` option. We've added it because it is changed very often, and we wanted it to avoid creating a triple-nested section just to specify this.

I haven't added here the docs for the `samplers` section. I hope I'll have time to add support for the `mcmc_moves` too before writing that new page.

andrrizzi:
There's a Py3.6 test error related to the SAMS online analysis, but I believe it's a pre-existing error, unrelated to this. I'll merge this to the `sams` branch and we can correct the test in a separate PR.

PR #908, opened 2018-03-08
Improve SAMS State Histograms and Refactor from_storage

Lnaden:
Given the large number of changes, and some things which need reviewed, I made this
into a PR instead of quick fix.

This changes how SAMS gets it `state_histograms` by storing a shape `[n_states]`
vector each iteration to the same location instead of relying on the
`reporter.replica_thermodynmaic_states` variable of shape `[iteration, n_states]`
which gets slower to access as `iteration` increases due to NetCDF file chunking.
Implemented on request of @jchodera

I would also like some feedback regarding how I [changed the online-free-energy test](https://github.com/choderalab/yank/compare/sams...histograms_and_storage?expand=1#diff-dd3d51a3d6465f978ab53b80e79929c1R1508).
Since SAMS may not sample well enough in the short run for MBAR to converge, I added
a test extension which tries to run longer up to a hard limit to see if more samples
fix the undersampling problem. We'll eventually refine that test to better
handle SAMS in the future, but I think this should work for now.

The MultistateSampler `from_storage` function has been refactored and split into
3 functions:
* Public `from_storage` function with an unchanged signature, calls the other 2
* `_instantiate_sampler_from_storage` class method which simply calls the `__init__` of the sampler after reading options from storage object
* `_restore_sampler_from_reporter` instance method which restores the state of the sampler from a reporter object.

This break down allows subclasses to implement either of the hidden methods to help with
restoring the state in a more granular way instead of having to rely on making a
full `from_storage` sub-method which does everything for the parent, but then the child
would need to re-open the reporter, reapply the settings, and it annoyed the IDE's
since they thought the `super().from_storage` returned an instance of the parent, and
not the child, making it complain about any additional calls in the `from_storage`
sub method unique to the child
Could use review @andrrizzi if possible (code changes shorter than reasoning)

Also fixes a bug in the online-analysis storage subroutines of the `Reporter` where
searching for alternate variable names threw a TypeError due to a convenience thing netCDF4's tries to do with key processing not handling the `in` method the way I figured it would.

Finally, I made the sampler base test class in `test_sampling.py` print out something
more helpful when it starts in on a new sampler when `--nocapture` is set. Should
help us debug a bit faster in the future.

### DIFF
--- a/Yank/multistate/multistatereporter.py
+++ b/Yank/multistate/multistatereporter.py
@@ -1044,12 +1044,8 @@ class MultiStateReporter(object):
         logZ : np.array with shape [n_states]
             Dimensionless logZ
         """
-        # TODO: Remove commented block (or whole thing) when tests pass -LNN
         data = self.read_online_analysis_data(None, "logZ")
         return data['logZ']
-        # online_group = self._storage_analysis.groups['online_analysis']
-        # logZ = online_group.variables['logZ'][iteration]
-        # return logZ
 
     def write_logZ(self, iteration: int, logZ: np.ndarray):
         """
@@ -1064,17 +1060,6 @@ class MultiStateReporter(object):
             Dimensionless log Z
         """
         self.write_online_data_dynamic_and_static(iteration, logZ=logZ)
-        # TODO: Remove commended block if tests work (Could also remove function entirely)-LNN
-        # analysis_nc = self._storage_analysis
-        # if 'logZ' not in analysis_nc.dimensions:
-        #     analysis_nc.createDimension('logZ_length', len(logZ))
-        #     if 'online_analysis' not in analysis_nc.groups:
-        #         online_group = analysis_nc.createGroup('online_analysis')
-        #         # larger chunks, faster operations, small matrix anyways
-        #         online_group.createVariable('logZ', float, dimensions=('iteration', 'logZ_length'),
-        #                                     zlib=True, chunksizes=(1, len(logZ)), fill_value=MISSING_VALUE)
-        # online_group = analysis_nc.groups['online_analysis']
-        # online_group.variables['logZ'][iteration] = logZ
 
     # TODO: Remove function if tests pass -LNN
     def read_mbar_free_energies(self, iteration):
@@ -1285,9 +1270,9 @@ class MultiStateReporter(object):
     def _find_alternate_variable(iteration, variable, storage):
         """Helper function to figure out what went wrong when data not found"""
         iter_var = variable + "_history"
-        if iteration is None and iter_var in storage:
+        if iteration is None and iter_var in storage.variables:
             return True
-        elif iteration is not None and variable in storage:
+        elif iteration is not None and variable in storage.variables:
             return True
         return False
 
@@ -1311,7 +1296,6 @@ class MultiStateReporter(object):
         else:
             return data
 
-
     def _determine_netcdf_variable_parameters(self, iteration, data, storage):
         """
         Pre-determine the variable information needed to create the variable on the storage layer
@@ -1321,16 +1305,16 @@ class MultiStateReporter(object):
             # Scalar data
             size = 1
             try:
-                dtype = data.dtype # numpy
+                dtype = data.dtype  # numpy
             except AttributeError:
-                dtype = type(data) # python
+                dtype = type(data)  # python
         else:
             # Array data
             size = len(data)
             try:
-                dtype = data.dtype # numpy
+                dtype = data.dtype  # numpy
             except AttributeError:
-                dtype = type(data[0]) # python
+                dtype = type(data[0])  # python
 
         data_dim = "dim_size{}".format(size)
 
--- a/Yank/multistate/multistatesampler.py
+++ b/Yank/multistate/multistatesampler.py
@@ -1130,6 +1130,22 @@ class MultiStateSampler(object):
         termination is delayed so that the file is not written only with
         partial data if the program gets interrupted.
 
+        Subclasses should not attempt to modify this function as it can
+        force either duplicated or missed ``sync()`` calls
+        """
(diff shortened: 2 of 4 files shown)


andrrizzi:
There may be a problem with MPI in `from_storage()`. Overall this strategy should work, but I have few design suggestions based on my personal opinion. In any case, if you manage to make `from_storage()` work with MPI this is good to be merged for me!

**from_storage**
I don't think the current way of splitting `from_storage()` will work with MPI. By the time the super implementation `MultiStateSampler._restore_sampler_from_reporter` is done, only the root MPI process has an opened reporter, so I think `self.read_logZ` will fail.

As far as I understand, the only piece of `from_storage()` that will be different among different classes is this (https://github.com/choderalab/yank/blob/sams/Yank/multistate/multistatesampler.py#L228-L240) because different samplers may need to read read different attributes from their reporters. We could leave `from_storage` largely untouched but have an extra function `MultiStateSampler._read_attributes_from_reporter(reporter) -> dict`, and have in `from_storage()` something like
```python
def from_storage(cls, storage):
    ...
    attributes = cls._read_attributes_from_reporter(reporter)
    reporter.close()
    for attr, value in attributes.items():
        setattr(sampler, attr, value)
    ...
```
so that subclasses will only have that function to override. The `SAMS.initialize/update_stage` could just go after calling `super().from_storage()` since, as far as I understand, they don't need the reporter.

**histogram**
For the histogram problem (i.e. the property `_state_histogram` becomes slower with the number of iterations), I'm not sure that storing the histogram is the best solution. The bottleneck is almost certainly the I/O operation rather than the computation of the histogram itself, which I would expect to be very fast even for very large number of iterations. Instead of storing redundant information on disk, we could compute the histogram only once at the beginning, and update it internally as we go.

Lnaden:
> The bottleneck is almost certainly the I/O operation rather than the computation of the histogram itself

Sure, the bottleneck is IO. That was the original problem since the first implementation called `read_replica_thermodynamic_states` for the whole simulation, forcing the search to go through each of the very disjointed and very small `chunksize=1` that is iteration which takes longer and longer to search through as the simulation goes on. I just refined the calculation way a little bit to be more clear while i was at it. I also don't think its a good idea to just store it internally because if we crash, we lose that information and its hard to restore. I suppose as part of the `from_storage` call we can use the old mechanism and store that information locally then.

andrrizzi:
> if we crash, we lose that information and its hard to restore.

I don't think it'll be too hard. We could even avoid to hardcode it in `from_storage` if we keep using the property. I had in mind something like
```python
@property
def _state_histogram():
    if not hasattr(self, '_cached_histogram'):
        # read states from disk and compute histogram
    return self._cached_histogram

def _update_histogram(thermodynamic_states):
    # update the internally cached histogram here
    # this is called in the update after each iteration
```

Lnaden:
> ...only the root MPI process has an opened reporter, so I think self.read_logZ will fail.

Okay I see what you mean there. I could do a bit more refactoring to something similar to how I have `_pre_write_create` in reverse where all needing the reporter we are restoring from are fetched and stored prior to closing the reference reporter and opening the head node reporter. Just further sub-divide the work. It may get a bit confusing which is why it would be a proper (small) refactor to avoid too much dividing

andrrizzi:
> Just further sub-divide the work. It may get a bit confusing which is why it would be a proper (small) refactor to avoid too much dividing

Hmm, maybe I'm misunderstanding what you're proposing, but why not adopting something like `_read_attributes_from_reporter()` that I proposed above? That would avoid an extra function. We wouldn't need the `_instantiate_sampler_from_storage` or the `_restore_sampler_from_reporter` in that case.

Lnaden:
> but why not adopting something like _read_attributes_from_reporter()

Its effectively the same scheme, just moves the work somewhere else.

> We wouldn't need the  _instantiate_sampler_from_storage or the _restore_sampler_from_reporter in that case. 

I would actually like to have both methods as they are now. From my perspective its a bit of future-proofing by having maximum flexibility, and to make sure we don't have any class oddities later. 

The `_instantiate_sampler_from_storage` I want to keep mainly so the IDE's stop complaining that we are returning the wrong class, which with a `classmethod` factory, we kind of were, just because of python's idea of who the `cls`/`self` is. e.g. the old way where everything was in `from_storage`, calling `super(SAMSSampler, cls).from_storage` technically returns only something which looks an awful lot like just a `MultiStateSampler`, which then we had to do ugly things to crack open and set all the SAMS things, and the IDE's still complained about it. 

As for `_read_attributes_from_reporter()`, we can just move around some of the other calls to make `_instantiate_sampler_from_storage` behave like that. Specifically, we move the 
```        
# Close reading reporter.
        reporter.close()
```
and 
```
        # We open the reporter only in node 0.
        self._reporter = reporter
        mpi.run_single_node(0, self._reporter.open, mode='a',
                            broadcast_result=False, sync_nodes=False)
```
from the `_instantiate_sampler_from_storage` to the primary `from_storage` call as the last actions since it will be the last thing any setup scheme will need. That way we continue to protect attribute setting, subclasses don't have to worry about all the `open` and `close` calls (handled purely by super class), and subclasses can just do whatever they need to sub-method `_instantiate_sampler_from_storage` with the assumption of an open reporter. Same number of functions, no extra engineering involved, just move exiting calls around.

Lnaden:
This look good now, @andrrizzi, @jchodera?

jchodera:
I think most of these are positive changes, though this is a net *increase* in lines of code, but I'm confused about the histogram behavior: From what I can tell, the `_state_histogram` property generates a state histogram from the storage the first time it is called, but I don't see that it is never actually updated to be current. 

Am I missing something?

jchodera:
@Lnaden pointed out it is updated here: https://github.com/choderalab/yank/pull/908/files#diff-a3a2806a37b6877068dfc229ba0c1273R352

PR #910, opened 2018-03-09
Fix reference before assignment bug on resuming

andrrizzi:


jchodera:
Looks like we may still have to fix up analysis a bit, but merging this now!
Thanks @andrrizzi !

PR #913, opened 2018-03-12
Remove dependencies from setup.py - cut down meta.yaml build

### LINKED ISSUE #876: Remove the `install_requires` from the `setup.py` file
Since YANK cannot be installed purely by `EasyInstall` or `pip` because of dependencies, I propose removing the `install_requires` keyword, or at least comment it out for documentation purposes. This would let us reduce the number of items on the `meta.yaml` `requirements: build` section and speed up build times.

Lnaden:
This speeds up conda build times by reducing the number of packages which have to be downloaded.

Fixes #876

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -14,22 +14,7 @@ requirements:
   build:
     - python
     - cython
-    - numpy
-    - scipy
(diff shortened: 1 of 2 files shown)


Lnaden:
For documentation, this saves about 2.5 to 3 min per build.

PR #915, opened 2018-03-15
Unbias restraint analysis

### LINKED ISSUE #882: Unbiasing of the harmonic restraint
This is my plan for implementing the harmonic restraint unbiasing in the analysis.

1. Parse solute-only trajectory and recompute the centroid distances and restraint energies.
2. After decorrelating, discard from the analysis all the frames with restraint energy greater than a threshold. The function will have an argument defining it, but **what should the default value be?** 2kT? 3kT?
3. Recompute the standard state correction by integrating a different potential that is the same as the restraint until it gets to the threshold, at which point it goes to infinity.
4. If the simulation used the anisotropic long-range dispersions correction, subtract the restraint energies to that state; otherwise, create a new state for MBAR that has the energy of the lambda=1 state minus the restraint energy.

I'd implement this in an external script for now (starting from [this one](https://github.com/choderalab/yank-validation/blob/temp/debug_scripts/debug_hostguest.py), which is almost ready) since there are so many changes going on in the analysis, and I'd like to avoid re-doing the same work twice. Once the new analysis API will be merged into `master`, I'll bring it into the analysis.

andrrizzi:
Fix #882.

Not ready yet. I still have to add a bunch of tests (and fix the bugs), modify the CLI arguments, and do some cleanup, but this is a first draft of the implementation for the unbiasing of Harmonic/Flatbottom restraints. It also adds a system to invalidate automatically cached values once the attributes they depend on are changed. This allows, for example, to quickly compute free energy profile as a function of the restraint cutoff or the number of iterations, which could be performed only very very slowly before these changes.

Tests won't pass until choderalab/openmmtools#336 is merged, and a new openmmtools release is cut.

### DIFF
--- a/Yank/commands/analyze.py
+++ b/Yank/commands/analyze.py
@@ -97,15 +97,15 @@ def dispatch(args):
 def extract_analyzer_kwargs(args, quantities_as_strings=False):
     """Return a dictionary with the keyword arguments to pass to the analyzer."""
     analyzer_kwargs = {}
-    if args['skipunbiasing']:
+    if args['--skipunbiasing']:
         analyzer_kwargs['unbias_restraint'] = False
-    if args['energycutoff']:
-        analyzer_kwargs['restraint_energy_cutoff'] = float(args['energycutoff'])
-    if args['distcutoff']:
+    if args['--energycutoff']:
+        analyzer_kwargs['restraint_energy_cutoff'] = float(args['--energycutoff'])
+    if args['--distcutoff']:
         if quantities_as_strings:
-            distcutoff = args['distcutoff'] + '*angstroms'
+            distcutoff = args['--distcutoff'] + '*angstroms'
         else:
-            distcutoff = float(args['distcutoff']) * unit.angstroms
+            distcutoff = float(args['--distcutoff']) * unit.angstroms
         analyzer_kwargs['restraint_distance_cutoff'] = distcutoff
     return analyzer_kwargs
 
--- a/Yank/commands/analyze.py
+++ b/Yank/commands/analyze.py
@@ -43,9 +43,18 @@ Free Energy Required Arguments:
 
 Free Energy Optional Arguments:
   --skipunbiasing               Skip the radially-symmetric restraint unbiasing. This can be an expensive step.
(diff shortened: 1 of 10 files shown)


andrrizzi:
- I've added three new argument to the analyzer (and respective CLI arguments) `unbias_restraint`, `restraint_energy_cutoff`, and `restraint_distance_cutoff`.
- By default the distance cutoff is computed automatically as the 99.9-percentile of the distance distribution at the lambda=1 state.
- I've added a new plot in the notebook that shows the distributions of the restraint energies/distances and the cutoff so it should be easy to check if the cutoff makes sense.

I haven't tested the notebook figure yet. I'll have to generate some data first, but I think this pretty much ready.

andrrizzi:
This should be ready for review now. Not sure why tests haven't started yet.

Yank/commands/analyze.py
```diff
@@ -37,6 +41,12 @@
 Free Energy Required Arguments:
   -s=STORE, --store=STORE       Storage directory for NetCDF data files.
 
+Free Energy Optional Arguments:
+  --skipunbiasing               Skip the radially-symmetric restraint unbiasing. This can be an expensive step.
+  --distcutoff=DISTANCE         The restraint distance cutoff (in angstroms) to be used to unbias the restraint.
```
jchodera: What's the default behavior if `--skipunbiasing` isn't specified? Is it a distance- or energy-based unbiasing cutoff?

andrrizzi: Ah! Right, I still have to update the CLI docs. See above:

> By default the distance cutoff is computed automatically as the 99.9-percentile of the distance distribution at the lambda=1 state.

This is only if the user doesn't provide any distance/energy cutoff.

jchodera: Awesome, thanks!

andrrizzi:
Thanks! I'm testing the notebook on some data right now before merging. I'll probably have to fix a couple of things.

> There's significant new additional complexity added, so it will be great to sync up in more detail about this at some point int he near future.

Of course, I can give you an overview of the new things I introduced (mostly the `CachedProperty` descriptor), and of the reasons why I had to introduce the complexity.

PR #916, opened 2018-03-20
Add YAML syntax for MCMC moves

### LINKED ISSUE #898: Allow to specify the MultiStateSampler and MCMCMoves in the YAML script

andrrizzi:
Fix #898 .

For the existing global options that were injected into the MCMC move, I've kept `timestep` and `nsteps_per_iterations` (now `default_timestep` and `default_nsteps_per_iterations` respectively) because I thought these are options we'll have to change often, and I've deleted `collision_rate`, `mc_sigma_displacement` and `integrator_splitting` that I don't remember I ever had to change for production simulations. They can still be set in the `mcmc_moves` block, of course, but let me know if you think I should do things differently.

This requires choderalab/openmmtools#336 (and a new `openmmtools` release) for tests to pass.


jchodera:
Awesome! Can we specify `timestep` and `nsteps_per_iteration` in the MCMC block instead? I can see cases where we might use different blocks with different timesteps, so I'd prefer these being able to be specified in those MCMC blocks.

andrrizzi:
It can be specified in both. The global options `default_timestep` and `default_nsteps_per_iterations` are used only if the default `MCMCMove` is used (i.e. if no `mcmc_move` is given in the `samplers` section).

Thanks for the reviews! I'll wait for tests to pass before merging.

jchodera:
> The global options default_timestep and default_nsteps_per_iterations are used only if the default MCMCMove is used (i.e. if no mcmc_move is given in the samplers section)

Brilliant! Thanks!

andrrizzi:
Uh! I just realized this wasn't merged.

Lnaden:
Seems like an easy enough fix though, most all my changes were doc-related and I can make another pass though if something breaks.

PR #919, opened 2018-03-22
Fix all doc warnings

Lnaden:
Fix all the outstanding doc warnings, broken links, and DeprecationWarning, *all of them*

Fixed missing citation links
Docs now build on Sphinx 1.[567].X (more of a pain than I would like, cc @jchodera)

I don't think we actually use anything in the shipped `notebook_sphinxext`
file and is a leftover things from when we initially copied the docs
from MDTraj, so we might want to remove this in the future.

To fix the citations, I had to comment out many unused citations to get the warnings to stop, we can uncomment them if we actually cite them.


Lnaden:
Zero functional code changes, so I am going to merge this in

PR #925, opened 2018-03-23
Update automatic equilibration detection to handle SAMS

jchodera:
This PR makes changes to the multistate sampler analysis code to allow automatic equilibration detection to work with SAMS and other methods that adjust log weights dynamically.

@lnaden: I had to roll back a bit of the generalization of the `u_n` negative log deviance timeseries fed to automated equilibration detection because I couldn't find a good way to preserve this in generality. We can figure out a better way to generalize that component later. The rest of the infrastructure worked well, though!

I also had to implement a few hacks that access some online analysis data directly, since I couldn't figure out how to (1) check if the online analysis data existed, and (2) retrieve particular parts of it through the existing `Reporter` API. This could undoubtedly use refinement.

This hasn't been extensively tested beyond the unit tests. I'm brainstorming fast but useful tests, as well as testing on `lilac` right now.

PR #926, opened 2018-03-24
FIRE minimizer and MPI support for Cray XK nodes

jchodera:
On `lilac`, minimization of 50K atom systems (like c-Met) often took ~30 minutes/minimization, with several stages of minimization required (fully-interacting system, alchemically-modified system) to start a run. 

This PR implements support for a tuned version of the FIRE minimizer as a `CustomIntegrator`, now enabled by default:

Erik Bitzek, Pekka Koskinen, Franz Gaehler, Michael Moseler, and Peter Gumbsch.
Structural Relaxation Made Simple. PRL 97:170201, 2006. [DOI](http://dx.doi.org/10.1103/PhysRevLett.97.170201)

For c-Met, this brings minimization times down from 30 minutes to 30 seconds.

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -320,9 +320,10 @@ class AlchemicalPhaseFactory(object):
             tolerance = self.options['minimize_tolerance']
             max_iterations = self.options['minimize_max_iterations']
             scheme = self.options['minimize']
-            if scheme not in ['L-BFGS', 'FIRE']:
+            if scheme in ['yes', 'YES']:
                 scheme = 'FIRE'
-            alchemical_phase.minimize(tolerance=tolerance, max_iterations=max_iterations, scheme=scheme)
+            if scheme not in ['no', 'NO', False, None]:
+                alchemical_phase.minimize(tolerance=tolerance, max_iterations=max_iterations, scheme=scheme)
 
         # Randomize ligand if requested.
         if self.options['randomize_ligand']:
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -228,7 +228,7 @@ class AlchemicalPhaseFactory(object):
 
     DEFAULT_OPTIONS = {
         'anisotropic_dispersion_cutoff': 'auto',
-        'minimize': 'FIRE',
+        'minimize': True,
         'minimize_tolerance': 1.0 * unit.kilojoules_per_mole/unit.nanometers,
         'minimize_max_iterations': 1000,
         'randomize_ligand': False,
@@ -316,14 +316,10 @@ class AlchemicalPhaseFactory(object):
         alchemical_phase = self.create_alchemical_phase()
 
         # Minimize if requested.
-        if self.options['minimize']:                
+        if self.options['minimize']:
             tolerance = self.options['minimize_tolerance']
             max_iterations = self.options['minimize_max_iterations']
-            scheme = self.options['minimize']
-            if scheme in ['yes', 'YES']:
-                scheme = 'FIRE'
-            if scheme not in ['no', 'NO', False, None]:
-                alchemical_phase.minimize(tolerance=tolerance, max_iterations=max_iterations, scheme=scheme)
+            alchemical_phase.minimize(tolerance=tolerance, max_iterations=max_iterations)
 
         # Randomize ligand if requested.
         if self.options['randomize_ligand']:
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -228,9 +228,9 @@ class AlchemicalPhaseFactory(object):
 
     DEFAULT_OPTIONS = {
         'anisotropic_dispersion_cutoff': 'auto',
-        'minimize': True,
+        'minimize': 'FIRE',
         'minimize_tolerance': 1.0 * unit.kilojoules_per_mole/unit.nanometers,
-        'minimize_max_iterations': 0,
+        'minimize_max_iterations': 1000,
         'randomize_ligand': False,
         'randomize_ligand_sigma_multiplier': 2.0,
         'randomize_ligand_close_cutoff': 1.5 * unit.angstrom,
@@ -316,10 +316,13 @@ class AlchemicalPhaseFactory(object):
         alchemical_phase = self.create_alchemical_phase()
 
         # Minimize if requested.
-        if self.options['minimize']:
+        if self.options['minimize']:                
             tolerance = self.options['minimize_tolerance']
             max_iterations = self.options['minimize_max_iterations']
-            alchemical_phase.minimize(tolerance=tolerance, max_iterations=max_iterations)
+            scheme = self.options['minimize']
+            if scheme not in ['L-BFGS', 'FIRE']:
+                scheme = 'FIRE'
+            alchemical_phase.minimize(tolerance=tolerance, max_iterations=max_iterations, scheme=scheme)
 
         # Randomize ligand if requested.
         if self.options['randomize_ligand']:
--- a/Yank/multistate/multistatesampler.py
+++ b/Yank/multistate/multistatesampler.py
@@ -1231,7 +1231,9 @@ class MultiStateSampler(object):
 
         # Use the FIRE minimizer
         integrator = FIREMinimizationIntegrator(tolerance=tolerance)
-        context, integrator = mmtools.cache.global_context_cache.get_context(thermodynamic_state, integrator=integrator)
+
+        # Create context
+        context = thermodynamic_state.create_context(integrator)
 
         # Set initial positions and box vectors.
         sampler_state.apply_to_context(context)
@@ -1260,11 +1262,14 @@ class MultiStateSampler(object):
 
         # Get the minimized positions.
         sampler_state.update_from_context(context)
-
+        
         # Compute the final energy of the system for logging.
         final_energy = thermodynamic_state.reduced_potential(sampler_state)
         logger.debug('Replica {}/{}: final energy {:8.3f}kT'.format(
             replica_id + 1, self.n_replicas, final_energy))
+
+        # Clean up the integrator
+        del context
 
         # Return minimized positions.
         return sampler_state.positions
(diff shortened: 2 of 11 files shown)


jchodera:
This feature isn't urgent to merge. I needed if for ORNL Titan, where minimization was impeding my ability to benchmark.

I'll have to update this after we merge in #925

jchodera:
And here's an example for a 50K atom Abl:inhibitor system:
```
MultiStateSampler._initialize_reporter at 0x2b35510a7488>
2018-03-24 16:48:44,844 - DEBUG - yank.yank - Minimizing 1 sampler states in the reference thermodynamic state
2018-03-24 16:48:44,844 - DEBUG - yank.mpi - Node 1/1: execute _minimize_sampler_state(0)
2018-03-24 16:49:05,073 - DEBUG - yank.yank - Sampler state 1/1: initial energy -206567.715kT
2018-03-24 16:49:05,073 - DEBUG - yank.yank - Using FIRE: tolerance 1.0 kJ/(nm mol) max_iterations 1000
2018-03-24 16:49:09,692 - DEBUG - yank.yank - Sampler state 1/1: final energy -278742.356kT
2018-03-24 16:49:09,693 - DEBUG - yank.mpi - Node 1/1: allgather results of _minimize_sampler_state
2018-03-24 16:49:09,696 - DEBUG - yank.mpi - Node 1/1: executing <bound method MultiStateReporter.write_sampler_states of <yank.multistate.multistatereporter.MultiStateReporter object at 0x2b35bccc9be0>>
2018-03-24 16:49:09,710 - DEBUG - openmmtools.utils - Storing sampler states took    0.014s
2018-03-24 16:49:09,711 - DEBUG - yank.multistate.multistatesampler - Minimizing all replicas...
2018-03-24 16:49:09,711 - DEBUG - yank.mpi - Node 1/1: execute _minimize_replica(0)
2018-03-24 16:49:35,586 - DEBUG - yank.multistate.multistatesampler - Replica 1/1: initial energy -278665.736kT
2018-03-24 16:49:35,586 - DEBUG - yank.multistate.multistatesampler - Using FIRE: tolerance 1.0 kJ/(nm mol) max_iterations 1000
2018-03-24 16:49:40,175 - DEBUG - yank.multistate.multistatesampler - Replica 1/1: final energy -282069.935kT
2018-03-24 16:49:40,176 - DEBUG - yank.mpi - Node 1/1: sending results of _minimize_replica to 0
2018-03-24 16:49:40,180 - DEBUG - yank.mpi - Node 1/1: executing <bound method MultiStateReporter.write_sampler_states of <yank.multistate.multistatereporter.MultiStateReporter object at 0x2b35bccc9be0>>
2018-03-24 16:49:40,195 - DEBUG - openmmtools.utils - Storing sampler states took    0.015s
2018-03-24 16:49:40,195 - DEBUG - openmmtools.utils - Minimizing all replicas took   30.484s
```
Each minimization took 1-4 s, and the whole minimization phase took 30 s.

jchodera:
This should be ready for review! We probably want to merge #915 first and then update this.

Yank/experiment.py
```diff
@@ -230,7 +230,7 @@ class AlchemicalPhaseFactory(object):
         'anisotropic_dispersion_cutoff': 'auto',
         'minimize': True,
         'minimize_tolerance': 1.0 * unit.kilojoules_per_mole/unit.nanometers,
-        'minimize_max_iterations': 0,
+        'minimize_max_iterations': 1000,
```
andrrizzi: Can you update the default in the YAML pages docs for this options?

Yank/multistate/multistatesampler.py
```diff
@@ -1237,11 +1242,26 @@ def _minimize_replica(self, replica_id, tolerance, max_iterations):
...
+                logger.debug('NaN encountered in FIRE minimizer; falling back to L-BFGS after resetting positions')
+                sampler_state.apply_to_context(context)
+                openmm.LocalEnergyMinimizer.minimize(context, tolerance, max_iterations)
+            else:
+                raise e
 
         # Get the minimized positions.
         sampler_state.update_from_context(context)
 
+        # TODO: Do we have to destroy the FIRE integrator-associated context if we aren't going to use it again?
```
andrrizzi: Nope! Deleting the integrator has no practical effect if you are using the `ContextCache`. The context and the integrator will be purged automatically after a while.

Yank/yank.py
```diff
@@ -1423,6 +1438,8 @@ def _minimize_sampler_state(sampler_state_id, sampler_states, thermodynamic_stat
         logger.debug('Sampler state {}/{}: final energy {:8.3f}kT'.format(
             sampler_state_id + 1, len(sampler_states), final_energy))
 
+        # TODO: Do we have to destroy the FIRE integrator bound context if we aren't going to use it again?
```
andrrizzi: Same here.

Yank/yank.py
```diff
@@ -1413,7 +1415,20 @@ def _minimize_sampler_state(sampler_state_id, sampler_states, thermodynamic_stat
             sampler_state_id + 1, len(sampler_states), initial_energy))
 
         # Minimize energy.
-        openmm.LocalEnergyMinimizer.minimize(context, tolerance, max_iterations)
+        if (max_iterations is None) or (max_iterations == 0):
+            fire_iterations = 1000
```
andrrizzi: Same here.

Yank/multistate/multistatesampler.py
```diff
@@ -1237,11 +1242,26 @@ def _minimize_replica(self, replica_id, tolerance, max_iterations):
             replica_id + 1, self.n_replicas, initial_energy))
 
         # Minimize energy.
-        openmm.LocalEnergyMinimizer.minimize(context, tolerance, max_iterations)
+        if (max_iterations is None) or (max_iterations == 0):
+            fire_iterations = 1000
```
andrrizzi: Can you update the docstring of `MultiStateSampler.minimize()` describing this behavior? If I recall correctly, the current docs don't allow `None` as a valid value for `max_iterations`.

Yank/yank.py
jchodera: Ah! I see. Most of the functions I modified didn't have complete docstrings, and I hadn't even realized the function call hierarchy led back up to `AlchemicalPhase.minimize()`, so I had no idea what the behavior was supposed to be. Having docstrings even for private methods will make it easier in future to ensure we keep the contract the function makes with its caller.

I'll add these in and modify the behavior to match the original contract.

Yank/multistate/multistatesampler.py
jchodera: Will `ContextCache` reset the global parameters to the defaults in the requested `integrator` object before returning the context/integrator pair if it finds it already exists?

Yank/experiment.py
jchodera: Done!

jchodera:
I've updated the behavior and will do a bit more testing, @andrrizzi.

Will `ContextCache` reset the global parameters to the defaults in the requested integrator object before returning the context/integrator pair if it finds it already exists?

andrrizzi:
Thanks!

> Will `ContextCache` reset the global parameters to the defaults in the requested integrator object before returning the context/integrator pair if it finds it already exists?

Nope! If you need to make sure that some global variables are reset at each execution, you'll have to add it to the code after the `get_context()` call. The context hash should be correctly matched since you're instantiating a new integrator each time you request it, so the contexts should be recycled correctly.

jchodera:
> Nope! If you need to make sure that some global variables are reset at each execution, you'll have to add it to the code after the get_context() call. The context hash should be correctly matched since you're instantiating a new integrator each time you request it, so the contexts should be recycled correctly.

Ah. I can see how this could lead to huge issues (especially when trying to use nonequilibrium integrators, like @pgrinaway does.) 

To follow the "rule of least surprise", we should change the behavior of the context cache to always reset the integrator (and probably system) global parameters to the requested `system` and `integrator` by default, with flags allowing the program to override this behavior if they know what they're doing.

andrrizzi:
> we should change the behavior of the context cache to always reset the integrator (and probably system)

Yeah, we should probably standardize this. When I made the first implementation of `ContextCache` I didn't consider the case of integrators with a state. The system should not be a problem. That's taken care by `ThermodynamicState`.

A couple of considerations from the top of my head:
- We'll have to force our integrators to expose a `copy_state()` method. I wouldn't push into `ContextCache` the responsibility of figuring out automatically what to set.
- This behavior should probably be optional, not mandatory, as I can see cases in which you actually want to preserve the integrators variables in a sequence of `MCMCMoves`.

andrrizzi:
Btw, this behavior is explicitly stated in the documentation https://github.com/choderalab/openmmtools/blob/master/openmmtools/cache.py#L357-L359, although it could certainly benefit from a `warning` tag.

jchodera:
> As a consequence, the returned integrator is guaranteed to be in the same state as the one provided, but it can be a different instance.

Doesn't that imply that we *don't* need to modify global variables in the returned `integrator`, since the returned integrator is "guaranteed to be in the same state as the one provided"?

andrrizzi:
Oh, damn, yes. I remember now I implemented a hack to make it work some time ago, but it works only for a limited set of attributes (that can be updated at runtime): https://github.com/choderalab/openmmtools/blob/master/openmmtools/cache.py#L438-L457

jchodera:
Ok, this sounds pretty dangerous. That docstring definitely isn't right, then. 

Let's reassess later this week after the sams release. There should be a way to make this more robust.

In the meantime, how can I force ContextCache to delete the context after I'm done with it to ensure the global integrator parameters are correctly reinitialized next time?

andrrizzi:
You can just avoid using `ContextCache` if you don't have a way to reset the integrator after `get_context()`. Just use `thermodynamic_state.create_context(integrator)`.

Lnaden:
Looks stable enough to me, we can use this as an initial YANK release and do spot fixes later

PR #928, opened 2018-03-26
Add serialization to the notebooks

Lnaden:
Also renders no mater what format you specify


jchodera:
Is this something we should get into the SAMS PR before release?

Lnaden:
Yes, its a very small fix to enable something that is already in the notebooks. I just had to modify the CLI file

Lnaden:
This should very few issues with #915, but I see the few it is. Merge done

andrrizzi:
Yeah, no worries! This will be easy to resolve.

Issue #931, opened 2018-03-27
Fix randomly failing analyze tests


andrrizzi:
Just saving here as a reminder.

This assertion sometimes fails because of `nan`s: https://github.com/choderalab/yank/blob/master/Yank/tests/test_analyze.py#L277-L278 . We may be able to fix it by using `np.allclose(..., equal_nan=True)`.

Lnaden:
This should be fixed by #932

PR #932, opened 2018-03-28
Critical bugfixes for SAMS release

jchodera:
This fixes bugs in two of the features I added to the SAMS release:
* Cray detection now correctly checks environment variables instead of static list of possible MPI environment variables (whoops!)
* FIRE minimization debug message now correctly prints `max_iterations` instead of nonexistent variable

### DIFF
--- a/Yank/mpi.py
+++ b/Yank/mpi.py
@@ -150,7 +150,7 @@ def get_mpicomm():
 
     # Cray machines are usually old and sick; prevent them from causing trouble with CUDA caches
     # by explicitly using different CUDA cache paths for each process
-    if 'ALPS_APP_PE' in variables:
+    if 'ALPS_APP_PE' in os.environ:
         cuda_cache_path = os.path.abspath(os.path.join('nvcc-cache', str(mpicomm.rank)))
         if not os.path.exists(cuda_cache_path):
             os.makedirs(cuda_cache_path)
--- a/Yank/multistate/multistatesampler.py
+++ b/Yank/multistate/multistatesampler.py
@@ -1250,7 +1250,7 @@ class MultiStateSampler(object):
                 while integrator.getGlobalVariableByName('converged') < 1:
                     integrator.step(50)
             else:
-                logger.debug('Using FIRE: tolerance {} max_iterations {}'.format(tolerance, fire_iterations))
+                logger.debug('Using FIRE: tolerance {} max_iterations {}'.format(tolerance, max_iterations))
                 integrator.step(max_iterations)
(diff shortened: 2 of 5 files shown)


jchodera:
@Lnaden : We're still having trouble with the `test_analyze.py:TestMultiPhaseAnalyzer` tests failing due to `nan`s in travis tests. Maybe we can merge this and look into that separately?

Lnaden:
Thats on my todo list today. It only recently starting having that problem so I'll dig into whats changed

Lnaden:
I'm trying to replicate the tests locally but I can't. Can either of you, @andrrizzi or @jchodera ?

andrrizzi:
Yeah, I was able to reproduce locally. It seems to happen only with the SAMS test suites.

andrrizzi:
Maybe we can just try to use what I suggested in #931 (using `equal_nan=True`, it's False by default).

Lnaden:
I don't like setting `equal_nan=True` because I'm worried that it will mask errors we should actually handle.

Lnaden:
Those tests should never throw a `nan` and if we do, thats something to look into.

Lnaden:
Okay! I found the bug. Its stochastic from MBAR estimating free energy between identical, unsampled states with very low number of samples (~< 10). I have increased the number of default samples while also making the tests more robust.

I also found a bug in the additional test samples loop where the criteria to stop after X cycles didn't actually stop the cycle. Whoops.

This should fix both the `MultiStateAnalyzer` test bug and the MBAR creation test bug.

Lnaden:
Tests finally pass. We are approaching the upper limit of travis run time tests, so we may want to think about figuring a way to make some tests faster, or at least split them up.

One ideas is the `test_experiment.test_charged_ligand` which takes > 40% of the total test time, thats probably the place I would start

andrrizzi:
> One ideas is the `test_experiment.test_charged_ligand` which takes > 40% of the total test time, thats probably the place I would start

Wow! We can probably change the tested molecules to smaller molecules. As long as they have the same charge, it will still test what we want.

Lnaden:
Is it possible to do counterion addition in vacuum by force? Right now its doing it in explicit + PME, so that is probably what the slow part is

andrrizzi:
I think so! I thought the bottleneck was charging `imatinib` in the setup pipeline, but it could be adding the explicit solvent.

Lnaden:
Was there anything else needed for this pr?

jchodera:
Nope!

PR #933, opened 2018-03-28
Use receptor or solute as anchor molecule for imaging

andrrizzi:
This PR fixes a case where MDTraj raise an error when trying to guess the anchor molecule used for imaging, both in `extract_trajectory` and when computing restraint distances. It also skips molecule imaging for non-periodic systems.


jchodera:
Awesome! Will this still work if you try to use `yank analyze` on hydration free energy datasets?

andrrizzi:
It should work! We use the solute atoms if there's no receptor in `extract_trajectory()`, and there are no restraints in hydrations free energy so we should never reach the unbiasing part of the code.

PR #946, opened 2018-04-09
Fix #945: Issues when using relative import with openeye

### LINKED ISSUE #945: Issues with is_openeye_installed on ORNL Titan
[`utils.is_openeye_installed('oeomega')`](https://github.com/choderalab/yank/blob/master/Yank/utils.py#L1216) fails on ORNL Titan with miniconda Python 3.6 because [`module = importlib.import_module('openeye', tool)`](https://github.com/choderalab/yank/blob/master/Yank/utils.py#L1256) succeeds, but does not return the correct submodule `openeye.oeomega`. The alternative syntax [module = importlib.import_module('openeye.' + tool)](https://github.com/choderalab/yank/blob/master/Yank/utils.py#L1258) works, however. Can we simply use this other syntax by default?

andrrizzi:
@jchodera can you check if this fixes the problem you've been having on Titan?

Fixes #945


andrrizzi:
Whops! Sorry, there's still a small fix to this I need to add.

Lnaden:
Yes, trap that. If someone provides a bad name, it should error

jchodera:
Looks good to me---thanks!

PR #952, opened 2018-04-17
Pin cerberus to 1.1

### LINKED ISSUE #950: Schemas need updated for Cerberus 1.2
As title says, Cerberus 1.2 changed some logic involving the `ionic_strength` and `solvent_model` interplay. Looking into it.

Lnaden:
It will be a much larger pain to convert to 1.2 as there are a few things
which changed. Until we move to the public schema, I don't think its worth
the time sink to fix this proper.

Fixes #950


andrrizzi:
Sounds good to me!

PR #953, opened 2018-04-18
Added API UML class diagram

andrrizzi:
I've added to the online docs the UML class diagram I made for the proposal on the index page of the API.


jchodera:
Looks great!

Lnaden:
Is there a way we can auto generate the png/xml?

andrrizzi:
I've been playing with auto-generators when I was programming with C++ and Java, but I never found one that was able to generate a pretty, readable diagram. It may be harder for a language like Python since it's duck-typed, but maybe there are now tools that can read type annotations out there now.

Lnaden:
Ah, okay, I did not realize how manual it is. I did not see the XML file in the Draw.io diagram folder. never mind then. LGTM, although it will be outdated quickly with the RMSD restraint

andrrizzi:
Ok, merging then, thanks!

> although it will be outdated quickly

Yeah, I know, sorry. This is pretty much why I added the disclaimer "This is not meant to be a comprehensive and accurate representation of the API". Better than not having it though!

PR #955, opened 2018-04-18
Fix #954: MultiStateSampler.run(None) doesn't work when resuming.

### LINKED ISSUE #954: Sampler's run method doesn't work correctly when the n_iteration argument is not specified
If `n_iterations` is `None`, the current algorithm compute incorrectly the maximum iteration, and it run for a shorter time than specified (if it runs at all). This affects the ability to resume through the API, although it doesn't affect resuming through the YAML interface since we always pass `n_iteration`.

I'll open a bugfix PR soon.

andrrizzi:


Lnaden:
Won't this fail if someone has just stopped a simulation but not collapsed it? E.g. if someone did `.run(50)` and then `.run(None)`, then it will run for the full number of iterations in addition to the 50

andrrizzi:
> then it will run for the full number of iterations in addition to the 50

Hmm. I think `sampler._is_completed(None)` just checks that `sampler.iteration >= sampler.number_of_iterations` so it should behave as expected and stop at `number_of_iterations`.

Lnaden:
Okay, then that will work. I think there is some duplicated checking going on there, but its fine as is

andrrizzi:
Ok, thanks for the review!

PR #970, opened 2018-05-01
Fix bug in detection logic determining stage

jchodera:
This fixes a bug in `_update_stage()` where only the iteration where histograms are flat is set to `asymptotically-optimal`.

### DIFF
--- a/Yank/multistate/sams.py
+++ b/Yank/multistate/sams.py
@@ -559,7 +559,7 @@ class SAMSSampler(MultiStateSampler):
             else:
                 raise ValueError("Unknown flatness_criteria %s" % flatness_criteria)
 
-            if advance:
+            if advance or ((self._t0 > 0) and (self._iteration > self._t0)):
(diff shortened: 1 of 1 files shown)


Yank/multistate/sams.py
```diff
@@ -559,7 +559,7 @@ def _update_stage(self):
             else:
                 raise ValueError("Unknown flatness_criteria %s" % flatness_criteria)
 
-            if advance:
+            if advance or ((self._t0 > 0) and (self._iteration > self._t0)):
```
andrrizzi: Just to understand this better, in what situation we can have `self._t0 > 0` and `self._iteration <= self._t0` at the same time?

jchodera: Good point, but once we start adding the ability to back up by the checkpoint interval to recover, I'm not sure if just checking `self._t0 > 0` is sufficient anymore

andrrizzi: Ah, I see, good point.

PR #972, opened 2018-05-01
Make restoration and last iteration more robust

Lnaden:
This is the first of a couple changes to make checkpoint restoration and access more robust.

Next step will to be make `MultiStateAnalyzer` able to detect corruption and back up X number of checkpoints. I split the PRs to minimize amount which has to be reviewed and to set the underlying structure first.

I also changed some logic on masking maximum number of available iterations to prevent unassociated data from being accessed, cut down on some lines.

@jchodera @andrrizzi any other exposed methods/functionality you think we need before I try to make the analyzer able to handle data corruption better?

@andrrizzi what would be the effect in the `MultiStateAnalyzer` if I changed the other `CachedProperty` objects to depend on `max_n_iterations`?

### DIFF
--- a/Yank/multistate/multistatesampler.py
+++ b/Yank/multistate/multistatesampler.py
@@ -251,7 +251,7 @@ class MultiStateSampler(object):
         # Read iteration and online analysis info.
         reporter.open(mode='r')
         options = reporter.read_dict('options')
-        iteration = reporter.read_last_iteration(full_iteration=False)
+        iteration = reporter.read_last_iteration(last_checkpoint=False)
         # Search for last cached free energies only if online analysis is activated.
         if options['online_analysis_interval'] is not None:
             target_error = options['online_analysis_target_error']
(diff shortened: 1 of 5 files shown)


Yank/multistate/multistatesampler.py
```diff
@@ -886,26 +886,53 @@ def _restore_sampler_from_reporter(self, reporter):
...
+                    internal_neighborhoods, internal_energy_unsampled_states, internal_n_accepted_matrix,
+                    internal_n_proposed_matrix, internal_last_mbar_f_k, internal_last_err_free_energy)
+
+        # Keep trying to resume further and further back from the most recent checkpoint back
+        checkpoints = reporter.read_checkpoint_iterations()
+        checkpoint_reverse_iter = iter(checkpoints[::-1])
+        while True:
+            try:
+                checkpoint = next(checkpoint_reverse_iter)
+                output_data = _read_options(checkpoint)
```
andrrizzi: It may be worth testing if the netcdf file doesn't get messed up after the error. If it does, we might have to have this loop outside the whole `_restore_sampler_from_reporter()` method.

andrrizzi:
> any other exposed methods/functionality you think we need before I try to make the analyzer able to handle data corruption better?

Nothing comes to mind right now!

> what would be the effect in the `MultiStateAnalyzer` if I changed the other `CachedProperty` objects to depend on `max_n_iterations`?

Pretty much all the `CachedProperty` objects are already indirectly dependent on `max_n_iterations` indirectly. For example, `_decorrelated_state_indices_ln` depends on `equilibration_data`, which depends on `max_n_iterations`.

The `dependencies` attribute of `CachedProperty` should only list the values that it needs to explicitly access to compute it. This way, for example, if we make changes to the way we compute `equilibration_data` we won't have to change all the other properties as well, which don't care on what `equilibration_data` depends.

Yank/multistate/multistatesampler.py
Lnaden: What would we be checking for here. What is the exact metric of the netcdf file is "messed up?" And the sampler will have no idea about the netcdf file in theory, so we would have to make a `Reporter` method which does like a integrity check either as a whole or at an iteration.

andrrizzi: Sorry, by testing I didn't mean writing an actual test. I'm aware this is very hard to reproduce. I meant we should be on the lookout since after the first HDF5 error (or similar), the netcdf dataset object could be left in an inconsistent state and we may have to close the dataset and open a new one instead of reading previous iterations from the same dataset object.

> And the sampler will have no idea about the netcdf file in theory

I agree. I was only proposing as a possible solution to the problem above (and only in the case the problem was real) to move the `while True` loop from `_restore_sampler_from_reporter()` to `from_storage()` to close/open the dataset at every loop iteration.

Nothing that needs to be done if we don't incur any problems with this implementation though!

PR #978, opened 2018-05-03
Change Boresch Restraints energy function and auto-atom method

Lnaden:
The Boresch restraints now pick their atoms based on bonded atoms only
to reduce the chance the two outer torsions can drift to co-linear,
causing NaN's. I tied to pick make the algorithm not include new
dependencies like a graph module (since its so simple), but its a bit
clunky right now and could use a look over if you could, @andrrizzi

The energy function of the torsion restraint in the Boresch is now

```
(K/2) * (1-cos(phi - phi0)) * (pi^2)/2
```
which is an always positive function wrapping at `phi-phi0 = [0, 2*pi]`.
The `pi^2` scaling factor ensures the function has the same maximum as the
`(phi-phi0)^2` harmonic function, but does have a steeper slope overall.
I'm open to discussion on this.

Because the functional form changed, there is no longer an analytical
solution to the standard state correction (its all numeric),
everything related to the `standard_state_correction_method` has been removed

cc @jchodera for discussion on the math parts and energy function choices.

### DIFF
--- a/Yank/restraints.py
+++ b/Yank/restraints.py
@@ -21,7 +21,6 @@ standard state correction.
 # ==============================================================================
 
 import abc
-import math
 import random
 import inspect
 import logging
@@ -33,8 +32,9 @@ import scipy.integrate
 import mdtraj as md
 import openmmtools as mmtools
 from simtk import openmm, unit
+import scipy.sparse.csgraph as cs
 
-from . import pipeline, utils
+from . import pipeline
 from .utils import methoddispatch
 
 logger = logging.getLogger(__name__)
@@ -704,7 +704,8 @@ class RadiallySymmetricRestraint(ReceptorLigandRestraint):
     # Internal-usage: properties and methods for subclasses.
     # -------------------------------------------------------------------------
 
-    @abc.abstractproperty
+    @property
+    @abc.abstractmethod
     def _are_restraint_parameters_defined(self):
         """bool: True if the restraint parameters are defined."""
         pass
@@ -1297,13 +1298,12 @@ class Boresch(ReceptorLigandRestraint):
                     K_r/2 * [|r3 - l1| - r_aA0]^2 +
                     + K_thetaA/2 * [angle(r2,r3,l1) - theta_A0]^2 +
                     + K_thetaB/2 * [angle(r3,l1,l2) - theta_B0]^2 +
-                    + K_phiA/2 * [dihedral(r1,r2,r3,l1) - phi_A0]^2 +
-                    + K_phiB/2 * [dihedral(r2,r3,l1,l2) - phi_B0]^2 +
-                    + K_phiC/2 * [dihedral(r3,l1,l2,l3) - phi_C0]^2
+                    + K_phiA/2 * hav[dihedral(r1,r2,r3,l1) - phi_A0] * pi^2 +
+                    + K_phiB/2 * hav[dihedral(r2,r3,l1,l2) - phi_B0] * pi^2 +
+                    + K_phiC/2 * hav[dihedral(r3,l1,l2,l3) - phi_C0] * pi^2
                 }
 
-
-    , where the parameters are:
+    , where ``hav`` is the Haversine function ``(1-cos(x))/2`` and the parameters are:
 
         ``r1``, ``r2``, ``r3``: the coordinates of the 3 receptor atoms.
 
@@ -1330,6 +1330,25 @@ class Boresch(ReceptorLigandRestraint):
 
     The class supports automatic determination of the parameters left undefined
     in the constructor through :func:`determine_missing_parameters`.
+
+
+    This function used to be based on the Boresch orientational restraints [1] and has similar form to
+    its energy equation
+
+        .. code-block:: python
+
+            E = lambda_restraints * {
+                    K_r/2 * [|r3 - l1| - r_aA0]^2 +
+                    + K_thetaA/2 * [angle(r2,r3,l1) - theta_A0]^2 +
+                    + K_thetaB/2 * [angle(r3,l1,l2) - theta_B0]^2 +
+                    + K_phiA/2 * [dihedral(r1,r2,r3,l1) - phi_A0]^2 +
+                    + K_phiB/2 * [dihedral(r2,r3,l1,l2) - phi_B0]^2 +
+                    + K_phiC/2 * [dihedral(r3,l1,l2,l3) - phi_C0]^2
+                }
+
+    However, the form at the top is periodic with the dihedral angle and imposes a more steep energy penalty
+    while still maintaining the same maximum at the periodic border.
+
 
     *Warning*: Symmetry corrections for symmetric ligands are not automatically applied.
     See Ref [1] and [2] for more information on correcting for ligand symmetry.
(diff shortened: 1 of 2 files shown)


davidlmobley:
cc @elkhoury who this will affect.

Yank/restraints.py
```diff
@@ -33,8 +32,9 @@
 import mdtraj as md
 import openmmtools as mmtools
 from simtk import openmm, unit
+import scipy.sparse.csgraph as cs
```
andrrizzi: This may be unused.

jchodera:
> The pi^2 scaling factor ensures the function has the same maximum as the
(phi-phi0)^2 harmonic function, but does have a steeper slope overall.
I'm open to discussion on this.

The maximum isn't super relevant---I think we should be aiming to match the force constant for the Taylor expansion around `dphi = 0`.

Yank/restraints.py
```diff
@@ -1297,13 +1298,12 @@ class Boresch(ReceptorLigandRestraint):
                     K_r/2 * [|r3 - l1| - r_aA0]^2 +
                     + K_thetaA/2 * [angle(r2,r3,l1) - theta_A0]^2 +
                     + K_thetaB/2 * [angle(r3,l1,l2) - theta_B0]^2 +
-                    + K_phiA/2 * [dihedral(r1,r2,r3,l1) - phi_A0]^2 +
-                    + K_phiB/2 * [dihedral(r2,r3,l1,l2) - phi_B0]^2 +
-                    + K_phiC/2 * [dihedral(r3,l1,l2,l3) - phi_C0]^2
+                    + K_phiA/2 * hav[dihedral(r1,r2,r3,l1) - phi_A0] * pi^2 +
```
jchodera: Functions are always written `f(x)`, not `f[x]`, right?

Lnaden: Yes, I think that was leftover from the other nested parenthesis from the previous form, I'll fix that on a second pass

Yank/restraints.py
```diff
@@ -1914,26 +1844,62 @@ def compute_atom_str(input_string, topography_key):
             raise ValueError(error_msg.format(min_distance, max_distance))
         r3_l1_pairs = r3_l1_pairs[indices_of_in_range_pairs].tolist()
 
+        def find_bonded_to(input_atom_index, comparison_set):
```
jchodera: It isn't easier to just iterate through forcefield `HarmonicAngleForce` terms? The advantage of doing that would also be that we would be guaranteed none of those would have `theta0 = 0` or `180 degrees`, while just looking at three bonded atoms could mean they are still collinear.

jchodera:
Looks pretty good, modulo some comments!

Yank/restraints.py
Lnaden: I thought about that, and that would require a a different enumeration through the thermodynamic_state -> system -> forces -> harmonic angle force. Although this internal function does not accept a thermodynamic state, the only time its called is from the outer function which does, so I can make that change pretty easily, this was just working with what I already had to try and minimize the structural changes.

If you think that is a better option, I can rework it

jchodera: What do you think @andrrizzi?

jchodera:
One more request: Can we make sure that even internal/private functions we write have docstrings documenting `Parameters` and `Returns`? I consider this absolutely essential if more than one person is developing the code.

Yank/restraints.py
```diff
@@ -1599,33 +1633,33 @@ def determine_missing_parameters(self, thermodynamic_state, sampler_state, topog
...
+                self.restrained_ligand_atoms = initial_restrained_ligand
+                self.restrained_receptor_atoms = initial_restrained_receptor
                 restrained_atoms = self._pick_restrained_atoms(sampler_state, topography)
                 self.restrained_receptor_atoms = restrained_atoms[:3]
                 self.restrained_ligand_atoms = restrained_atoms[3:]
 
                 # Determine restraint parameters for these atoms.
                 self._determine_restraint_parameters(sampler_state, topography)
 
+                # TODO: JDC/AXR do we need a check for robustness now that we are on numerical only?
```
Lnaden: I realized in my bugfixes that we no longer have the analytical correction, so is this loop where we check robustness even needed anymore?

Yank/restraints.py
andrrizzi: If it further cuts the number of lines of code, I'm +1.

andrrizzi:
> I could subclass this, or we could make a type of Boresch-like abstract class?

Between the two, I think I would prefer a hierarchy with a partially abstract parent class.

> Do you need SAMS with the original Boresch to validate the bromodomains?

I'm not 100% sure. Probably HREX is enough, but I'll have to use the latest version of YANK anyway for the paper.

Yank/restraints.py
andrrizzi: If we're going to use only numerical also for the old functional form, I'm +1 on removing this check.

jchodera:
@Lnaden : We want to preserve the "force constant"---the leading linear coefficient in front of `dphi` in the force---not the force at `dphi = 0`.

For a spring,
```
U(x) = (K/2) x^2
F(x) = - K x
```
`K` is the force constant.

Lnaden:
I didn't change the force constant?

The energy expression for the torsions changed from
```
(K_phiA/2)*dphi_A^2 
dphi_A = dA - floor(dA/(2*pi)+0.5)*(2*pi); 
dA = dihedral(p1,p2,p3,p4) - phi_A0;
```
(which is just a correction for `phi-phi0` in the limits due to OpenMM dihedral computation at the limits, if i remember correctly).

And modified it to:
```
(K_phiA/2)*uphi_A +
uphi_A = pi^2 * (1-cos(dA))/2; 
dA = dihedral(p1,p2,p3,p4) - phi_A0;
```

The leading `K/2` term did not change.

jchodera:
@Lnaden : Consider a Taylor series expansion of the potential `U(dphi) = f(dphi)` about the `dphi = 0`:
```
U(phi) = U(0) + U'(0) * dphi + (1/2) U''(0) * dphi^2 + O(dphi^3)
```
Here, `U'(dphi)` is the first derivative and `U''(dphi)` is the second derivative.

If the minimum energy is at `dphi = 0`, then `U'(0) = 0` and we have
```
U(phi) = U(0) + (1/2) U''(0) * dphi^2 + O(dphi^3)
```
which is approximately equal to the harmonic form `U(0) + (K/2) * dphi^2` near `dphi = 0`.

When I say that we want to preserve the force constant, I mean we want the constant `K` to be the same when you switch from the original form
```
(K_phi/2)*dphi^2
```
to the new form
```
alpha * (1-cos(dphi)) ~ alpha * (1 - [1 - (1/2) dphi^2]) = alpha * (1/2) dphi^2
```
where `alpha` is the constant we want to determine so as to have
```
(K_phi/2)*dphi^2 = alpha * (1/2) dphi^2
```
You can therefore see that we want `alpha = K_phi`, and so you want to use the modified form
```
(K_phiA/2)*uphi_A;
uphi_A = (1-cos(dA)); 
dA = dihedral(p1,p2,p3,p4) - phi_A0;
```

Lnaden:
I fixed the scaling factor on the `K/2*alpha` term, removed the `pi^2` term.

I also generalized the `Boresch` restraint class into a `BoreschLike`to allow different functional forms. The `Boresch` sub-class uses the normal harmonic functions which can be found in the paper. The "Haversined Torsional Boresch," `HaverBorTor`, class uses the periodic torsional I implemented. 

@jchodera @andrrizzi if you could look over the refactor part, and probably help with the name since I think "HaverBorTor" is not a great name.

Lnaden:
The increased line counts mostly comes from the abstract class + a bunch of doc string

jchodera:
Thanks!

devtools/conda-recipe/meta.yaml
```diff
@@ -24,7 +24,7 @@ requirements:
     - numpy >=1.11
     - scipy
     - cython
-    - netcdf4
+    - netcdf4 ==1.3.1  # TODO: Fix this right after bugfix: "always return masked array by default, even if there are no masked values"
```
jchodera: Do you want to add this to the changelog too?

PR #986, opened 2018-05-11
Change timeseries to only consider a subset of nodes for equilibration

### LINKED ISSUE #980: Speed up `yank analyze` to usable speeds by eliminating time-consuming unnecessary steps
Currently, `yank analyze` run on SAMS trajectories takes an horrifically long time.

We need to prune out inessential computations and do what we can to accelerate this.

Some ideas:
* Eliminate the calls to [`show_mixing_statistics`](https://github.com/choderalab/yank/blob/master/Yank/analyze.py#L110) in calls to [`analyze_phase()`](https://github.com/choderalab/yank/blob/master/Yank/analyze.py#L108) since this data is only printed and not captured in any way. This call is slow.
* The call to [`pymbar.timeseries.detectEquilibration()`](https://github.com/choderalab/yank/blob/master/Yank/multistate/utils.py#L111) currently uses [`nskip=1`](https://github.com/choderalab/pymbar/blob/master/pymbar/timeseries.py#L739-L747) by default. This should be automatically selected so that ~40-100 time origins are considered for equilibration, since there is no point in considering every snapshot in a 30,000-snapshot energy trace a as a potential time origin.

Lnaden:
The analyze modules now will only consider up to 1000 points by default in the timeseries fed into them for equilibration. This is a tunable parameter.

Added warning for future removal of the `get_equilibration_data` in `multistate.utils`

The mixing stats do not print out by default anymore when you invoke `analyze_phase`

All of these changes should make analysis faster.

Fixes #980

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -105,7 +105,35 @@ class YankMultiStateSamplerAnalyzer(multistate.MultiStateSamplerAnalyzer, YankPh
             self._computed_observables['standard_state_correction'] = ssc
         return self._computed_observables['standard_state_correction']
 
-    def analyze_phase(self, cutoff=0.05):
+    def analyze_phase(self, show_mixing=False, cutoff=0.05):
+        """
+        Auto-analysis function for the phase
+
+        Function which broadly handles "auto-analysis" for those that do not wish to call all the methods on their own.
+
+        This variant computes the following:
+
+        * Equilibration data (accessible through ``n_equilibration_iterations`` and ``statistical_inefficiency``
+          properties)
+        * Optional mixing statistics printout
+        * Free Energy difference between all states with error
+        * Enthalpy difference between all states with error (as expectation of the reduced potential)
+        * Free energy of the Standard State Correction for the phase.
+
+        Parameters
+        ----------
+        show_mixing : bool, optional. Default: False
+            Toggle to show mixing statistics or not. This can be a slow step so is disabled for speed by default.
+        cutoff : float, optional. Default: 0.05
+            Threshold below which % of mixing mixing from one state to another is not shown.
+            Makes the output table more clean to read (rendered as whitespace)
+
+        Returns
+        -------
+        data : dict
+            A dictionary of analysis objects, in this case Delta F, Delta H, and Standard State Free Energy
+            In units of kT (dimensionless)
+        """
         number_equilibrated, g_t, _ = self._equilibration_data
         self.show_mixing_statistics(cutoff=cutoff, number_equilibrated=number_equilibrated)
         data = {}
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -135,7 +135,8 @@ class YankMultiStateSamplerAnalyzer(multistate.MultiStateSamplerAnalyzer, YankPh
             In units of kT (dimensionless)
         """
         number_equilibrated, g_t, _ = self._equilibration_data
-        self.show_mixing_statistics(cutoff=cutoff, number_equilibrated=number_equilibrated)
+        if show_mixing:
+            self.show_mixing_statistics(cutoff=cutoff, number_equilibrated=number_equilibrated)
         data = {}
         # Accumulate free energy differences
         Deltaf_ij, dDeltaf_ij = self.get_free_energy()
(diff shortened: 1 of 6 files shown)


Lnaden:
I'm still generating some data to test my notebook changes, but I think they are correct and I wanted to open the PR for review now

Lnaden:
There are a couple bugs I am finding with the test with real data, fixing them

Yank/analyze.py
```diff
@@ -105,7 +105,35 @@ def get_standard_state_correction(self):
             self._computed_observables['standard_state_correction'] = ssc
         return self._computed_observables['standard_state_correction']
 
-    def analyze_phase(self, cutoff=0.05):
+    def analyze_phase(self, show_mixing=False, cutoff=0.05):
```
andrrizzi: I don't think the new `show_mixing` argument is used here.

Yank/multistate/multistateanalyzer.py
```diff
@@ -1840,7 +1840,12 @@ def _get_equilibration_data(self, energies=None, replica_state_indices=None):
         # Discard equilibration samples.
         # TODO: if we include u_n[0] (the energy right after minimization) in the equilibration detection,
         # TODO:         then number_equilibrated is 0. Find a better way than just discarding first frame.
-        equilibration_data = list(utils.get_equilibration_data(u_n[1:]))
+        i_t, g_i, n_effective_i = utils.get_equilibration_data_per_sample(u_n[1:])
+        n_effective_max = n_effective_i.max()
+        i_max = n_effective_i.argmax()
+        n_equilibration = i_t[i_max]
+        g_t = g_i[i_max]
+        equilibration_data = [n_equilibration, g_t, n_effective_max]
```
andrrizzi: Apologies if I'm missing something. What's the difference between using this and keep using `utils.get_equilibration_data()` as before?

Yank/reports/notebook.py
```diff
@@ -682,7 +687,11 @@ def format_trace_plot(plot: plt.Axes, trace_forward: np.ndarray, trace_reverse:
                         (u_forward, s_forward, free_energy_trace_f), (u_reverse, s_reverse, free_energy_trace_r)]:
                     u_n = analyzer.get_effective_energy_timeseries(energies=energy_sub,
                                                                    replica_state_indices=state_sub)
-                    number_equilibrated, g_t, neff_max = analyze.multistate.utils.get_equilibration_data(u_n)
+
+                    i_t, g_i, n_effective_i = analyze.multistate.get_equilibration_data_per_sample(u_n)
+                    i_max = n_effective_i.argmax()
+                    number_equilibrated = i_t[i_max]
+                    g_t = g_i[i_max]
```
andrrizzi: Same here (i.e. use directly `utils.get_equilibration_data()`?

Yank/analyze.py
Lnaden: Yeah, I saw that in my testing, easy fix.

Yank/multistate/multistateanalyzer.py
Lnaden: Good question, that is by design. the `get_equilibration_data_per_sample` is more general than `get_equilibration_data` so I wanted to actually remove `get_equilibration_data` in a future release. To that end, I took all the code which called it and removed its called, but others who use the API may not, so it trows a warning now.

Lnaden:
There are still some issues with online analysis, so I'm working on that with a couple other small fixes

jchodera:
> The analyze modules now will only consider up to 1000 points by default in the timeseries fed into them for equilibration. 

Why 1000? If it's detectably slower than 100, let's just use 100 as default.

No need to fix online analysis right now---can we just get this merged ASAP and a new release cut?

Lnaden:
1000 is not detectable slower than the 100 on the Cmet set of ~40k iterations. Each one takes about 22 min to process now, which is better than the one hour it was taking. I think the MBAR online analysis is the next speedup-point

Lnaden:
But, Ill change it

Lnaden:
@andrrizzi you had mentioned you had some last PR stuff to the whats new doc. Can you open a PR with that and set the setup.py version to 0.22? I will handle all the merging and conflict issues and cut the release today

jchodera:
Thanks!

> 1000 is not detectable slower than the 100 on the Cmet set of ~40k iterations. Each one takes about 22 min to process now, which is better than the one hour it was taking. I think the MBAR online analysis is the next speedup-point

I'm a bit confused here: we are first trying to speed up offline analysis via "yank analyze", so I'm not sure how speeding up online analysis will help that. 

I'll see if I can profile "yank analyze" to see what is taking so long.

jchodera:
> Sorry. I meant we need to make offline analysis pull from online analysis if its present. That would be the next point of speed up is what I am saying.

Ah! I get it now---thanks for the clarification!

PR #987, opened 2018-05-12
Speed up analysis by only retrieving end thermodynamic states as needed

jchodera:
Examining logging timesteps suggests the slowest step in `yank analyze` is currently retrieval of the end thermodynamic states:
```
[chodera@lilac:c-Met_hinge_binders]$ yank analyze --store sams-twostage-rmsd-mcligand-dense/experiments/cmethingesinglehinge0
2018-05-12 11:05:01,293: Reading energies...
2018-05-12 11:05:01,629: Done.
2018-05-12 11:05:01,630: Assembling effective timeseries...
2018-05-12 11:05:01,973: Done.
2018-05-12 11:05:01,993: Checking if we need to unbias the restraint...
2018-05-12 11:05:01,993: Trying to get radially symmetric restraint data...
2018-05-12 11:05:01,993: Retrieving end thermodynamic states...
2018-05-12 11:16:09,873: Isolating restraint force...
2018-05-12 11:16:27,693: No force of type class 'openmmtools.forces.RadiallySymmetricRestraintForce' could be found. The restraint will not be unbiased.
2018-05-12 11:16:27,758: Reading energies...
2018-05-12 11:16:27,778: Done.
2018-05-12 11:16:27,778: Assembling effective timeseries...
2018-05-12 11:16:27,985: Done.
2018-05-12 11:16:28,004: Assembling uncorrelated energies...
2018-05-12 11:16:28,081: Found expanded cutoff states in the energies!
2018-05-12 11:16:28,081: Free energies will be reported relative to them instead!
2018-05-12 11:16:28,081: Done.
2018-05-12 11:16:28,084: Computing free energy differences...
2018-05-12 11:16:30,663: Done.
2018-05-12 11:16:30,664: Computing covariance matrix...
2018-05-12 11:18:35,459: Deltaf_ij:
```
Specifically, [this line](https://github.com/choderalab/yank/blob/master/Yank/multistate/multistateanalyzer.py#L1231) takes approximately ten minutes, regardless of the length of the output file:
```
end_states = self._get_end_thermodynamic_states()
```
The call to [`_get_end_thermodynamic_states()`](https://github.com/choderalab/yank/blob/master/Yank/multistate/multistateanalyzer.py#L614-L626) unnecessarily reads all the thermodynamic states and discards all but the first and last. I have no idea why this takes ten minutes---something we have to fix in the future---but this is trivial to speed up by only retrieving the two states we need.

This PR adds a reporter method to just retrieve the end states, which reduces the time for this step from 10 minutes to 1 minute:
```
2018-05-12 12:10:35,976: Retrieving end thermodynamic states...
2018-05-12 12:11:40,464: Isolating restraint force...
```

### DIFF
--- a/Yank/multistate/multistatereporter.py
+++ b/Yank/multistate/multistatereporter.py
@@ -490,7 +490,6 @@ class MultiStateReporter(object):
 
                 # Create ThermodynamicState object.
                 states[state_type].append(mmtools.utils.deserialize(serialized_state))
-                state_id += 1
 
         return [states['thermodynamic_states'], states['unsampled_states']]
 


jchodera:
More generally, *why* are we retrieving the end thermodynamic states anyway? Do we even use this information if we're not computing restraint energies?

jchodera:
Or is it just to [retrieve `kT`](https://github.com/choderalab/yank/blob/master/Yank/multistate/multistateanalyzer.py#L629-L638)?
```python
    @property
    def kT(self):
        """
        Quantity of boltzmann constant times temperature of the phase in units of energy per mol
        Allows conversion between dimensionless energy and unit bearing energy
        """
        if self._kT is None:
            self._get_end_thermodynamic_states()
        return self._kT
```
Also, is this even correct if the end states differ in temperature?

jchodera:
It seems like we also use it [here](https://github.com/choderalab/yank/blob/master/Yank/multistate/multistateanalyzer.py#L1226-L1252) in unpacking a restraint force.

@andrrizzi : Instead of spending a huge amount of time deserializing these things, can't we just store (1) `kT` and (2) the serialized restraint force in metadata for a YANK calculation? It seems like that would speed things up by another couple orders of magnitude.

Lnaden:
I think its because of the `kT` we need the thermodynamic states before unbiasing, I think we need them now that we have unbiasing, but @andrrizzi can say more.

> Also, is this even correct if the end states differ in temperature?

No, but all the properties we care about are evaluated at the same temp since we need the ratio of partition functions. The assumption we went with was you will only use the kT in the analyzer for converting the final property value to real units, which is measured between the endstates. Internally, everything is in kT, so the math still works. I have decided to just go with this assumption and known limitation until someone actually requests the engineering to handle every state by its own kT, which would take a bunch of re-factoring.

jchodera:
> I think its because of the kT we need the thermodynamic states before unbiasing, I think we need them now that we have unbiasing, but @andrrizzi can say more.

1. Do we need them if we are not unbiasing?
2. Can we store `kT` elsewhere?

> No, but all the properties we care about are evaluated at the same temp since we need the ratio of partition functions. The assumption we went with was you will only use the kT in the analyzer for converting the final property value to real units, which is measured between the endstates. Internally, everything is in kT, so the math still works. I have decided to just go with this assumption and known limitation until someone actually requests the engineering to handle every state by its own kT, which would take a bunch of re-factoring.

That doesn't make any sense. If everything is in reduced units, partition function ratios and reduced free energy differences are perfectly well-defined. The only time we should need `kT` is when we are computing new non-reduced energies.

If we *do* assume `kT` is the same for all states, we should probably `assert` that at some point to ensure that is satisfied.

andrrizzi:
There is an easy optimization that we can implement to retrieve only the temperature from the thermodynamic state stored in the netcdf file. That won't be a problem with the current storage format.

> If we do assume kT is the same for all states, we should probably assert that at some point to ensure that is satisfied.

If I remember correctly, we do check that the temperatures are the same at the end-points before computing their DG.

> (2) the serialized restraint force in metadata for a YANK calculation? It seems like that would speed things up by another couple orders of magnitude.

If we need to recompute distances and energies, this won't save much time, but we'll add complexity to the code and lose generality. If we end up implementing the restraints with the `CVForce` as we discussed and figure out an extensible way for storing the distances, both of these won't be necessary so I would keep it as it is until we change the way the restraints are handled altogether.

jchodera:
> If we need to recompute distances and energies, this won't save much time, but we'll add complexity to the code and lose generality. If we end up implementing the restraints with the CVForce as we discussed and figure out an extensible way for storing the distances, both of these won't be necessary so I would keep it as it is until we change the way the restraints are handled altogether.

I think refactoring all the restraints to use `CVForce` is the best way forward here.

Yank/multistate/multistateanalyzer.py
```diff
@@ -612,18 +612,18 @@ def n_states(self):
...
-            self._kT = self._end_thermodynamic_states[0].kT
+        """Read thermodynamic states at the ends of the protocol.
+
+        Returns
+        -------
+        end_thermodynamic_states : list of ThermodynamicState
+            The end thermodynamic states
+
+        """
+        self._end_thermodynamic_states = self._reporter.read_end_thermodynamic_states()
```
andrrizzi: Hmm. Why removing the `if self._end_thermodynamic_states is None` here? This will read twice the states during a normal analysis with restraint unbiasing.

jchodera: Good catch! That was unintentional. I'll fix in the next PR.

jchodera:
> Personally, I would have implemented this by augumenting reporter.read_thermodynamic_state() with an argument specifying which state indices to retrieve rather than creating a new method. I think it's more general and it doesn't make the assumption that the unsampled states are the end states. But we can make the modification later if you agree.

Please go for it! If there *was* a `read_thermodynamic_state()` method, I would have used that, but we only have a method `read_thermodynamic_states()` that reads *all* of them and uses some undocumented property of some thermodynamic states to achieve some sort of complex compression I was not familiar with, so this was the only change I was able to confidently make.

We should start thinking about how to expose a clean, simple, consistent API for the Reporter in case we ever want to ditch the NetCDF backend.

andrrizzi:
> Please go for it!

Ok, will do! I'll open an issue for now describing the optimizations I have in mind and I'll put it in my queue for later since we are in a rush for this PR.

Lnaden:
Besides the whatsnew pr you opened and the fix to the `if self._end_thermodynamic_states is None`, is there anything else we need before we cut a release?

andrrizzi:
Not on my side!

andrrizzi:
Ok! I'll modify the release history in my other PR.

andrrizzi:
Hmm. Actually, we are breaking backward compatibility since we're not offering an analytical Boresch correction anymore.

jchodera:
Ah! Ok, let's go for a new .0 release then

Lnaden:
Is there anything else we want to put it in then? If not, I'll get a release going

andrrizzi:
I think we're good!

PR #995, opened 2018-05-22
Analyze pulls from Online analysis if present

### LINKED ISSUE #979: Improve `yank analyze` integration with online analysis
To speed up analysis and user experience with analysis in general, we want to rely more on online analysis.

To help do this, we want to:
* Enable online analysis by default, minimizing its cost as much as possible by using a larger checkpoint interval
* Have `yank analyze` pull from the online analysis data by default, adding a flag to force recomputation if desired

Lnaden:
This is the realization of #979. It is not everything on that list due to some practical considerations, but here is the full list of things changed:

* Online analysis interval can now be set to checkpoint interval 
* Online analysis runs by default
* Online analysis min iterations now 200
* Checkpoint interval now 200
* Analysis now tries to get the online data by default 
* Online analysis does not attempt unbias harmonic restraints, this can be relaxed when we switch to CustomCVForces for restraint distances

Things not implemented from the issue list:
* Do not allow users to specify where the reference states are in the YAML
* Only uses the f_k as the initial MBAR guess to speed up the process, it does not use anything else. MBAR still has to be solved, but should be much faster with the online guess than not. Generalizing this so each component specifically reads entries in the online data would require re-engineering the entire analysis and online code + making users specify the reference states.

Fix #979

### DIFF
--- a/Yank/multistate/multistateanalyzer.py
+++ b/Yank/multistate/multistateanalyzer.py
@@ -480,10 +480,20 @@ class PhaseAnalyzer(ABC):
         For instance, the initial guess of relative free energies to give to MBAR would be something like:
         ``{'initial_f_k':[0,1,2,3]}``
 
+    use_online_data : bool, optional, Default: True
+        Attempt to read online analysis data as a way to hot-start the analysis computation. This will attempt to
+        read the data stored by the MultiStateAnalyzer.
+
+        If this is set to ``False``, the online analysis data is not read.
+
+        If this is set to ``False`` after being initialized, the :class:`CachedProperty` dependencies are all
+        invalidated and properties will be computed from scratch on next observables
+
+        If no online data is found, this setting has no effect
+
     registry : ObservablesRegistry instance
         Instanced ObservablesRegistry with all observables implemented through a ``get_X`` function classified and
         registered. Any cross-phase analysis must use the same instance of an ObservablesRegistry
-
 
     Attributes
     ----------
@@ -497,6 +507,7 @@ class PhaseAnalyzer(ABC):
     kT
     reporter
     registry
+    use_online_data
 
     See Also
     --------
@@ -505,7 +516,8 @@ class PhaseAnalyzer(ABC):
     """
     def __init__(self, reporter, name=None, reference_states=(0, -1),
                  max_n_iterations=None, analysis_kwargs=None,
-                 registry=default_observables_registry):
+                 registry=default_observables_registry,
+                 use_online_data=True):
         """
         The reporter provides the hook into how to read the data, all other options control where differences are
         measured from and how each phase interfaces with other phases.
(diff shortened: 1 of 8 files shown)


Lnaden:
Ability for Sampler Online Analysis interval to be equal to the reporter checkpoint interval now handled at the application layer, instead of the API layer.

I have written a test which can handle all the corner cases I could think of, and good thing too because there were a bunch of them.

jchodera:
> Online analysis min iterations now 200
> Checkpoint interval now 200

This is great for now, but longer-term, instead of having an arbitrary minimum number of iterations, do we want something like a minimum number of fractional state visits? That way, we would adaptively switch on online analysis at an appropriate time for all of our multistate samplers and have it be meaningful.

Lnaden:
Is this good to merge?

Yank/multistate/multistateanalyzer.py
```diff
@@ -692,7 +729,7 @@ def _update_cache(self, key, new_value, check_changes=False):
         except KeyError:
             invalidate_cache = False
         else:
-            if check_changes and old_value == new_value:
+            if check_changes and self._check_equal(old_value, new_value):
```
andrrizzi: Just for curiosity, what does this do differently than `==` equality?

Yank/multistate/multistatesampler.py
```diff
@@ -82,22 +82,28 @@ class MultiStateSampler(object):
         The number of iterations to perform. Both ``float('inf')`` and
         ``numpy.inf`` are accepted for infinity. If you set this to infinity,
         be sure to set also ``online_analysis_interval``.
-    online_analysis_interval : None or Int >= 1, optional, default None
+    online_analysis_interval : = None or Int >= 1, optional, default: 200
```
andrrizzi: Typo `=`?

andrrizzi:
Sorry for the dealy, I'm doing my review now!

Yank/multistate/multistatesampler.py
```diff
@@ -381,28 +390,30 @@ def __set__(self, instance, new_value):
...
             return number_of_iterations
 
         @staticmethod
-        def _oa_interval_validator(instance, online_analysis_interval):
+        def _oa_interval_validator(_, online_analysis_interval):
             """Check the online_analysis_interval value for consistency"""
             if online_analysis_interval is not None and (
                             type(online_analysis_interval) != int or online_analysis_interval < 1):
-                raise ValueError("online_analysis_interval must be an integer 1 or greater, or None")
+                import pdb; pdb.set_trace()
```
andrrizzi: debug

Yank/multistate/multistateanalyzer.py
Lnaden: Numpy arrays dont do `if array == array` without a `any` or `all`, and `all` does not work on non iterable.

Yank/multistate/multistatesampler.py
Lnaden: Oh, good catch. Thanks

PR #1006, opened 2018-05-29
Easy fixes for Milestone 0.22.3

### LINKED ISSUE #999: Online analysis failures should not kill entire YANK job
I am running the [c-Met benchmarks](https://github.com/choderalab/yank-benchmark/tree/jdc-sams-tweaks/c-Met_hinge_binders), which are dying via online analysis failures such as this:
```
  File "/home/chodera/miniconda/bin/yank", line 11, in <module>
    load_entry_point('yank==0.22.1', 'console_scripts', 'yank')()
  File "/home/chodera/miniconda/lib/python3.6/site-packages/yank/cli.py", line 73, in main
    dispatched = getattr(commands, command).dispatch(command_args)
  File "/home/chodera/miniconda/lib/python3.6/site-packages/yank/commands/script.py", line 136, in dispatch
    yaml_builder.run_experiments()
  File "/home/chodera/miniconda/lib/python3.6/site-packages/yank/experiment.py", line 793, in run_experiments
    send_results_to='all')
  File "/home/chodera/miniconda/lib/python3.6/site-packages/yank/mpi.py", line 520, in distribute
    *other_args, **kwargs)
  File "/home/chodera/miniconda/lib/python3.6/site-packages/yank/mpi.py", line 375, in exec_tasks
    raise all_errors[0]
RuntimeError: Group 1/13 Node 1/1: Online Analysis has failed too many times! Please check the latest logs to see the thrown errors!
```
We do *not* want the whole job to die if the online analysis fails. The only reason the online analysis is being performed is to (1) see if we want to terminate the job early, and (2) save time in post-simulation analysis. Neither of these tasks should require the entire calculation to be terminated should the online analysis fail.

### LINKED ISSUE #1000: Online analysis YAML documentation is broken
There's a section in the YAML docs that points to online analysis parameters:
image
Clicking on this link reveals no online analysis parameters in sight:
http://getyank.org/latest/yamlpages/samplers.html#yaml-samplers-online-analysis-parameters

Lnaden:
Easy fixes:

Fixes #999 by preventing the online analysis MBAR failure accumulation from stopping the simulation (will still fail if something else goes catastrophically wrong)

Fixes #1000 by doing some doc restructuring. There is a better fix which involves figuring out how to do recursive function calls in the javascript... I decided to do the easier one.

### DIFF
--- a/Yank/multistate/multistatesampler.py
+++ b/Yank/multistate/multistatesampler.py
@@ -1433,6 +1433,10 @@ class MultiStateSampler(object):
             # wasn't below the target threshold before, it won't stop MultiStateSampler now.
             bump_error_counter = True
             self._online_error_bank.append(e)
+            if len(self._online_error_bank) > 6:
+                # Cache only the last set
+                self._online_error_bank.pop(0)
+            free_energy = None
         else:
             self._last_mbar_f_k = mbar.f_k
             free_energy = free_energy[idx, jdx]
@@ -1447,12 +1451,13 @@ class MultiStateSampler(object):
         # Raise an exception after 6 times MBAR gave an error.
         if bump_error_counter:
             self._online_error_trap_counter += 1
-            if self._online_error_trap_counter >= 6:
+            # Will never be true, but code left in place in case we change logic to allow again
+            if self._online_error_trap_counter >= np.inf:
                 logger.debug("Thrown MBAR Errors:")
                 for err in self._online_error_bank:
                     logger.debug(str(err))
                 raise RuntimeError("Online Analysis has failed too many times! Please "
-                                   "check the latest logs to see the thrown errors!")
+                                   "check the latest logs to see the latest thrown errors!")
             # Don't write out the free energy in case of error.
             return
 
--- a/docs/whatsnew.rst
+++ b/docs/whatsnew.rst
@@ -9,6 +9,7 @@ The full release history can be viewed `at the GitHub yank releases page <https:
 0.22.2 Topography Property Copy
 -------------------------------
 - Critical bug fix for Topology where ions of charged ligands were considered part of the ligand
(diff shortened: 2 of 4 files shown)


docs/whatsnew.rst
```diff
@@ -9,6 +9,7 @@ The full release history can be viewed `at the GitHub yank releases page <https:
 0.22.2 Topography Property Copy
 -------------------------------
 - Critical bug fix for Topology where ions of charged ligands were considered part of the ligand
+- Online analysis MBAR failures can no longer halt simulations
```
jchodera: Shouldn't this be listed under `0.22.3`, not `0.22.2`?

Yank/multistate/multistatesampler.py
```diff
@@ -1447,12 +1451,13 @@ def _offline_analysis(self):
         # Raise an exception after 6 times MBAR gave an error.
         if bump_error_counter:
             self._online_error_trap_counter += 1
-            if self._online_error_trap_counter >= 6:
+            # Will never be true, but code left in place in case we change logic to allow again
+            if self._online_error_trap_counter >= np.inf:
```
jchodera: What if we add a YAML option that governs how we treat errors like this? We can come up with a proposal for that in an issue and leave that implementation to a future release.

docs/whatsnew.rst
Lnaden: I never cut a 0.22.2, so this would sill fall under that

Yank/multistate/multistatesampler.py
Lnaden: I can make this an option, It would be pretty strait forward

jchodera: Shoot, sorry--my mistake. Can you fix the milestones for the tagged issues then?

PR #1009, opened 2018-05-30
Add option to force full trajectory analysis

Lnaden:
Added CLI and API options to force analyzer to use the full trajectory instead of the auto-equilibration and decorelation of the trajectory to compute the free energy.

Decorrelation data can still be computed, but any thing which creates the MBAR object or anything from that automatically will use the full trajectory

### DIFF
--- a/Yank/multistate/multistateanalyzer.py
+++ b/Yank/multistate/multistateanalyzer.py
@@ -491,6 +491,16 @@ class PhaseAnalyzer(ABC):
 
         If no online data is found, this setting has no effect
 
+    use_full_trajectory : bool, optional, Default: False
+        Force the analysis to use the full trajectory when automatically computing equilibration and decorrelation.
+
+        Normal behavior (when this is False) is to discard the initial trajectory due to automatic equilibration
+        detection, and then subsample that data to generate decorelated samples. Setting this to ``True`` ignores
+        this effect, even if the equilibration data is computed.
+
+        This can be changed after the ``PhaseAnalyzer`` has been created to re-compute properties with or without
+        the full trajectory.
+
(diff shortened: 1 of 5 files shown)


Yank/multistate/multistateanalyzer.py
```diff
@@ -2009,7 +2034,7 @@ def _decorrelated_N_l(self, instance):
     _unbiased_decorrelated_u_ln = CachedProperty(
         name='unbiased_decorrelated_u_ln',
         dependencies=['unbias_restraint', 'restraint_energy_cutoff', 'restraint_distance_cutoff',
-                      'decorrelated_state_indices_ln', 'decorrelated_u_ln', 'decorrelated_N_l'],
+                      'decorrelated_state_indices_ln', 'decorrelated_u_ln', 'decorrelated_N_l', 'use_full_trajectory'],
```
andrrizzi: I wouldn't add a dependency on `use_full_trajectory` here, in `_unbiased_decorrelated_N_l`, and `mbar`. Only the properties that use `use_full_trajectory` explicitly in the function should. The chain of dependency will be correctly resolved.

`decorrelated_state_indices_ln` should depend on `use_full_trajectory` though through `self._decorrelated_iterations`.

Lnaden:
Okay, i have fixed the dependency tree and added the check in the `_decorrelated_iterations` call

Lnaden:
> This should work! A heads-up that this will effectively increase the time spent computing restraint distances/energies by 3-4 times.

Fair, but that is to be expected when you force say "analyze everything" vs "analyze a subset." Thats why this is an "opt-in" option because should expect users to know this, and its not standard-use anyways

Issue #1008, opened 2018-05-30
YANK with SAMS performance info


nividic:
Hi there, 
I was trying the latest conda version of YANK 0.22.1 running ABFE on toluene bonded to lysozyme. To perform 1000 YANK iterations (complex state) by using a GTX 1080 the required simulation time ~ 2days and 9hrs. The same system simulated by using YANK 0.20.1 requires ~ 16hrs. Is this simulation  time increases expected? 
Thanks

Lnaden:
Well that does look suspicious!

Could you provide the YAML file(s) you are using to run this job? And are you using this on a cluster or locally; parallel or serial?

The structure of the YAML file changed a bit in 0.21.0, but has defaults to make is somewhat compatible with the old version. So my first guess is that there is a default to make each iteration between the two versions unequal, but we would need more info first.

nividic:
@Lnaden Thanks a lot for your help. I have attached the two yaml files related to the different YANK version that I'm using. The files has been taken from the YANK experiment directory between the two runs. The only difference is the update of the yaml syntax between the two YANK versions as you sad. 

[yamls.zip](https://github.com/choderalab/yank/files/2055337/yamls.zip)

Lnaden:
So from this it looks like the constructed `samplers` and `mcmc` moves will be nearly similar based on your `default_X` settings. However, this is the update where `online_analysis` is turned *on* by default. So every 200 iterations, the simulation is halting to do a full batch of analysis. Although that should still not explain the total time spent, especially with that few of states. 

try setting `online_analysis_interval: null` in the mean time. @andrrizzi any ideas on what else could be causing the slowdown?

jchodera:
Could you include your full input files so we can debug further here?

nividic:
@Lnaden @jchodera Hi there, I have attached the link to the yank files required to reproduce my reported issue between the two yank versions 0.20.1 and 0.22.1. In the directories I left the partial yank generated logs (I interrupted the runs after few yank iterations) to show the discrepancy in the simulation time. Hopefully, you will be able to reproduce the problem. Thank you so much for your help. 

https://drive.google.com/open?id=17r3fJ7uwbB7xQEXjQfe7_rf3aAc_VH3o

jchodera:
Apologies for taking so long to look into this, @nividic!

Both 0.20.1 and 0.22.1 end up using replica-exchange, which is going to be much slower than the new `SAMSSampler`. We should move you to over to that as soon as possible.

[...]

PR #1016, opened 2018-06-15
Reduce the number of calls to NetCDF writing in SamplerState storage

Lnaden:
This changes the `_write_sampler_states_to_given_file` function to put all `SamplerState` positions and box vectors into a single NumPy array before storing to disk. This reduces the number of times the data has to be accessed, which gets slower as the chunk size of the iteration increases (which is now the checkpoint interval). 

This is a pseudo fix to #1008 of which it reduces the write time from 100s to about 5s on 1 GTX 1080. Its still not as fast as if we put the chunksize for the iterations to 1, but its much better. We could reduce the chunksize, but that will then slow down the read access.

Some potential problems which we may have to address later. Not really a "need to answer now," but mostly for documentation.
* We may need to come around on another pass to limit the maximum chunk size so we don't wind up with massive chunks. Not a problem now, but could be with larger checkpoint intervals.
* Systems with huge numbers of particles may run out of memory assigning all replica positions to a single numpy array.

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -236,7 +236,7 @@ class AlchemicalPhaseFactory(object):
         'randomize_ligand_close_cutoff': 1.5 * unit.angstrom,
         'number_of_equilibration_iterations': 0,
         'equilibration_timestep': 1.0 * unit.femtosecond,
-        'checkpoint_interval': 200,
+        'checkpoint_interval': 50,
         'store_solute_trajectory': True
     }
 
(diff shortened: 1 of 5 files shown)


Lnaden:
This PR should speed up every simulation with more than 1 sampler. 

There will technically be speed lost on single-replica simulations but will probably be unnoticeable since the only time increase would be how long it takes to allocate a `[n_replicas x n_particles x 3]` NumPy array in memory.

There will also technically be speed lost if your checkpoint interval is 1, but for the same reason. I don't think it will be noticed either

Lnaden:
I'll be making a cut of YANK once this gets merged in.

PR #1020, opened 2018-06-18
Add draft of "yank analyze cluster" CLI

jchodera:
This PR adds an early experimental version of the `yank analyze cluster` CLI option needed by @steven-albanese for analyzing Id1 data.

The algorithm:
* Compute per-snapshot weights (using MBAR) representing the relative weight of each snapshot in the fully interacting state
* Cluster the remaining snapshots
* Assign relative populations to the clusters
* Sort clusters by population, writing only most populous clusters
* Sample representative snapshots from the clusters proportional to their weights, writing out PDB files
* Write out cluster populations

The code could use a lot of refactoring. It currently accesses the `complex.nc` NetCDF directly, but this could be greatly simplified as we reorganize and refactor the `MultiStateAnalyzer` to expose smaller operational chunks via an API.

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -22,12 +22,17 @@ Extends classes from the MultiStateAnalyzer package to include the
 import os
 import abc
 import yaml
-import mdtraj
 import logging
+import itertools
+
 import numpy as np
-import simtk.unit as units
+import mdtraj as md
+
+import simtk.unit as unit
 import openmmtools as mmtools
-from pymbar import timeseries
+from pymbar import timeseries, MBAR
+from msmbuilder.cluster import RegularSpatial
+
 from . import multistate
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


jchodera:
@Lnaden and @andrrizzi : Maybe you can can pull in the latest changes from `master` since this was opened and review this at some point?

jchodera:
Alan Graves suggests we add an option to cluster the whole receptor+ligand complex coordinates, rather than just the ligand coordinates, in case the ligand and protein change conformation together.

jchodera:
Here's the CLI help for @sgill2:
```
  yank analyze cluster --refpdb=REFPDB --complexnetcdf=FILEPATH [--prefix=PREFIX] [--filter=FILTERDIST] [--cutoff=CUTOFFDIST] [--nsnapshots=NSNAPSHOTS] [--threshold=THRESHOLD] [-v | --verbose]
 
Cluster Required Arguments:
  --refpdb=REFPDB               Reference PDB filename for solvated complex
  --complexnetcdf=FILEPATH      Path to the complex analysis NetCDF file
  --prefix=PREFIX               Prefix to use for output cluster PDB files and populations (default: cluster)
  --filter=FILTERDIST           Discard snapshots where the ligand is farther than this minimum heavy atom distance from the protein, in nanometers (default: 0.3)
  --cutoff=CUTOFFDIST           Heavy-atom RMSD separation between clusters, in nanometers (default: 0.3)
  --nsnapshots=NUM_SNAPSHOTS    Number of snapshots per cluster to write (default: 5)
  --cluster_filter_threshold=THRESHOLD    Threshold to use for which clusters to include (default: 0.95)
```

jchodera:
Thanks! Would be great to get this into the next feature release!

PR #1025, opened 2018-07-02
Auto-Multi-Analysis

Lnaden:
This PR implements 2 major improvements:

1. All YANK "Automatic Analysis" codes go through a singular, common class: the `YankAutoExperimentAnalyzer`. So the API, the Notebooks, and the CLI all invoke and work through the same class. Even the print functions now parse the regularized output. This does change some of the keys you may be used to in other codes.
2. It is now possible to analyze every experiment defined through 1 yaml file through the API and CLI with the `MultiExperimentAnalyzer`, which also supports MPI parallelization. 

The `YankAutoExperimentAnalyzer` can output a singular Python dictionary of all the analysis values it tracked (e.g. if invoked through API, you can get all the numbers instead of dumping to command line), and the `MultiExperimentAnalyzer` can combine those outputs together into one dictionary and serialize to a pickle object, ready for use by downstream, post-yank programs. 

I think I caught all the corner cases, and this update should be drop-in replaceable to any other API code. There might be some CLI code which needs changed though. 

The API docs should be fixed and I also fixed the OpenEye block in `yank selftest` while I was tinkering with the CLI.

There is alot here, but I could not think of how to make it smaller since its all inter-dependent, so it will take some time to go over.


jchodera:
Thanks for putting this together!

A few general comments, though we can likely merge this and address these later if needed:
* Do we need the `YankAuto` prefix for `YankAutoExperimentAnalyzer`? It's in the `yank` package, and it's written in Python so it's probably automated. Could this just be `ExperimentAnalyzer` That would be consistent with`MultiExperimentAnalyzer` analyzing multiple experiments.
* I'm not sure `YankAutoExperimentAnalyzer` can really be considered a "data model", since it doesn't have an object representation with accessors that provides a one-to-one correspondence container for the data. Instead, it's a utility class that allows you to return some computed analysis data, in dict form. If we really wanted an object data model, we could have a data model where you yield, say, an `ExperimentResult` class that has attributes that represent the result of an experiment, but I think this is OK for now since it will be super convenient to use in present form.
* The unit handling at the `YankAutoExperimentAnalyzer` could potentially be streamlined if we instead optionally ask for all quantities to be converted to a single unit system when creating the object. If not provided, unit-bearing quantities are returned; if provided, everything ends up as dimensionless quantities in the requested units. 
* We might add a few key quantities like temperature and (if defined) pressure to the quantities that are returned for the experiment, since it is important to know these to understand the quantities you are returning. You might also return the unique experiment ID.
* Let's remember to add a bit of info on how to use this feature via the CLI (which we expect to be important!) to the docs: http://getyank.org/latest/analysis.html
* We should also add an example of programmatic analysis somewhere, either to the docs or as docstring examples in the docstrings.

Thanks again!

Lnaden:
@jchodera I have fixed many of the items requested:

* Renamed YankAutoExperimentAnalyzer to ExperimentAnalyzer 
* Stopped calling the ExperimentAnalyzer a "data module"
* Fixed API docs: links and actual rendering of strings on decorated functions
* Added example to the ExperimentAnalyzer in docstring
* Updated the Analysis docs in general (non API)

jchodera:
Looks great! Thanks!

Lnaden:
These tests actually passed. The error is just a conda HTTP error which does not actually halt anything, which is silly, but okay. The actual yank tests all passed. Merging!

PR #1029, opened 2018-07-06
Fix bug in multi-experiment paths

Lnaden:
as it says on the tin

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -848,7 +848,9 @@ class MultiExperimentAnalyzer(object):
     def __init__(self, script, **builder_kwargs):
         self.script = script
         self.builder = ExperimentBuilder(script=script, **builder_kwargs)
-        self.paths = self.builder.get_experiment_directories()
+        # Trap bug where paths ending in the separator cause errors later, ensure they don't end in it.
+        self.paths = tuple(path[:-1] if path[-1] == os.sep else path
+                           for path in self.builder.get_experiment_directories())
 
     def run_all_analysis(self, serialize_data=True, serial_data_path=None, **analyzer_kwargs):
         """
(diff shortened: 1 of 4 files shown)


jchodera:
Looks like there's still a test failure:
```
FAIL: Test that the auto analysis and multi-exp analysis code works
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/test/lib/python3.5/site-packages/yank-0.23.1-py3.5-linux-x86_64.egg/yank/tests/test_analyze.py", line 134, in test_auto_analyze
    assert sorted(exp_directories) == sorted(multi_auto.paths)
AssertionError
```

Yank/analyze.py
```diff
@@ -848,7 +848,9 @@ class MultiExperimentAnalyzer(object):
     def __init__(self, script, **builder_kwargs):
         self.script = script
         self.builder = ExperimentBuilder(script=script, **builder_kwargs)
-        self.paths = self.builder.get_experiment_directories()
+        # Trap bug where paths ending in the separator cause errors later, ensure they don't end in it.
+        self.paths = tuple(path[:-1] if path[-1] == os.sep else path
+                           for path in self.builder.get_experiment_directories())
```
andrrizzi: It may be worth normalizing the paths directly in `ExperimentBuilder.get_experiment_directories()` or maybe in `ExperimentBuilder._get_experiment_dir()` with `os.path.normpath()` instead, so that nobody using that function will stumble into the same problem. That function should always return a path without a trailing `/`.

Lnaden: `normpath` is a way better solution. ading it now

PR #1030, opened 2018-07-07
We don't need to create an nvcc-cache on cray machines

jchodera:
This PR removes some code I had thought was necessary for Cray machines, but turns out not to be.

### DIFF
--- a/Yank/mpi.py
+++ b/Yank/mpi.py
@@ -148,15 +148,6 @@ def get_mpicomm():
     # Report initialization
     logger.debug("MPI initialized on node {}/{}".format(mpicomm.rank+1, mpicomm.size))
 
-    # Cray machines are usually old and sick; prevent them from causing trouble with CUDA caches
-    # by explicitly using different CUDA cache paths for each process
-    if 'ALPS_APP_PE' in os.environ:
-        cuda_cache_path = os.path.abspath(os.path.join('nvcc-cache', str(mpicomm.rank)))
-        if not os.path.exists(cuda_cache_path):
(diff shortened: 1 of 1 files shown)


jchodera:
Test failure is unrelated:
```
FAIL: Test each of the steps of the MBAR creation process
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/test/lib/python3.6/site-packages/yank-0.23.1-py3.6-linux-x86_64.egg/yank/tests/test_analyze.py", line 320, in test_mbar_creation_process
    assert u_sampled.shape == (self.n_replicas, self.n_states, self.n_steps)
AssertionError: 
```
@Lnaden : Any idea what is causing that test failure?

Lnaden:
Huh... no thats a very strange one to see. The only thing I can think of is some corruption in the u_sampled output, but its odd that its showing up here.

Lnaden:
I think that error cropped up at one point when we I was generalizing the `multistate` classes, but its odd to see it here since this is a branch off the current master. I'll pull this down local and see if I can reproduce

Lnaden:
From what I see, it looks like that analysis code had the same problem we sometimes see where the very short test we run fails to sample enough and MBAR cannot converge, so any time we try to access the properties, it does not come back out right. 

I have tried to make this more robust over the versions, I have never found a good solution which does not involve some very long time. It appears to be random, so I have restarted the test.

Issue #947, opened 2018-04-11
Add mcmc moves and samplers section to YAML file created in experiment folders


andrrizzi:
The 'mcmc_moves' and 'samplers' sections are currently ignored.

andrrizzi:
Fixed in #1038 .

Issue #1036, opened 2018-07-13
Planning and ID'ing the CVs we want to give Restraints


Lnaden:
This thread is to track the exact variables we want to isolate as we convert the restraints to the new [CustomCVForce](http://docs.openmm.org/development/api-python/generated/simtk.openmm.openmm.CustomCVForce.html) for easier variables.

The goal is to use each Collective Variable as a tracking for quantities which are hard to extract later. E.g. The distance the 2 centroids are from each other every iteration in `RadiallySymmetricRestraints`

These CV values are fetched from the `Context` through a `CustomCVForce.getCollectiveVariableValues(context)` call.

There will also need to be some engineering changes. We will want to come up with a way to neatly track and report the restraints. The problem is that the `multistate` module only knows of restraints as it pertains to the analysis, and the knowledge of the `CustomCVForce` objects does not go past the `yank.AlchemicalPhase` class. So in order to get the `CustomCVForce.getCollectiveVariableValues` call to be fetched at run time, we will need to engineer a general `CustomCVForce` reader for either the `mulstistate.MultiStateSampler` and `multistate.MultiStateReporter`, and/or we have to engineer the `openmmtools.States` module to handle that and then have the `MultiStateReporter` interface with it. I'd like some feed back here.

As the title says, I'd also like to use this issue to enumerate, explicitly, the CV's we want to create and track for each of the major force classes:
* RadiallySymmetricRestraints
* BoreschLike
* RMSD

These changes will be part of the larger #1015

Lnaden:
Re-posting an idea I shared in Slack for discussion

Idea for extracting CV's: We have to tell the `MultiStateSampler` to cycle through all of its known `thermodynamic_state` which are passed to it in `create`, then tell it to search for states which implement some common CV function.

I think when we create the common parent class for the `RestraintState` (yank) and `AlchemicalState` (oepnmmtools) we would want to add this call to their interface. I think since its a new OpenMM feature others may find non-restraint uses for the CVs and want a common way to extract them. cc #1015

Lnaden:
from @andrrizzi:
Hmm. The collective variable seems to be more relatable to `SamplerState` rather than the thermodynamic state as it changes with integration. It could be something read by `SamplerState.update_from_context()`.

from me: The problem is that the actual call has to come from the `CustomCVForce` object which is tied to the `ThermodynamicState`

@andrrizzi:
[...]

Issue #1015, opened 2018-06-14
Allow running calculations with multiple restraints


andrrizzi:
Since we don't have an analytical correction for the RMSD restraint, we may need to introduce the ability to run with multiple restraints (e.g. RMSD + harmonic) so that we can turn off the rigidifying restraint at the end of the protocol while still reducing the space accessible by the ligand in the decoupled state and being able to bring the ligand to standard concentration.

The only idea I had so far that would make the design very extensible is to have a `CompoundRestraint` class that can be passed as a single restraint.

Controlling multiple restraint at the same time is essentially the same problem of controlling multiple alchemical regions at the same time, so we should consider these two together. For now, the solution that I and @adw62 adopted for alchemical regions is to have each global variable controlled by a different `AlchemicalState`, which is assigned to a particular region so something similar may work also for `RestraintState`.

jchodera:
I think we want to allow multiple restraints for the *same* alchemical region, rather than requiring these to be in separate regions. Could we simply allow each restraint for each region to have a separate label or name?

We could tackle this at the same time as reworking the restraints to use and store `CustomCVForce`s.

andrrizzi:
I agree. I meant that the implementative solutions would be similar, not that we should assign one restraint for each reason. In both cases, we'll have separate variables controlling different parts of the `Alchemica/RestraintState`. Sorry if it was unclear.

Lnaden:
Here are the required code changes I see for the multi-restraints:

* Change the `restraints.py` classes to support different names for the restraint variable, defaulting to `lambda_restraints`
* Change `AlchemicalPhase.create(restraint=...)` keyword to support `None`, a `ReceptorLigandRestraint` or a list of `ReceptorLigandRestraint` (we can expand this to a `LigandRestraint`-ish thing later)
* Apply all the restraints to the alchemical phase

I want to propose the following YAML layout for the multi-`restraints` class:

* Create a new YAML block called `restraint` which supports arbitrary names (like `systems` have names). Keys are just like the current `experiment:restraint:....` block is now. Add new option `lambda_name: X` which takes a string to be cast to `lambda_X`, allows each restraint to be tied to either unique or existing values (including sterics and electrostatics)
* Change `experiment` block to now have a `restraints: R` option where `R` is either a single or list of restraint(s) defined in the `restraints` block. 
    * This would break backwards compatibility and all of our examples, but I don't know of a clean way to continue supporting the old and the new way without having both side-by-side and being confusing. 
[...]

PR #1038, opened 2018-07-16
Add new restraints Application layer interface

Lnaden:
Note: This PR is being merged into the new `multirestraint` branch. Much like the SAMS/MultiStateSampler branch, I wanted to make a branch we can all work on that is not master as we start this process, comment, and review as we would normal PRs. I have not configured travis to run on this branch yet, but my local tests are passing. 

* Added new `restraints` block to the YAML
    * Exactly the same as it used to be in `experiments` block but with unique ID string
    * Added new `name` key which can be used to target custom `lambda_name` variables, optional and defaults to `restraints`. Non-unique to let multiple restraints be controlled by same variable.
    * Block is 100% optional
* `experiments` block allows `restraint: target` to target restraint in `restraints` block.
    * Supports old style to preserve backwards compatibility
    * Supports targeting a single restraint from the `restraints` block
    * Supports a list of targets in the `restraints` block to use multi-restraints
* Updated and added tests for new application layer feature
    * Note: Tests should pass but there is no underlying mechanism to use this yet
* Updated docs
    * New `restraints` doc page
    * Updated `experiments` page with behavior changes to the `restraint` keyword
    * Updated index pages to include new page and pointers.

A key thing: These application layer changes should be 100% compatible with the older versions. So we should not have to have users change their YAML files like we did with the MultiStateSampler conversion. That said, this is still a major new feature and the start of the 0.24.0 release.

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -722,7 +722,9 @@ class ExperimentBuilder(object):
         # Save raw YAML content that will be needed when generating the YAML files
         self._expanded_raw_yaml = copy.deepcopy({key: yaml_content.get(key, {})
                                                  for key in ['options', 'molecules', 'solvents',
-                                                             'systems', 'protocols']})
+                                                             'systems', 'protocols',
+                                                             'samplers', 'mcmc_moves',
+                                                             'restraints']})
 
         # Validate options and overwrite defaults
         self._options.update(self._validate_options(yaml_content.get('options', {}),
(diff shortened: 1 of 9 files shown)


Lnaden:
This is also progress on #1036 and #1015

jchodera:
Thanks for taking a stab at this, @Lnaden!

When extending the YAML syntax, it helps to add some concrete examples instead of just defining the schemas. Could you draft an example YAML file that illustrates the new structure you envision here?

andrrizzi:
Ah! Actually, I think this PR has the same fault as the `samplers` and `mcmc_moves` section, and it doesn't get saved in the generated YAML file (see #947). We can correct it in a separate PR if you prefer.

jchodera:
@lnaden: Repeating this here, since I think it was overlooked:

Could you draft an example YAML file that illustrates the new structure you envision here? 

The PR doesn't include an actual example of what you are envisioning the new YAML syntax to look like for a real project that uses this feature.

andrrizzi:
> That's on purpose. It puts the name right back on it after it calls the constructor.

I meant that when `name` will be an argument of the restraints' constructor, we should remove it. Unless I misremember the plan and `name` won't be an `__init__` parameter.

Lnaden:
`name` will be part of the constructors. We need a CV variable name, and `restraints` is going to be the default.

Lnaden:
Thats also why I am proposing this into the `multirestraint` branch and not master, because I know it will break things because this pr is just not ready yet.

Lnaden:
I have made a couple changes to this part of the multi-restraints.

@jchodera I have added a hard example YAML showing the new format [here](https://github.com/choderalab/yank/blob/31085db63a1df1040692e3d3d78461c92e8753f6/docs/yamlpages/experiments.rst) as part of the docs

@andrrizzi I should have fixed #947 with this, could you please review [the diff](https://github.com/choderalab/yank/pull/1038/files#diff-1957ca1a1cdd9d394cea46a4708af56e) and make sure it looks good to you? This will include the `mcmc_moves`, `samplers` and `restraints` blocks only if they are needed. I have tested them all locally. I also kind of generalized the write-out code in case we need to make some changes to it in the future.

Yank/experiment.py
```diff
@@ -2707,16 +2731,21 @@ def _generate_yaml(self, experiment, file_path, overwrite_protocol=None):
...
+                   ['restraints', restraint_section],
+                   ['protocols', prot_section],
+                   ['experiments', exp_section]
+                   ]
+        for block_id, block in content:
+            if block:
+                yaml_content += yaml.dump({block_id: block}, **dump_options)
 
         # Export YAML into a file
+        import pdb; pdb.set_trace()
```
andrrizzi: Found it :) !

Lnaden:
I always miss the 1. Thank you!

PR #1057, opened 2018-07-31
Adding modeller functionality for mutations

steven-albanese:
Still need to add a modeller test

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ script:
   # Install extras for OpenEye and testing
   - conda install --yes --quiet pip nose nose-timer coverage
   # Install extras for Modeller
-  - conda isntall --yes modeller
+  - conda install --yes modeller
   # Install OpenEye toolkit
   - pip install $OPENEYE_CHANNEL openeye-toolkits && python -c "import openeye; print(openeye.__version__)"
   # Build the recipe
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ install:
   - source devtools/travis-ci/install.sh
   - export PYTHONUNBUFFERED=true
   # Add org channel
-  - conda config --add channels omnia --add channels conda-forge
+  - conda config --add channels omnia --add channels conda-forge --add channels salilab
   # Add omnia dev channels
   - if [ $DEVOMNIA ]; then conda config --add channels omnia/label/dev; fi
   # Update everything
@@ -36,6 +36,8 @@ script:
   - source activate test
   # Install extras for OpenEye and testing
   - conda install --yes --quiet pip nose nose-timer coverage
+  # Install extras for Modeller
+  - conda isntall --yes modeller
   # Install OpenEye toolkit
   - pip install $OPENEYE_CHANNEL openeye-toolkits && python -c "import openeye; print(openeye.__version__)"
   # Build the recipe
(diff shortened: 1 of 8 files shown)


steven-albanese:
Just realized I need to add `modeller` and the license in order for the tests to work

andrrizzi:
> Just realized I need to add modeller and the license in order for the tests to work

Ah, right! I'm not sure if we can do something similar to what we do with the openeye license.

steven-albanese:
From the modeller website: 
```
You will be prompted after installation to edit a file to add your Modeller license key. Alternatively, set the KEY_MODELLER environment variable to your license key before you run 'conda install'.
```

Yank/pipeline.py
```diff
@@ -943,6 +945,111 @@ def add_missing_atoms(value):
...
+            model.build(initialize_xyz=False, build_method='INTERNAL_COORDINATES')
+
+        else:
+            logger.info('pdbfixer: No mutations will be applied since "WT" specified.')
+
+    process_directive('apply_mutations', apply_mutations_modeller, None)
+
+    # Check that there were no extra options
+    if len(directives) > 0:
+        raise ValueError("The 'pdbfixer:' block contained some nodes that it didn't know how to process: {}".format(directives))
```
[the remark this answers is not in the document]
steven-albanese: Thanks for catching this!

steven-albanese:
@jchodera @Lnaden do you have any thoughts on how to handle the license? It's a string of letters that seems to be constant for the last few years, but I'm not sure if we need to protect it somehow for travis

Lnaden:
https://docs.travis-ci.com/user/environment-variables/#defining-encrypted-variables-in-travisyml

steven-albanese:
Ok it looks like the builds are passing now. The only outstanding issue is whether I should check to make sure that no double mutations are being passed

Lnaden:
That is probably worth testing, since our pdbfixer supports double mutants

steven-albanese:
Ok -- I'll raise a `ValueError` if a double mutant gets passed that prints the offending mutant and says it's not supported by modeller right now

steven-albanese:
I think this is ready when someone has a chance to look at it! Let me know if there are any other edits I should make.

.travis.yml
```diff
@@ -20,7 +20,7 @@ install:
   - source devtools/travis-ci/install.sh
   - export PYTHONUNBUFFERED=true
   # Add org channel
-  - conda config --add channels omnia --add channels conda-forge
+  - conda config --add channels omnia --add channels conda-forge --add channels salilab
```
andrrizzi: It looks like the `omnia` channel already has a `modeller` package. Not sure if it's the same.

steven-albanese: I'm not sure, but I *think* they are different

steven-albanese:
Updated the docs and also fixed the default behavior for when `chain_id` isn't selected

docs/yamlpages/molecules.rst
```diff
@@ -239,6 +239,58 @@ PDBFixer is applied after ``strip_protons`` if both are requested.
...
+* ``T315I`` : a single mutation that changes Thr at resid 315 to Ile
+
+``chain_id`` can be either an integer or string. Strings are case sensitive!
+If ``chain_id`` is not specified, it defaults to ``0`` (the first chain in the file will be selected).
+
+Examples for specifying ``chain_id``:
+* ``0`` : the first chain in the file will be selected
+* ``B`` : chain B will be selected
+
+Modeller is applied after both ``strip_protons`` and ''pdbfixer'' if either (or both) are requested.
```
andrrizzi: I think you meant to use \` instead of ' for `pdbfixer` here.

PR #1059, opened 2018-08-02
Convert Restraints to CV-Based

Lnaden:
This PR converts ALL of the restraints to `CustomCVForce` objects.

This feature now requires OpenMMTools 0.16.0, which is  still in development, but the local tests pass. The OpenMMTools PR/branch which I am using is here:

choderalab/openmmtools#363

This changes many things so could use a review. Many of the changes here are reflections of the linked OpenMMTools issues

### DIFF
--- a/Yank/restraints.py
+++ b/Yank/restraints.py
@@ -1011,7 +1011,7 @@ class RadiallySymmetricRestraint(ReceptorLigandRestraint):
 # Harmonic protein-ligand restraint.
 # ==============================================================================
 
-class Harmonic(RadiallySymmetricRestraint):
+class Harmonic(OpenMM73, RadiallySymmetricRestraint):
     """Impose a single harmonic restraint between ligand and protein.
 
     This can be used to prevent the ligand from drifting too far from the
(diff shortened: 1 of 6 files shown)


Yank/tests/test_restraints.py
```diff
@@ -290,7 +290,8 @@ def test_partial_parametrization():
                 particles, _ = force.getBondParameters(0)
                 assert particles == tuple(restraint.restrained_receptor_atoms + restraint.restrained_ligand_atoms)
             # RMSD restraint.
-            elif OpenMM73.dev_validate and isinstance(force, openmm.CustomCVForce):
+            elif OpenMM73.dev_validate and isinstance(force, openmm.CustomCVForce) and \
```
andrrizzi: If remember correctly, the `OpenMM73.dev_validate` we added some time ago is unnecessary now since the CV force is openmm>=7.2 and we've pinned it in the conda recipe, correct?

Yank/restraints.py
```diff
@@ -1635,12 +1466,51 @@ def restrain_state(self, thermodynamic_state):
...
+                                   self.restrained_ligand_atoms[0],
+                                   [])
+        force_components.append(['boresch_torsion_a', torsion_a_force])
+        torsion_b_force = openmm.CustomTorsionForce('theta')
+        torsion_b_force.addTorsion(self.restrained_receptor_atoms[1],
+                                   self.restrained_receptor_atoms[2],
+                                   self.restrained_ligand_atoms[0],
+                                   self.restrained_ligand_atoms[1],
+                                   [])
+        force_components.append(['TorsionB', torsion_b_force])
```
andrrizzi: I think you forgot to change `TorsionB` to `boresch_torsion_b`.

Yank/restraints.py
```diff
@@ -1635,12 +1466,51 @@ def restrain_state(self, thermodynamic_state):
...
+        torsion_c_force = openmm.CustomTorsionForce('theta')
+        torsion_c_force.addTorsion(self.restrained_receptor_atoms[2],
+                                   self.restrained_ligand_atoms[0],
+                                   self.restrained_ligand_atoms[1],
+                                   self.restrained_ligand_atoms[2],
+                                   [])
+        force_components.append(['TorsionC', torsion_c_force])
```
andrrizzi: Same here.

Yank/tests/test_restraints.py
Lnaden: RMSD Force is still a 7.3 thing, but I need to fix that to be more generic

Yank/restraints.py
Lnaden: Ah, I did, that might explain the NaN's

Yank/tests/test_restraints.py
andrrizzi: Ah! I see, I forgot about the RMSD restraint.

PR #1062, opened 2018-08-03
Implementation of Multi-Restraint

Lnaden:
This PR completes the Multi-restraint implementation, including the ability to control the restraints through different variables.. There are some outstanding things still.

* What is the correct way to handle the Standard State Correction for multiple restraints? It won't be additive and depends on the restrained atoms and restraints themselves.
* Making the analysis actually use the CustomCVForce features is missing. But that can be implemented separate from this PR
* I'm not sure what the best way to test this feature in a way that does not require long simulation (and I don't know what the validation would be, see the first point)

### DIFF
--- a/Yank/restraints.py
+++ b/Yank/restraints.py
@@ -333,7 +333,7 @@ class ReceptorLigandRestraint(ABC):
         return self._restraint_name
 
     @property
-    def full_restraint_name(self):
+    def controlling_parameter_name(self):
         """
         Provide the full global parameter for this restraint
 
@@ -1011,11 +1011,11 @@ class Harmonic(RadiallySymmetricRestraint):
             return mmtools.forces.HarmonicRestraintBondForce(spring_constant=self.spring_constant,
                                                              restrained_atom_index1=particles1[0],
                                                              restrained_atom_index2=particles2[0],
-                                                             controlling_parameter_name=self.full_restraint_name)
+                                                             controlling_parameter_name=self.controlling_parameter_name)
         return mmtools.forces.HarmonicRestraintForce(spring_constant=self.spring_constant,
                                                      restrained_atom_indices1=particles1,
                                                      restrained_atom_indices2=particles2,
-                                                     controlling_parameter_name=self.full_restraint_name)
+                                                     controlling_parameter_name=self.controlling_parameter_name)
 
     def _determine_restraint_parameters(self, thermodynamic_state, sampler_state, topography):
         """Automatically choose a spring constant for the restraint force.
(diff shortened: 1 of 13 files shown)


Yank/restraints.py
```diff
@@ -313,8 +320,31 @@ class ReceptorLigandRestraint(ABC):
...
+    def __init__(self, restraint_name=None):
+        self._restraint_name = restraint_name
+
+    @property
+    def restraint_name(self):
+        """Provide the restraint_name suffix"""
+        return self._restraint_name
+
+    @property
+    def full_restraint_name(self):
```
andrrizzi: I would call this something like `controlling_parameter_name` to make the connection clear to the `openmmtools` stuff and because technically I wouldn't say that `lambda_restraints_myrestraintname` is a name for the restraint.

Yank/restraints.py
```diff
@@ -2576,12 +2656,13 @@ def restrain_state(self, thermodynamic_state):
...
         # Create an CustomCVForce
-        energy_expression = 'lambda_restraints * step(dRMSD) * (K_RMSD/2)*dRMSD^2; dRMSD = (RMSD-RMSD0);'
+        energy_expression = ('{} * step(dRMSD) * (K_RMSD/2)*dRMSD^2; '
+                             'dRMSD = (RMSD-RMSD0);'.format(self.full_restraint_name))
         energy_expression += 'K_RMSD = %f;' % self.K_RMSD.value_in_unit_system(unit.md_unit_system)
         energy_expression += 'RMSD0 = %f;' % self.RMSD0.value_in_unit_system(unit.md_unit_system)
         restraint_force = openmm.CustomCVForce(energy_expression)
         restraint_force.addCollectiveVariable('RMSD', rmsd_cv)
-        restraint_force.addGlobalParameter('lambda_restraints', 1.0)
+        restraint_force.addGlobalParameter('{}'.format(self.full_restraint_name), 1.0)
```
andrrizzi: Isn't `'{}'.format(self.full_restraint_name)` equivalent to just `self.full_restraint_name`?

Yank/tests/test_restraints.py
```diff
@@ -144,21 +165,28 @@ def general_restraint_run(options):
...
         Deltaf_ij, dDeltaf_ij = analyzer.get_free_energy()
         # Correct the sign for the fact that we are adding vs removing the restraints
         DeltaF_simulated = Deltaf_ij[-1, 0]
         dDeltaF_simulated = dDeltaf_ij[-1, 0]
-        DeltaF_restraints = ncfile.groups['metadata'].variables['standard_state_correction'][0]
+        DeltaF_restraints = reporter.read_dict('metadata')['standard_state_correction']
+        # DeltaF_restraints = ncfile.groups['metadata'].variables['standard_state_correction'][0]
         ncfile.close()
-
+        import pdb; pdb.set_trace()
```
andrrizzi: Gotcha!

Yank/tests/test_restraints.py
```diff
@@ -95,43 +95,64 @@ def build_test_case():
...
+    restraint_name: waffle
 
 protocols:
   absolute-binding:
     complex:
       alchemical_path:
         lambda_restraints:     [0.0, 0.25, 0.5, 0.75, 1.0]
         lambda_electrostatics: [0.0, 0.00, 0.0, 0.00, 0.0]
-        lambda_sterics:        [0.0, 0.00, 0.0, 0.00, 0.0]
+        lambda_sterics:        [0.0, 0.00, 0.0, 0.00, 0.0]{waffle_augment}
```
andrrizzi: lol

PR #1063, opened 2018-08-07
Move CLI imports into functions for speed

Lnaden:
In order to prevent the CLI from taking long times due to importing everything, especially when the user mis-types a command for docopt to handle, import statements for the CLI functions have been moved to the actual functions instead of the module header


jchodera:
Awesome! Thanks!

PR #1065, opened 2018-08-10
Clean up Modeller Feature, Cut 0.23.3

Lnaden:
* Added more robust install check
* Getting Modeller from Omnia channel now
* Improved the docs with more info + better rendering
* Releases 0.23.3


Lnaden:
I think if people are looking for the Modeller enhancement, they will know where to look

Lnaden:
But it should be easy to add

steven-albanese:
I think you're right -- no need to add it.

PR #1074, opened 2018-08-23
FIxing modeller bug for WT case

steven-albanese:
Modeller was throwing an error when given `WT` in the yaml pipeline. 

* fixing the way this is being handled 
* Adding a test for the WT case

### DIFF
--- a/Yank/pipeline.py
+++ b/Yank/pipeline.py
@@ -1021,6 +1021,7 @@ def apply_modeller(input_file_path, output_file_path, directives):
 
         else:
             logger.info('modeller: No mutations will be applied since "WT" specified.')
+            alignment.append_model(model, align_codes='WT')
 
     process_tool_directive(directives, 'apply_mutations', apply_mutations_modeller, None)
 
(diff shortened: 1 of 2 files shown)


steven-albanese:
Tests are passing, but I'm not sure if I've got the test for the WT case correct

andrrizzi:
Whops! Sorry, it looks like the test I suggested fails because when using `WT` with modeller it doesn't save the pdb file. If that's expected, feel free to substitute
```python
setup_molecule_output_check(exp_builder._db, mol_id, output_path)
os.remove(output_path)  # Remove file for next check.
```
with
```python
exp_builder._db._setup_molecules(mol_id)
assert not os.path.exists(output_path)
```
Can you also update the release history (docs/whatsnew.rst) before merging?

steven-albanese:
Hmm weirdly, it looks like it's actually the second `setupt_molecule_output_check` that is failing: https://github.com/choderalab/yank/blob/bd9a4932d9c704dfa914b32f2373a6c8af82f69b/Yank/tests/test_experiment.py#L1356

Do I need to reinitialize the `ExperimentBuilder` object in between the wild type and mutant case?

andrrizzi:
Hmm. I don't think you should! Can you check what `exp_builder._db.is_molecule_setup(mol_id)` returns after the call to `os.remove(output_path)`?

andrrizzi:
Tests finally pass! Should I merge this and go ahead with a bugfix release?

steven-albanese:
Yep!

PR #1084, opened 2018-09-10
MPI exceptions logging and more robust NetCDF closures

andrrizzi:
I've made a few small modifications to fix the stack trace that is logged when an exception is raised in an MPI process and to make more robust the NetCDF handling. In particular,

- I've implemented `AlchemicalPhase.__del__` and make `yank.experiments.Experiment` force garbage collection after deleting the alchemical phase.
- I've minimized the number of `Dataset.open/close()` in the multi-state samplers.
- The reporter now attempts to open the NetCDF file 5 times before giving up and raising the error.

The `switch_phase_interval` is working much better for me after these modifications, which is hopeful.

### DIFF
--- a/Yank/multistate/multistatereporter.py
+++ b/Yank/multistate/multistatereporter.py
@@ -120,8 +120,8 @@ class MultiStateReporter(object):
             logger.debug("Initial checkpoint file automatically chosen as {}".format(checkpoint_storage))
         else:
             checkpoint_storage = os.path.join(dirname, checkpoint_storage)
-        self._storage_file_analysis = storage
-        self._storage_file_checkpoint = checkpoint_storage
+        self._storage_analysis_file_path = storage
+        self._storage_checkpoint_file_path = checkpoint_storage
         self._storage_checkpoint = None
         self._storage_analysis = None
         self._checkpoint_interval = checkpoint_interval
@@ -137,7 +137,7 @@ class MultiStateReporter(object):
 
         Classes outside the Reporter can access the file string for error messages and such.
         """
-        return self._storage_file_analysis
+        return self._storage_analysis_file_path
 
     @property
     def _storage(self):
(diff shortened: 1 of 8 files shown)


andrrizzi:
This should be ready to be merged!

Yank/multistate/multistatereporter.py
```diff
@@ -253,102 +253,69 @@ def open(self, mode='r', convention='ReplicaExchange', netcdf_format='NETCDF4'):
...
+                storage_dir = os.path.normpath(os.path.dirname(storage_path))
+                os.makedirs(storage_dir, exist_ok=True)
+
+        # Analysis file.
+        # ---------------
+
+        # Open analysis file.
+        self._storage_analysis = self._open_dataset_robustly(self._storage_analysis_file_path,
+                                                             mode, version=netcdf_format)
+        # TODO - AR: What's the purpose of set_auto_mask(False)? When we create the
```
jchodera: We no longer need it since we're not using masking!

Yank/multistate/sams.py
```diff
@@ -373,13 +373,14 @@ def _mix_replicas(self):
         # TODO: We may be able to refactor this to simply have different update schemes compute neighborhoods differently.
         # TODO: Can we allow "plugin" addition of new update schemes that can be registered externally?
         with mmtools.utils.time_it('Mixing of replicas'):
-            jump_and_mix = self._JumpAndMixPacket(self.n_replicas, self.n_states)
+            # Initialize statistics. This matrix is modified by the jump function and used when updating the logZ estimates.
+            replicas_log_P_k = np.zeros([self.n_replicas, self.n_states], np.float64)
```
jchodera: This is much clearer now---thanks!

Yank/multistate/sams.py
```diff
@@ -433,35 +434,37 @@ def _local_jump(self, jump_and_mix_data):
...
-            # Compute unnormalized log probabilities for all thermodynamic states
+        for replica_index, current_state_index in enumerate(self._replica_thermodynamic_states):
             neighborhood = self._neighborhood(current_state_index)
+
+            # Compute unnormalized log probabilities for all thermodynamic states.
+            log_P_k = np.zeros([n_states], np.float64)
             for state_index in neighborhood:
-                u_k[state_index] = self._energy_thermodynamic_states[replica_index, state_index]
-                log_P_k[state_index] = self.log_weights[state_index] - u_k[state_index]
+                u_k = self._energy_thermodynamic_states[replica_index, state_index]
```
jchodera: To match the equations, it would be clearer to write:
```python
u_k = self._energy_thermodynamic_states[replica_index, :]
log_P_k[state_index] = - u_k[state_index] + self.log_weights[state_index] 
```

andrrizzi: Good idea! Will change.

Issue #1026, opened 2018-07-02
More robust counterion identification


jchodera:
Currently, we use this code to detect ion residues:
https://github.com/choderalab/yank/blob/9bd78df1897b897ac7287b4cdfab55f84b3761d5/Yank/yank.py#L185-L191

It looks for ion residue names that contain `+` or `-`: e.g. `Na+` and `Cl-`. 

If your residue is instead named `NA` or `CL`, it will fail to identify the ion.

It would be useful if we can make this more robust.

PR #1086, opened 2018-09-11
added 'NA' and 'CL' to set {'+', '-'} identifying ions

kmboehm:
Addressing issue  #1026

### DIFF
--- a/Yank/yank.py
+++ b/Yank/yank.py
@@ -188,7 +188,9 @@ class Topography(object):
         # Ions are all atoms of the solvent whose residue name show a charge.
         return [i for i in self._solvent_atoms
                 if '-' in self._topology.atom(i).residue.name or
-                '+' in self._topology.atom(i).residue.name]
+                '+' in self._topology.atom(i).residue.name or
+		'NA' in self._topology.atom(i).residue.name or
+		'CL' in self._topology.atom(i).residue.name]
 
     def add_region(self, region_name, region_selection, subset=None):
         """


Yank/yank.py
```diff
@@ -188,7 +188,9 @@ def ions_atoms(self):
         # Ions are all atoms of the solvent whose residue name show a charge.
         return [i for i in self._solvent_atoms
                 if '-' in self._topology.atom(i).residue.name or
-                '+' in self._topology.atom(i).residue.name]
+                '+' in self._topology.atom(i).residue.name or
+		'NA' in self._topology.atom(i).residue.name or
+		'CL' in self._topology.atom(i).residue.name]
```
andrrizzi: Thanks for the PR! A couple of suggestions:

1. It looks like you've configured your IDE to use tabs instead of spaces. We shouldn't mix the two types of indentation in Python. Can you switch to 4-space tabs the settings?
2. Using the `in` operator here risks to collect things we don't want. What about having maintaining a set of residue names we want to consider ions (beside the ones containing + and -) and have something like
```python
ION_RESIDUE_NAMES = {'NA', 'CL'}
return [i for i in self._solvent_atoms
    if '-' in self._topology.atom(i).residue.name or
    '+' in self._topology.atom(i).residue.name or
    self._topology.atom(i).residue.name in ION_RESIDUE_NAMES]
```
This way we check for equality instead of looking in the substrings.

kmboehm:
Thanks, guys!

Issue #1096, opened 2018-09-26
The latest YANK release doesn't have API docs


andrrizzi:
It appears that the API documentation is not pushed correctly without the `api` folder that was removed in #1092.

jchodera:
That's odd, since the api docs should be automatically generated.

PR #1092, opened 2018-09-22
Fix restraints and restraint tests

jchodera:
Fixes #1089

This PR fixes the incorrect standard state correction for Boresch-like restraints (#1089) and restraint test failures.

- [x] Get broken restraint unit tests running again
- [x] Update YAML to modern best practices
- [ ] See if we can get this to run on travis - **TOO SLOW**
- [x] Fix any detected issues

The issue was caused because `kT` was not converted to a unitless quantity before using it in in computing the integrand for quadrature in [this commit](https://github.com/choderalab/yank/commit/11570959a82effa9d2172b8616a493c1e18bddab).

More generally, we should probably *never* try to convert unit-bearing quantities to floating point quantities in a consistent unit system for computing integrands. We should always introduce a dimensionless value `x` that runs between 0 and 1 and multiply the integration range by this quantity. I haven't made that change here to minimize impact on other things.

I've also corrected #1091 and fixed a typo in the `mcmc` YAML docs that confused me while trying to update the YAML files.

PR #1097, opened 2018-09-26
Restore erroneously deleted API docs

jchodera:
This PR restores the API docs erroneously deleted in #1092.


jchodera:
Fixes #1096

andrrizzi:
Thank you!

PR #1099, opened 2018-09-27
Add YAML syntax tutorial and fix MC atoms

### LINKED ISSUE #1098: ExperimentBuilder should configure by default MCDisplacement/Rotation atom_subset
I've just realized that the `mcmc_moves` block doesn't treat specially the constructor argument `atom_subset` of `MCDisplacementMove` and `MCRotationMove`. This means that when the YAML specification doesn't contain a list of atom indices, the default is used (`None`), and the move tries to perform a rigid rotation/translation of the whole system instead of the ligand.

In the future, we'll want to integrate this with the region system, but for now, I'll implement a fix to use the ligand atoms when `atom_subset` is not given.

andrrizzi:
- Add a tutorial that cover the basics of the YAML syntax.
- Fix #1098: When redefining `MCDisplacement/RotationMove` in the `mcmc_moves` without specifying `atom_subset` caused the MC move to affect the whole system instead of only the ligand atoms.
- Add `net_charge` option to the `antechamber` pipeline for charged small molecules.

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -2897,7 +2897,8 @@ class ExperimentBuilder(object):
         if mcmc_move_id is None:
             mcmc_move = self._create_default_mcmc_move(experiment_description, default_mc_atoms)
         else:
-            mcmc_move = schema.call_mcmc_move_constructor(self._mcmc_moves[mcmc_move_id])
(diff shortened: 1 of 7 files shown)


jchodera:
Looks great, and thanks for adding the tutorial!

PR #1113, opened 2018-10-20
SAMS improvements

### LINKED ISSUE #1112: SAMS issues
I've finally managed to generate a sufficient amount of data for c-Met to start to debug what is going on with SAMS convergence.

It looks like, while the simulation would quickly switch into the `asymptotically-optimal` state after a few thousand iterations, this state would be lost as soon as the simulation switches phase or resumes. This can be seen in this `logZ` vs `iteration` plot where, after `switch_interval = 5000` iterations, the asymptotically optimal state is lost, `logZ[0]` is no longer pinned to zero, and large weight fluctuations cause the simulation to not converge.
image
While the range of the free energies for c-Met are close, we currently believe the fluctuations introduced into the log weights violates our assumption that the `(x_n, s_n)` pairs are conditionally distributed from `p(x_n | s_n)`:
image
Once we properly have the log weights decaying asymptotically, I'll repeat these tests, analyzing only the data where the log weights have approximately converged.

I'm working on a PR to fix this.

jchodera:
Fixes #1112 

* SAMS now properly restores `t0` and `stage`, which should correctly persist the stage across phase switches and resumes
* Added SAMS `logZ_history` plot to notebook reports
* SAMS `self._stage` is now an integer to more easily store this in storage (0 is initial, 1 is asymptotic phase)
* Suppress some verbose outputs
* If an asymptotically optimal stage is present in a SAMS simulation, the initial stage is discarded to equilibration and automatic equilibration detection is applied to the asymptotically optimal weight adjustment region only.
* Jupyter notebook reports now show SAMS weight convergence if present

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -409,21 +409,37 @@ class ExperimentAnalyzer(object):
             for i, phase_name in enumerate(self.phase_names):
                 serial = {}
                 analyzer = self.analyzers[phase_name]
+
+
                 # Data crunching to get timeseries
                 # TODO: Figure out how not to discard the first sample
                 # Sample at index 0 is actually the minimized structure and NOT from the equilibrium distribution
                 # This throws off all of the equilibrium data
-                self._n_discarded = discard_from_start
+                t0 = discard_from_start
+                self._n_discarded = t0
                 series = analyzer.get_effective_energy_timeseries()
-                if series.size <= discard_from_start:
+
+                # Update discard_from_start to match t0 if present
+                try:
+                    iteration = len(series)
+                    data = analyzer.reporter.read_online_analysis_data(iteration, 't0')
+                    t0 = max(t0, int(data['t0'][0]))
+                    logger.debug('t0 found; using initial t0 = {} instead of 1'.format(t0))
+                    self._n_discarded = t0
+                except Exception as e:
+                    # No t0 found
+                    pass
+
+                if series.size <= t0:
                     # Trap case where user has dropped their whole set.
                     # Rare, but happens, often with debugging
-                    discard_from_start = 0
+                    t0 = 0
                     logger.warning("Alert: analyzed timeseries has the same or fewer number of values as "
                                    "discard_from_start! The whole series has been preserved to ensure there is "
                                    "*something* to analyze.")
                     self._n_discarded = 0
-                self.u_ns[phase_name] = analyzer.get_effective_energy_timeseries()[discard_from_start:]
+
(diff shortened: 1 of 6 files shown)


jchodera:
Tests pass locally and on travis. 
I'm just testing on lilac now.

jchodera:
It appeared that resuming still did not properly restore the stage and `t0`.  I found (and hopefully fixed) the issue, but I'm not sure how/where in the testing infrastructure I can add a test that properties like this are correctly restored from the reporter. Any suggestions?

jchodera:
This should be ready for preliminary review!

andrrizzi:
These classes are also in a branch in `openmmtools` now so we need to remember to port these fixes there too.

jchodera:
> A new test in the TestSingleReplicaSAMS suite is probably the best place.

Is there currently a test that examines this? The complexity of the storage layer has grown so large that I am not actually sure how to implement tests like this given the deep call graphs.

jchodera:
> If reassign_velocities is False the MCMC moves don't touch the velocities, and we don't save them in the netcdf file so they may be left to zero by OpenMM when the Context is first created and stepped.

This must be what is happening now, and I think this is a major problem. How can we fix this?

Should YANK assign new velocities when the samplers are restored?

jchodera:
I'm experiencing some new energy artifacts at the frequency of `switch_experiment_interval`:
image
The energies dip significantly and then recover. Could the MCMC sampler be resetting the velocity to zero on resume instead of the Maxwell-Boltzmann distribution? I am using the sampler block:
```yaml
mcmc_moves:
    langevin:
        type: MCDisplacementMove # Monte Carlo ligand displacement
        type: MCRotationMove # Monte Carlo ligand rotation
        type: LangevinSplittingDynamicsMove
        timestep: 4.0*femtosecond
        collision_rate: 1.0/picosecond
        reassign_velocities: no
        splitting: 'V R O R V'
        n_steps: 500
        n_restart_attempts: 4
```

andrrizzi:
> How can we fix this? Should YANK assign new velocities when the samplers are restored?

We can either have the `MultiStateSampler` generate velocities the first time it calls `_propagate_replica` if it detects that `SamplerState.velocities is None`, or we could have the `IntegratorMove` do it even if `reassign_velocities is False` when it detects that velocities are 0.0 in the `Context`. The second one is more general, but it would add a somewhat expensive check at the beginning of each propagation.

jchodera:
I like the first idea!

jchodera:
Merging this now since tests pass, these are clear improvements, and future work can be carried out elsewhere.

PR #1148, opened 2019-02-25
Remove "sudo" keyword from Travis configuration

jugmac00:
... as its usage is deprecated, cf
https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,4 @@
 language: c
-sudo: false
 addons:
   apt:
     sources:


jchodera:
Thanks! More generally, we hope to update to use the [MolSSI CMS cookiecutter](https://github.com/MolSSI/cookiecutter-cms) repository structure here, so we might just migrate fully to that.

@andrrizzi : What do you think?

PR #1159, opened 2019-05-15
Add Checkpointing of Trailblaze Algorithm

SimonBoothroyd:
## Description
This PR aims to implement a basic checkpointing system for the trailblazing algorithm - currently if the algorithm is interrupted while running (e.g. if an insufficient wall clock limit is set by a job submission script), the partially computed optimized lambda states are lost.

## Status
- [x] Ready to review
- [ ] Ready to go

### DIFF
--- a/Yank/pipeline.py
+++ b/Yank/pipeline.py
@@ -1990,7 +1990,8 @@ def trailblaze_alchemical_protocol(thermodynamic_state, sampler_state, mcmc_move
         ``lambda_sterics``) and a list specifying the initial and final
         values for the path.
     checkpoint_path : str
-        The path to the experiment directory.
+        The path to the trailblaze checkpoint file. In none exisits,
+        one will be created.
     std_energy_threshold : float
         The threshold that determines how to separate the states between
         each others.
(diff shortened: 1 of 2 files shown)


Yank/pipeline.py
```diff
@@ -1962,7 +1963,7 @@ def _get_number_box_waters(self, *args):
 # ==============================================================================
 
 def trailblaze_alchemical_protocol(thermodynamic_state, sampler_state, mcmc_move, state_parameters,
-                                   std_energy_threshold=0.5, threshold_tolerance=0.05,
+                                   checkpoint_path, std_energy_threshold=0.5, threshold_tolerance=0.05,
```
jchodera: Do we want to make checkpointing optional, or required via the Python API? This approach seems to make it required, but it seems like it would be simple to make it optional.

andrrizzi: I agree it would be useful to have it optional in the function (although exposing it in the YAML will be slightly more complicated). It's possible test are not passing also because you're defining a positional argument after a keyword one.

andrrizzi:
I'm going to merge these changes so that I can pick up from here and store also the configurations generated by trailblaze in a separate PR.

andrrizzi:
Ah! I didn't notice the test failures. I'm going to make the changes directly here then.

andrrizzi:
The failing tests have been already fixed in #1168 so I think we can merge safely. I'll correct eventual issues in the next PR if there are any.

PR #1162, opened 2019-05-20
Importing logsumexp from scipy.special in Multistateanalyzer

dprada:
There was a `logsumexp` left to be fixed: in the file multistateanalyzer.
This was probably the source of the errors reported by Travis CI.


jchodera:
Thanks for the fix, @dprada! We migrated YANK to use the new [`openmmtools.multistate`](https://openmmtools.readthedocs.io/en/0.18.1/multistate.html) functionality, but haven't yet removed those files from the YANK repo, so it looks like we need to fix it in http://github.com/choderalab/openmmtools instead

Issue #1166, opened 2019-05-29
Restraint Energy Computation Expects Host-Guest at Start of System


SimonBoothroyd:
## Description
As discussed with @andrrizzi - It seems like Yank currently expects any host-guest molecules to be the first molecules in a system (i.e before any solvent and ions). Placing these at the end yields an index out of range exception in the `MultiStateSamplerAnalyzer._compute_restraint_energies` method.

## Relevant Stack Trace
```
"/home/xxx/lib/python3.7/site-packages/yank/multistate/multistateanalyzer.py", line 1543, in _compute_mbar_unbiased_energies
    weights_group2)
  File "/home/xxx/lib/python3.7/site-packages/yank/multistate/multistateanalyzer.py", line 1729, in _compute_restraint_energies
    sliced_sampler_state = sampler_state[original_restrained_atom_indices]
  File "/home/xxx/lib/python3.7/site-packages/openmmtools/states.py", line 2114, in __getitem__
    sampler_state._set_positions(copy.deepcopy(self._positions[item]),
  File "/home/xxx/lib/python3.7/site-packages/openmmtools/utils.py", line 352, in __getitem__
    return super(TrackedQuantity, self).__getitem__(item)
  File "/home/xxx/lib/python3.7/site-packages/simtk/unit/quantity.py", line 733, in __getitem__
    assert not is_quantity(self._value[key])
IndexError: index 3072 is out of bounds for axis 0 with size 144
"]
```

PR #1167, opened 2019-05-29
Fix Restrained Atom Indices

SimonBoothroyd:
Hopeful fix for #1166

### DIFF
--- a/Yank/multistate/multistateanalyzer.py
+++ b/Yank/multistate/multistateanalyzer.py
@@ -1669,6 +1669,18 @@ class MultiStateSamplerAnalyzer(PhaseAnalyzer):
         restraint_force = copy.deepcopy(restraint_force)
         is_periodic = restraint_force.usesPeriodicBoundaryConditions()
 
+        # Store the indices of the restrained atoms in the reduced system.
+        analysis_indices = self._reporter.analysis_particle_indices
+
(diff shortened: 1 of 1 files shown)


Yank/multistate/multistateanalyzer.py
```diff
@@ -1726,7 +1727,7 @@ def extract_decorrelated(cached_dict, dtype, unit):
                                                                 analysis_particles_only=True)
 
             for replica_idx, sampler_state in enumerate(sampler_states):
-                sliced_sampler_state = sampler_state[original_restrained_atom_indices]
+                sliced_sampler_state = sampler_state[reduced_restrained_atom_indices]
```
andrrizzi: I believe this won't work with the current implementation. Assume your solute-only sampler states covers particles [1000-2000], and in the simulation you're restraining [1500-1600] with [1900-2000]. I think this will re-compute the restraint distances and energies based on the positions of atoms [1000-1200].

It might be worth writing a regression test for this if you manage to encapsulate the interesting bit into a separate static function.

SimonBoothroyd: I'll look into this when I get a chance - thanks!

andrrizzi:
Whoops! I forgot to merge this, apologies. Closing this as the `multistate` subpackage was moved to OpenMMTools. I've ported the fix in choderalab/openmmtools#462.

PR #1175, opened 2019-07-17
Remove the deprecated multistate and mpi packages

andrrizzi:
In this PR, I have removed `yank.multistate.*` and `yank.mpi` that have been deprecated and moved to `openmmtools` and `mpiplus` respectively.


andrrizzi:
It looks like there's a small import error in `openmmtools` so this will have to wait for a bugfix release before tests can pass. This can wait anyway. We should also move `test_analyze.py` to `openmmtools`.

jaimergp:
> -10,816 

That's lovely (L)

jchodera:
Hooray!

andrrizzi:
Tests are passing now with the new `openmmtools` release.

Issue #1139, opened 2019-01-29
Expose more options for the trailblaze algorithm in YAML


andrrizzi:
Specifically, we should be able to:
- Start the protocol with a harmonic/flat-bottom restraint turned off;
- Control the threshold used to separate the states;
- Allow the configuration used at the end of the equilibration phase to be used as the new starting configuration for the free energy calculations.

PR #1176, opened 2019-07-23
Trailblaze: turn off harmonic restraints, resume, and use samples to seed repex

### LINKED ISSUE #968: Allow resuming the trailblaze algorithm
We'll have to store the current protocol, restrained system, current configuration, and previous energies.

### LINKED ISSUE #1137: Change the trailblaze alchemical path to start with a harmonic/flat-bottom restraint turned off
This is to avoid recomputing the restraint radius at analysis by default, which is very expensive.

andrrizzi:
## Description

- Fix #968: Allow resuming from the trailblaze checkpoint. This finishes up #1159 to recover also the last positions and box vectors when trailblaze is resumed after being interrupted. The frames are saved in a `dcd` file right now, although they are accessed with a utility function that is format-agnostic so we can change it later if this gets complicated.
- Use the samples collected during trailblaze to seed HREX replicas.
- Fix #1137: Change the default trailblaze path to start with turning off the harmonic/flatbottom restraint instead of keeping it always on and reweighting at analysis time. We will be able to expose it as an option once #1139 is implemented.
- Do not restrain receptor atoms right before starting trailblazing. The option is in the API, but we should expose this as well with #1139.
- Remove a lot of warnings due to `yaml.load(loader=...)` not supporting a default value for the `loader` argument.

## To do
- [x] Update docs and changelog.

## Status
- [x] Ready to go


Yank/experiment.py
```diff
@@ -2511,61 +2518,78 @@ def __getattr__(self, _):
...
+            # If default argument is used, determine number of equilibration
+            # iterations to relax the system before restraining the receptor
+            # atoms and start trailblazing. We don't need to do any minimization
+            # or equilibration if we're resuming.
+            trailblaze_dir_path = self._get_trailblaze_checkpoint_dir_path(
+                experiment_path, phase_name)
+
+            if not os.path.isfile(os.path.join(trailblaze_dir_path, 'protocol.yaml')):
+                # TODO automatic equilibration?
+                if n_equilibration_iterations is None:
```
jchodera: Do we document these defaults somewhere in the documentation?
I worry a bit that the nontrivial nature of these defaults might be accidentally missed by a user.
We could alternatively require that the user specify these if they `n_equilibration_iterations > 0` to simplify things.

docs/yamlpages/protocols.rst
```diff
@@ -114,14 +114,20 @@ Automatic Path Determination
          alchemical_path: auto
 
 
-YANK can automatically determine the parameters for ``lambda_X`` for you. YANK will distribute states between the fully
+YANK can automatically determine the sequence of values for the ``lambda_X`` values using an algorithm we called
+"thermodynamic trailblazing". YANK will distribute states between the fully
```
jchodera: We should also document the user-specifiable settings that can affect how trailblazing is conducted, like the number of equilibration steps, the number of samples/steps per energy recorded, the step size, etc.

Yank/experiment.py
andrrizzi: My plan is to implement let users control these options soon (in a separate PR for #1139), including the number of equilibration iterations, so that this won't be a problem.

docs/yamlpages/protocols.rst
andrrizzi: I agree. I was thinking of adding docs for the options when I'll expose them (see previous comment).

PR #1189, opened 2019-11-05
Remove ambermini and redistribution of the states in trailblaze

andrrizzi:
I'm testing switching from `ambermini` to `ambertools` here, and I've added an option to redistribute the states after trailblazing based on the std of the potential difference computed in both directions.

I'll have to merge and release choderalab/openmmtools#444 before merging for tests to pass, but they do pass locally.

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -1502,10 +1502,10 @@ class ExperimentBuilder(object):
                 n_samples_per_state:
                     type: integer
                     default: 100
-                std_potential_threshold:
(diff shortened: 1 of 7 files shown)


andrrizzi:
Actually, I think we can get rid of the amber direct dependency completely. We're using antechamber only through `openmoltools`, which is bringing in its own `ambermini` dependency so I think that specifying our own just increases the chance of conflicts.

andrrizzi:
See also choderalab/openmoltools#278 .

PR #1195, opened 2019-11-10
Fix bug causing crash when resuming redistributed protocol

andrrizzi:
I found another small bug that causes YANK to crash when trailblaze is interrupted and then resumed with the option `bidirectional_redistribution` set.

I'm going to make another bugfix release after merging this.

### DIFF
--- a/Yank/pipeline.py
+++ b/Yank/pipeline.py
@@ -2000,13 +2000,17 @@ class _DCDTrajectoryFile(mdtraj.formats.dcd.DCDTrajectoryFile):
                       (a, b, c), (alpha, beta, gamma))
 
 
-def read_trailblaze_checkpoint_coordinates(checkpoint_dir_path):
+def read_trailblaze_checkpoint_coordinates(checkpoint_dir_path, redistributed=True):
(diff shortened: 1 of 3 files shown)


jchodera:
We should probably establish some sort of more thorough testing-prior-to-release protocol since we're now regularly needing to issue multiple immediate point bugfixes following release.

I'd like to rope in @jaimergp here to help develop a release checklist and testing procedure that could help us in future iterations? I realize you need the bugfix releases out ASAP for the trailblazing paper, but for subsequent releases, it would be great to have him involved in testing and cutting releases.

@j-wags has developed a nice [release checklist](https://github.com/openforcefield/openforcefield/issues/286) for `openforcefield` releases, and we should probably have something here that also involves running some test suite on a real GPU cluster before cutting a release.

andrrizzi:
Absolutely, I completely agree, and I'd love it if @jaimergp had time to help.

Usually, the releases are tested internally for much longer, but as you point out, this was an exception due to the need to run the production calculations with a released version. The problem was that no test was covering this particular case (which I've now added), which shows up only in the particular case where something goes wrong during trailblaze, and the algorithm is resumed.

Issue #760, opened 2017-08-23
Add CHARMM input scheme


leelasd:
Hi,

 I have been using Parmed/OpenMM to implement OPLS-AA FF based simulations. Is there anyway to use these systems for performing hydration free energy/binding free energy calculations. Any suggestions are appreciated.

davidlmobley:
You would need to be more specific about what you want to know. There certainly is a way to do this, yes.

jchodera:
@leelasd : Can you provide more information about what you're hoping to do here? Are you already able to set up systems with these forcefields in some other program, like gromacs?

leelasd:
@jchodera @davidlmobley Sorry for the late reply.

I am trying to calculate binding free energy for Lysozyme-Benzene system. I am using PSF and PRM files (generated with VMD) to perform MD with Parmed/OpenMM. 
I attached the PSF, PDB and Parameter files used to do MD. 
[OMM_MD.zip](https://github.com/choderalab/yank/files/1275419/OMM_MD.zip)

I want to know if I can use the same Parmed/OpenMM with Yank to do binding FEP calculation. 
Any suggestions would be great.

jchodera:
We don't currently have a CHARMM reader, but it should be easy to add one so you can use these files! I'll change of the title of the issue to reflect this.

leelasd:
@jchodera can you point me to the module where you create Yank/OpenMM system from Gromacs file. I can contribute that part

jchodera:
Contributions would be welcomed!

I believe you would need to do the following:
* Add a section to [this function](https://github.com/choderalab/yank/blob/master/Yank/experiment.py#L1285-L1294) to make YANK aware of CHARMM system files
* Add a schema to let the user specify a sufficient combination of `pdb`, `psf`, and `prm` files [here](https://github.com/choderalab/yank/blob/master/Yank/experiment.py#L1332-L1353)
* Add a section to [this function](https://github.com/choderalab/yank/blob/master/Yank/pipeline.py#L381-L478) to handle loading of [CHARMM files using OpenMM](http://docs.openmm.org/latest/userguide/application.html#using-charmm-files)

@andrrizzi can help review your work if you run into trouble!

leelasd:
Thanks !! I will start working on that !!

davidlmobley:
Of course, the other approach is to see this as a ParmEd issue and not a Yank issue and just convert your files to a format Yank DOES support using ParmEd.

jchodera:
[...]

PR #1206, opened 2020-02-07
Added support for CHARMM-format input files

jeffcomer:
The new code allows a psf/pdb combination to used with phase1_path and phase2_path. An additional variable charmm_parameter_files allows any number of CHARMM prm/str files to be included. This commit should fulfill yank github issue #760 ( https://github.com/choderalab/yank/issues/760 ).

One current limitation is that Yank doesn't recognize the CHARMM names for counterions (SOD, CLA). My workaround is to do the replacements "s/SOD/Na+/" and "s/CLA/Cl-/" in the psf, pdb, and toppar_water_ions.str files.

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -1706,6 +1706,14 @@ class ExperimentBuilder(object):
             type: string
             dependencies: [phase1_path, phase2_path]
             check_with: directory_exists
+        charmm_parameter_files:
+            required: no
+            type: list
+            schema:
+                type: string
+                check_with: file_exists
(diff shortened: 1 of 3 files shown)


jeffcomer:
I don't think the "codeclimate — 1 issue to fix" is a problem. I added an optional keyword to read_system_files(), "charmm_parameter_files=None", which allows you to give a list of CHARMM .prm and .str files. This keyword is patterned after "gromacs_include_dir", which also allows additional information for a specific input file format. Like "gromacs_include_dir", it should not affect runs that don't use it.

PR #1213, opened 2020-03-21
Small fixes to docs and to solute-only trajectory

### LINKED ISSUE #1200: Fix docs on implicit solvent salt concentration option
When trying to set a default salt concentration, using the syntax written in the manual:
(implicit_solvent_salt_concentration: 0.15 * moles / liter)
I encounter the following error:

    raise YamlParseError(error.format(solvent_id, yaml.dump(solvent_validator.errors)))
    yank.experiment.YamlParseError: Solvent 'GBSA' did not validate! Check the schema error below for details
    implicit_solvent_salt_concentration:
    - unknown field

It seems the syntax is off. Has it been updated in the source without the manual page being updated?

### LINKED ISSUE #1212: Scrambled atoms in solute-only trajectory
@ajsilveira found an issue some time ago with systems prepared without the automatic pipeline in which the ion atoms appear before the solute atoms. This case causes the solute-only trajectory to represent the correct positions but for a different order of the atoms (with weird results).

andrrizzi:
- Fix #1200 and other small fixes to the documentation.
- Fix #1212, which caused a problem with the stored solute-only trajectories for systems prepared without the automatic pipeline.

### DIFF
--- a/Yank/experiment.py
+++ b/Yank/experiment.py
@@ -234,7 +234,7 @@ class AlchemicalPhaseFactory(object):
             if self.options['store_solute_trajectory']:
                 # "Solute" is basically just not water. Includes all non-water atoms and ions
                 # Topography ensures the union of solute_atoms and ions_atoms is a null set
-                solute_atoms = self.topography.solute_atoms + self.topography.ions_atoms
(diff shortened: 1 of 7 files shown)


andrrizzi:
I don't think the error in the Python 3.7 tests is related to this PR. This is good to go on my end.

Issue #1214, opened 2020-03-24
Trailblazing analysis error


mjw99:
Dear Authors,

I am currently attempting some trailblazing calculations using YANK for a screen of some drug fragments against the SARS-CoV-2 main protease. The repo is [here](https://github.com/mjw99/COVID_19_Drug_Binding).  

I am seeing this (as of commit c991b54) on the analysis of the x0072-explicit-trailblazing.yaml run: 
```
yank analyze --store=./x0072-explicit-trailblazing-output/experiments/

(openmm) [mw529@login-e-6 COVID_19_Drug_Binding]$ yank analyze --store=./x0072-explicit-trailblazing-output/experiments/
/home/mw529/.conda/envs/openmm/lib/python3.7/site-packages/yank/analyze.py:299: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  analysis = yaml.load(f)
2020-03-22 21:29:22,651: WARNING - openmmtools.multistate.multistatereporter - Warning: The openmmtools.multistate API is experimental and may change in future releases
2020-03-22 21:29:22,777: WARNING - openmmtools.multistate.multistateanalyzer - Warning: The openmmtools.multistate API is experimental and may change in future releases
2020-03-22 21:29:22,781: WARNING - openmmtools.multistate.multistatereporter - Warning: The openmmtools.multistate API is experimental and may change in future releases
2020-03-22 21:29:22,820: WARNING - openmmtools.multistate.multistateanalyzer - Warning: The openmmtools.multistate API is experimental and may change in future releases
2020-03-22 21:29:23,353: WARNING - yank.analyze - Alert: analyzed timeseries has the same or fewer number of values as discard_from_start! The whole series has been preserved to ensure there is *something* to analyze.
/home/mw529/.conda/envs/openmm/lib/python3.7/site-packages/numpy/core/_methods.py:217: RuntimeWarning: Degrees of freedom <= 0 for slice
  keepdims=keepdims)
/home/mw529/.conda/envs/openmm/lib/python3.7/site-packages/numpy/core/_methods.py:186: RuntimeWarning: invalid value encountered in true_divide
  arrmean, rcount, out=arrmean, casting='unsafe', subok=False)
/home/mw529/.conda/envs/openmm/lib/python3.7/site-packages/numpy/core/_methods.py:209: RuntimeWarning: invalid value encountered in double_scalars
[...]

Issue #1203, opened 2019-12-12
yank analyze error


jslim-furame:
I did 10 ns GBSA simulation (number of iterations = 10000, 1 iteration = 1ps). The simulation completed successfully. However, yank analyze gives the following error 

pymbar.utils.ParameterError: Warning: Should have \sum_n W_nk = 1.

What can I do? I read some answers posted in pymbar GitHub, but those answers are ambiguous.

andrrizzi:
Hi @jslim-furame . I'd try downgrading to pymbar 3.0.3 if you haven't done so already. This seems to have solved the issue for one of the users (see choderalab/pymbar#320).

jchodera:
@andrrizzi : Should we pin YANK's recipe to `pymbar==3.0.3` until @mrshirts figures out who broke pymbar and why?

andrrizzi:
I think that is a good idea.

PR #1216, opened 2020-03-26
pinning pymbar version

hannahbrucemacdonald:
Addresses issues #1214 and #1203


andrrizzi:
Merging now. I'll take care of fixing also the recipe in omnia-md/conda-recipes .

Issue #1260, opened 2021-10-14
Module not found after yank update


gsabbih6:
ModuleNotFoundError: No module named 'simtk.openmm.app.element'

mjw99:
This might be related to the recent [top level Python package rename in OpenMM 7.6](https://github.com/openmm/openmm/releases/tag/7.6.0). Try installing an earlier version of OpenMM to your conda environment, e.g.:
```
conda install -y -c conda-forge openmm=7.5.1 
```

jchodera:
@ijpulidos @mikemhenry : Can you help here? It looks like we need to cut new bugfix releases of yank and openmmtools that work with both openmm <7.6 and >=7.6

mikemhenry:
I think we just need to cut a new release of `openmmtools` since this PR should fix the namespace issues and work with both openmm <7.6 and >=7.6

https://github.com/choderalab/openmmtools/pull/528

I'll see what we need to do in yank

mikemhenry:
That will fix when the issue comes from OpenMMTools, I've got this PR https://github.com/choderalab/yank/pull/1263 to fix spots where yank imports openmm, but I need to fix the CI

PR #1180, opened 2019-09-01
Fix Cerberus and support for trailblaze options

andrrizzi:
* Fix #1141: YANK is now compatible with Cerberus >= 1.2, which means we should be able to remove cerberus from omnia.
* Fix #1139: Expose more options for the trailblaze algorithm in YAML. It is now possible to tweak the trailblaze runs. This is the example that I added to the Sphinx docs.

```yaml
protocols:
  {UserDefinedProtocol}:
    {PhaseName1}:
      trailblazer_options:
        # Number of initial equilibration iterations performed in the first state before running the algorithm.
        n_equilibration_iterations: 1000
        # Whether the heavy atoms of the receptor are positionally restrained with a harmonic potential during the algorithm.
        constrain_receptor: false
        # Number of samples used to estimate the standard deviation of the potential between two states.
        n_samples_per_state: 100
        # The "distance" in potential standard deviation between two intermediate states up to 'threshold_tolerance'.
        std_potential_threshold: 0.5  # in kT
        threshold_tolerance: 0.05  # in kT
        # Whether to traverse the path in the forward (given by 'alchemical_path') or reversed direction.
        reversed_direction: true
        # A variable controlling the path if 'alchemical_path' contains mathematical expressions.
        function_variable_name: lambda

      alchemical_path:
        # Deactivate electrostatic interactions before decoupling the Lennard-Jones potential.
        lambda_electrostatics: 2*(lambda - 0.5) * step(lambda - 0.5)
        lambda_sterics: step_hm(lambda - 0.5) + 2*lambda * step_hm(0.5 - lambda)
        lambda: [1.0, 0.0]
```
``alchemical_path: auto`` is still possible, with or without the new block ``trailblazer_options``.

PR #1263, opened 2021-12-13
Update for new OpenMM Namespace

mikemhenry:
Update to support openmm >= 7.6 namespace, but also support older versions of openmm as well

### DIFF
--- a/Yank/analyze.py
+++ b/Yank/analyze.py
@@ -30,7 +30,10 @@ from functools import wraps
 
 import mpiplus
 import numpy as np
-import simtk.unit as units
+try:
+    import openmm.unit as units
+except ImportError: # OpenMM < 7.6
+    import simtk.unit as units
 import openmmtools as mmtools
 from pymbar import timeseries
 
--- a/Yank/commands/analyze.py
+++ b/Yank/commands/analyze.py
@@ -151,7 +151,10 @@ def dispatch(args):
 
 def extract_analyzer_kwargs(args, quantities_as_strings=False):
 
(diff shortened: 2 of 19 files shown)


mikemhenry:
fixes https://github.com/choderalab/yank/issues/1260

ijpulidos (approved):
Thanks, this looks great!

mikemhenry:
@jchodera it looks like the CI for Yank has been broken for a very long time, do you know what could be causing these schema errors? It looks like the same errors are showing up in the scheduled CI happening on master 
https://github.com/choderalab/yank/actions 
So this PR isn't the cause, and it doesn't create any new issues, but if there is an easy fix for the schema errors, it would be nice to get the CI working again!

jchodera:
@mikemhenry : From memory, I think we've run into cases where updates to `cerberus` (the tool we use to provide helpful error messages when users specify incorrect YAML schemas) cause things to fail for us. We may just need to pin `cerberus` for now and sort out the issue later.

Here's an example of a previous instance in which cerberus broke things: https://github.com/choderalab/yank/pull/1180

mikemhenry:
Thanks! It looks like it is pinned to  1.1, but the PR makes it sound like it was updated to 1.2
`  - cerberus ==1.1 # yank uses buggy cerberus 1.1 behavior as a feature ¯\_(ツ)_/¯`
I will play around with that, thanks!

mikemhenry:
@jchodera you might want to start using this PR to base your yank updates on since it takes care of the namespace changes in openmm

PR #1267, opened 2022-03-09
Stable minimization

jfennick:
I have been experiencing some instabilities which, after much difficulty, I was able to track down to the minimizer. The observed behavior is that the minimizer will finish 'successfully', but some time later, during equilibration and/or propagating replicas, the simulation may or may not randomly crash with NaNs or even throw CUDA_ERROR_ILLEGAL_ADDRESS (700).

Nearly identical behavior can be found here https://github.com/openmm/openmm/issues/3414. In that ticket, the issue was that the box vectors were not being updated correctly. For stability, during minimization the box vectors should not be allowed to change at all. Subsequent equilibration at NPT should be used to adjust the box vectors.

In my case(s), the box vectors were changing only at the fourth significant figure, and yet that was sufficient to (usually) crash the simulations. Inserting Gradient Descent before FIRE seems to greatly ameliorate but not completely eliminate the instabilities. Replacing FIRE with L-BFGS (with or without Gradient Descent) also seems to fix the instabilities.

This PR adds three changes to improve the stability of minimization:

1. First and foremost, it temporarily disables the barostat during minimization to keep the box vectors fixed. This is probably the only change that is absolutely necessary to fix the instabilities.
2. It inserts a very short (24 timestep) Gradient Descent minimization before the main FIRE minimization. Gradient Descent is not asymptotically fast, but it is excellent for preconditioning a 'better' minimizer.
3. I have replaced FIRE with L-BFGS. Unlike FIRE, L-BFGS does not modify the box vectors (even when the pressure is not None) and it's a fine minimizer anyway.

### DIFF
--- a/Yank/yank.py
+++ b/Yank/yank.py
@@ -1452,32 +1452,17 @@ class AlchemicalPhase(object):
         logger.debug('energy {:8.3f}kT box_volume {:6.4f}nm**3 bxx {:2.4f}nm  byy {:2.4f}nm  bzz {:2.4f}nm'.format(
             energy, volume, bxx, byy, bzz))
 
-
-        # Use the FIRE minimizer
-        integrator = mmtools.integrators.FIREMinimizationIntegrator(tolerance=tolerance)
-
         # Create context
-        context = thermodynamic_state.create_context(integrator)
+        context = thermodynamic_state.create_context(integrator_grad)
 
         # Set initial positions and box vectors.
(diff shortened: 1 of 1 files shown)


jfennick:
See also https://github.com/choderalab/openmmtools/pull/557

jfennick:
FYI I just realized that I accidentally named this branch stable_equilibration instead of stable_minimization.

Issue #1152, opened 2019-03-28
yaml.load issues


jchodera:
```
/Users/choderaj/miniconda/lib/python3.6/site-packages/yank/experiment.py:1168: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
```
See: https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

Lnaden:
The new PyYaml does this. Here was the snippet I came up with to suppress that safely:

```python
 try:
    yaml_load = partial(yaml.load, Loader=yaml.FullLoader)
except AttributeError:
    yaml_load= yaml.load
```

since `yaml.FullLoader` isn't in the versions before that warning.

jchodera:
> The new PyYaml does this.

Can you clarify if you mean that the new pyyaml (1) emits this warning, or (2) automatically calls a safe loader?

Lnaden:
PyYaml emits the warning in the most recent version (5.1), I don't think its functionally doing anything different. The site it links to has some shortcut methods to chose the loader you want, but I don't know if all of those methods are in the older versions, which is why I use the `try...except`.

The site also says by default it uses the `FullLoader` which is safe, but the raw `load` without a `Loader` keyword may throw an error in the future.

lilyminium:
pyyaml 6.0 (released 27 days ago) now raises an error upon loading without a Loader. You may need to start pinning dependencies in releases or update the code.

Lnaden:
I recall going through the code and replacing all the `yaml.load`'s with safe versions that I could find. Are you seeing an error being thrown now somewhere and/or did I miss some of the loaders?

lilyminium:
The loader in `analyze.py`: https://github.com/choderalab/yank/blob/aef24d9f413e646a3c4e65581198e0b67ae7a21b/Yank/analyze.py#L298-L301

I noticed it because openff-evaluator uses it :-)

PR #1268, opened 2022-03-19
Update yaml loader in analyze.py

lilyminium:
Related to #1152

Not sure if tests are practical; will check if they get past the conda installation stage...


mikemhenry:
Thanks for the PR! :tada: 
Sorry it took so long for me to trigger these tests to run!

PR #1283, opened 2022-07-16
extend on resume

jfennick:
This PR allows the number of timesteps to be increased upon resuming a simulation. See https://github.com/choderalab/openmmtools/pull/604


jfennick:
Seems like the CI may be failing because of this issue? https://github.com/conda-incubator/setup-miniconda/issues/116 Maybe`conda-incubator/setup-miniconda@v2` needs to be updated to the latest version?

PR #1284, opened 2022-07-19
Add Python 3.9 and 3.10 to CI matrix

mattwthompson:
This won't work at first because of at least one API break in `collections` but what I detect downstream is almost surely more than what the actual test suite will find.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -19,13 +19,19 @@ jobs:
       fail-fast: false
       matrix:
         cfg:
-          - { os: ubuntu-latest, python-version: 3.7, docs: true,  openmm: latest }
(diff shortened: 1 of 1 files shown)


mikemhenry:
@mattwthompson see https://github.com/conda-forge/yank-feedstock/pull/4

mikemhenry:
Oh and https://github.com/choderalab/yank/pull/1286

So yah need to make some changes :smile: I had forgotten that you made this PR ages ago!

mattwthompson:
Heh, 2 days isn't quite ages - but happy to see you move forward on this, I think it's one of my last upstreams to get on 310!

mikemhenry:
> Heh, 2 days isn't quite ages - but happy to see you move forward on this, I think it's one of my last upstreams to get on 310!

You are right! I was thinking this was from a totally different thing!

